PR #1, opened 2017-01-05
Added timer to search, clear search on empty query

jeremythuff:
This is a very small experimental PR to feel out the process for contribution and to begin aligning my coding practices with those adopted by this project. I have implemented 2 small UX tweaks to the user search

1) All search results are shown when the search box is cleared using the backspace button, and

2) I noticed that a query was being sent to the backend with each keystroke, so I implemented a timeout that waits for the typing to halt before executing the query.

### DIFF
--- a/Users.js
+++ b/Users.js
@@ -61,7 +61,6 @@ class Users extends Component {
       addUserMode: false,
     };
 
-    this.searchTimer = null;
     this.onClickAddNewUser = this.onClickAddNewUser.bind(this);
     this.onClickCloseNewUser = this.onClickCloseNewUser.bind(this);
     this.onChangeFilter = this.onChangeFilter.bind(this);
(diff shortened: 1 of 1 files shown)


skomorokh:
Thanks! I'm not really back from vacation until after the 20th, just popping in part-way in case there were exciting problems and to add my 2 cents randomly like here :) 

Since we already use lodash elsewhere (and our dependencies do as well) I'd recommended using it to debounce that search box: 
https://lodash.com/docs#debounce

However, I also suspect when we are doing such optimisation in earnest we'll likely end up implementing something similar in stripes-connect along with redundant request handling / caching, especially if our Okapi APIs will allow us to batch them (though HTTP2 may obviate the need for that). eg. module authors can just throw everything at stripes-connect and let that layer handle the details.

OTOH this is a fairly cut and dry case so we'd probably want to debounce it even still.

MikeTaylor:
Hi, Jeremy. Great to see this!

I do agree with Jason, though, that if we're going to do the delay up at the front end, we'd want to use the lodash facility rather than rolling our own.

For now, could you please submit a smaller pull-request with only your change #1? Then we can verify that the process works as we expect; and then go from there.

PR #5, opened 2017-02-16
Added username to the view path pattern

jeremythuff:
So that it can be available right off the bat, the username has been added to the path pattern in addition to the user id. This is so that the permissions can be obtained via the username. This resolves [STRIPES-193](https://issues.folio.org/browse/STRIPES-193)

### DIFF
--- a/Users.js
+++ b/Users.js
@@ -158,9 +158,10 @@ class Users extends Component {
 
   onSelectRow(e, meta) {
     const userId = meta.id;
+    const username = meta.username;
     console.log('User clicked', userId, 'location = ', this.props.location);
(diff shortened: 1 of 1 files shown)


MikeTaylor:
I am uncomfortable about this. Do we _really_ want to redundantly include both userId _and_ username in the URL? What happens if we get a path those userId and username don't match?

skoczko:
We need this right now because permissions and login information is indexed using 'username' and we cannot "chain" manifests (username is available after retrieving the user record by id).

We decided to introduce this as a temporary measure and eventually move to a business-logic module that combines all data in one responses (and only needs id).

PR #6, opened 2017-02-22
Libapp 50

jeremythuff:
The ability to add and remove permissions from users. Additionally work has been done to filter permissions using a search input contained within the permissions dropdown. This is all done client-side and will be replaced with cql


MikeTaylor:
Thanks for this, Jeremy. It's past 6pm here now, so I won't review this tonight; then Jason and I are out at Rangle all day tomorrow, then I am flying home and won't arrive till well into Friday at which point I will collapse and sleep. The upshot is, I won't be able to review this pull-request before Monday.

So your call: if you like, I can just go ahead and merge the pull-request; or if you prefer, we can leave it till Monday and I can review it properly. Let me know how to proceed.

jeremythuff:
That may break down to exigency, but from my perspective, with it being my first feature to be merged, I suppose I would like the comfort of knowing it received a good looking over.

MikeTaylor:
OK. Niels-Erik is welcome to review it if he has time; if he doesn't I'll try to look at it on Monday. Sorry for the delay.

MikeTaylor:
I notice you use:
```
  addPermission(perm) {
    this.props.viewUserProps.mutator.usersPermissions.POST(perm, this.props.viewUserProps).then(() => {
      this.onToggleAddPermDD();
    });
  }
```
Does this actually work?

MikeTaylor:
I also notice this:
```
  removePermission(perm) {
    this.props.viewUserProps.mutator.usersPermissions.DELETE(perm, this.props.viewUserProps, perm.permissionName);
  }
```
What are you expecting to happen with the three arguments passed to the DELETE mutator? As as far as I can see in the Stripes Connect code, it only accepts one argument.

PR #10, opened 2017-04-06
Libapp 83

jeremythuff:
This is a feature complete PR for [LIBAPP-83](https://issues.folio.org/browse/LIBAPP-83).

The DELETE of permissions was implemented as a button on the permission details pane. This is not consistent with the prototype, but the drop down which housed the delete button in the prototype is not yet implemented in stripes.

[STRIPES-301](https://issues.folio.org/browse/STRIPES-301) is a bug that remains and is related to the RenderPermssions component not showing up in the permission set pane if the user does not have perms.user based permissions. This should be addressed, as it is not desired behavior.

### DIFF
--- a/UserPermissions.js
+++ b/UserPermissions.js
@@ -13,13 +13,14 @@ import css from './UserPermissions.css';
 
 const propTypes = {
   stripes: PropTypes.shape({
-    hasPerm: PropTypes.func.isRequired,
-  }).isRequired,
+      hasPerm: PropTypes.func.isRequired,
+    }).isRequired,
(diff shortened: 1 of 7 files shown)


MikeTaylor:
Folks, I have discussed this PR with Jeremy. Please no-one merge it: he's going to do some more work, then come back to me.

jeremythuff:
I have resolved the concerns related to [STRIPES-301](https://issues.folio.org/browse/STRIPES-301) and used eslint on my new code as far as I was able to make sense of/resolve the issues it brought to my attention.

PR #12, opened 2017-05-09
UIU-4 https for placehold.it

julianladisch:


julianladisch:
replace placehold.it link by local image: https://issues.folio.org/browse/UIU-35

PR #15, opened 2017-05-31
Uiu 1

jeremythuff:
This is a PR for issue [UIU-1](https://issues.folio.org/browse/UIU-1).

It is not an ideal solution, as it employs some work arounds which might be mitigated after [STRIPES-211](https://issues.folio.org/browse/STRIPES-211) is complete. It utilizes the componentDidMount lifecycle method to do a comparison of the list of permission sets to determine if a new one has been added. If there is a new member of that list, then it sets that new resource to the selected permissionSet.


MikeTaylor:
I think this is a really neat approach, actually. I don't know if I would have come up with it, but I am glad that you did!

PR #26, opened 2017-06-29
Uiu 109

jeremythuff:
For [UIU-109](https://issues.folio.org/browse/UIU-109)

This addresses code cleanliness concerns in the following ways:

Removed homogenized references to both state.selectedSet and props.SelectedSet in favor of state.selectedSet.

state.selectedSet is now instantiated explicitly in the constructor instead of implicitly in the validation method.

PermissionSet was refactored to ContainedPermissions which is less ambiguous and semantically more accurate.

Permissions code is ESLint compliant.


MikeTaylor:
So sorry that this slipped through my net, now nearly a month ago! I should have accepted it almost immediately. If this happens again, please ping me! Anyway, these changes are obviously good, so I will go ahead and merge now.

PR #58, opened 2017-08-22
Stripes 464

jeremythuff:
This is a top level module implementation of runModule from[ STRIPES-464](https://issues.folio.org/browse/STRIPES-464) by way of example.

It is dependent on [PR#35](https://github.com/folio-org/stripes-core/pull/35) to stripes-core, and should not be merged until that is in place.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## 2.10.0 (IN PROGRESS)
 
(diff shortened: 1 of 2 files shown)


MikeTaylor:
I think `runModule` happens needs to happen much earlier in the process than this -- most likely when the top-level modules are first instantiated in `stripes-core/src/moduleRoutes.js`. It's not a substitute for _connecting_ a module, but the way that we set up a React component to be a Stripes module at all.

PR #62, opened 2017-09-07
STCOM-20  Accordionate the Sections of User Details

JohnC-80:
This touched the components in the View User page, performing a bit of surgery on each to get the accordions wired up to eventually have them persist on a local resource.

Before I went through this step I attempted to keep the Accordions at the top, wrapping the components - but the design requires the accordion headers to contain buttons that interact with the sub-sections - and show/hide these buttons based on the open or closed status of the accordion. There isn't any good non-hacky way for a parent to reach in and use the code that's already present in a child component (only props, themselves) so I just ended up incorporating the Accordions themselves into the components, passing in the appropriate props to control them from ViewUser...

In addition to this, the AddressList component of stripes-components has also been modified... this branch alone will still render working components, you'd just see that the Address section is not affected.

Special credit goes to Zak Burke for echoing some of my sentiments over this process in cartoon-form: https://mainlynorfolk.info/john.kirkpatrick/images/large/farsideaccordion.jpg

### DIFF
--- a/ProxyPermissions.js
+++ b/ProxyPermissions.js
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types';
 import { Row, Col } from 'react-bootstrap';
 import MultiColumnList from '@folio/stripes-components/lib/MultiColumnList';
 import Pluggable from '@folio/stripes-components/lib/Pluggable';
+import { Accordion } from '@folio/stripes-components/lib/Accordion';
 
 import { getFullName, getRowURL, getAnchoredRowFormatter } from './util';
 
@@ -27,11 +28,9 @@ const propTypes = {
       PUT: React.PropTypes.func.isRequired,
     }),
   }).isRequired,
-  displayHeading: PropTypes.bool,
-};
-
-const defaultProps = {
-  displayHeading: true,
+  expanded: PropTypes.bool,
+  onToggle: PropTypes.func,
+  accordionId: PropTypes.string.isRequired,
 };
 
 class ProxyPermissions extends React.Component {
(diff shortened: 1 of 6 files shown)


MikeTaylor:
@mkuklis That's a good call. I created https://issues.folio.org/browse/UIU-208

MikeTaylor:
I assume the `backendDateStandard="YYYY-MM-DD"` addition to UserForm.js is unrelated to the rest of this PR?

MikeTaylor:
I think your `handleSectionToggle` function mutates the original state before returning the new one:
```
  handleSectionToggle({ id }) {
    this.setState((curState) => {
      const newState = curState;
      newState.sections[id] = !newState.sections[id];
      return newState;
    });
  }
```
Because `newState` is only a shallow copy of `curState`.

I don't know what the consequences of this are, but I fear they might be "everything works fine for six months, then weird things start happening and we have no idea why".

I think you want something like
```
  handleSectionToggle({ id }) {
    this.setState(Object.assign({}, curState, {
      sections: Object.assign({}, curState.section, { id: !curState.section[id] }
    });
  }
```

(*Warning*: this code has not been checked, not even for syntax.)

MikeTaylor:
Why has the `list-finehistory` part gone from ViewUser?

MikeTaylor:
Otherwise, this all looks good. It's encouraging how cleanly the accordions integrate. (The diffs look much bigger than they are because of the re-indentation of the contained code, of course.)

Only one more thing: have you linted the new code?

zburke:
Lint runs cleanly for me on this branch.

JohnC-80:
Thank you guys! @MikeTaylor I'll rewrite handleSectionToggle so that it performs as it should - and I'll try hooking them up to the local resource so that their state can persist (seems a low-hanging fruit)

fineHistory is gone because that part was purely static and non-functional at all - it's been there since the 'old days'. I can put it back with its own accordion if we really need it...

MikeTaylor:
Thanks, @JohnC-80, that all makes sense. No need at all to reinstate fine history, I was just curious.

I suggest you just fix the things I mentioned above -- probably just avoiding the state-mutation -- and then we can merge this PR. The work of moving the state into local resources is logically separate, and can be its own PR later.

JohnC-80:
Updated. Used lodash's `cloneDeep` to copy the state. looks like this: 
  ``` 
handleSectionToggle({ id }) {
    this.setState((curState) => {
      const newState = _.cloneDeep(curState);
      newState.sections[id] = !newState.sections[id];
      return newState;
    });
  } 
```

PR #67, opened 2017-09-28
Remove Return Date Column from Open Loans Tab. Fixes UIU-225

mkuklis:


mkuklis:
@zburke thanks for catching it!

PR #70, opened 2017-10-02
Support Bulk Renew on Loans. Fixes UIU-214.

mkuklis:
I decided to break `LoansHistory` into two separate components: `OpenLoans` and `ClosedLoans` since they started to get more complex.

### DIFF
--- a/LoanActionsHistory.js
+++ b/LoanActionsHistory.js
@@ -7,8 +7,6 @@ import KeyValue from '@folio/stripes-components/lib/KeyValue';
 import MultiColumnList from '@folio/stripes-components/lib/MultiColumnList';
 import Pane from '@folio/stripes-components/lib/Pane';
 import Paneset from '@folio/stripes-components/lib/Paneset';
-import fetch from 'isomorphic-fetch';
-
 import { formatDateTime, getFullName } from './util';
 import loanActionMap from './data/loanActionMap';
 
@@ -16,6 +14,7 @@ class LoanActionsHistory extends React.Component {
   static propTypes = {
     stripes: PropTypes.object.isRequired,
     resources: PropTypes.shape({
+      loanActions: PropTypes.object,
       loanActionsWithUser: PropTypes.object,
       userIds: PropTypes.object,
     }).isRequired,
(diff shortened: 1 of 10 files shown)


mkuklis:
@zburke unfortunately it looks like this was an exiting issue and a limitation in `stripes-connect`.

There is currently no way to refetch a resource after the connected component has been previously mounted and the same param is being passed (in this case the same loan). The fetch won't happen in this case. 

There was also a an existing limitation with joining users with loan history. The `componentWillReceiveProps` was getting out of hand there. I refactored it with `isomorphic-fetch` for the time being. This should hopefully resolve the issue you ran into. We may have to revisit this when `stripes-connect` has support for it or we switch to some other tool (GraphQL).

MikeTaylor:
We should feel deep, burning shame every time we revert to isomorphic fetch. Every use of it represents a failure -- sometimes of the underlying software, but more often in my experience a failure of visualisation by a developer. It's like falling back to modifying global variables in Lisp, or something. You can do it, it will get you where you need to go, but it's not _right_.

mkuklis:
@MikeTaylor you challenged me to take another look at it and I noticed that perhaps the issue here that we no longer rely on `props.data` (it was replaced by `props.resources`) but I noticed this in `stripes-connect`:

https://github.com/folio-org/stripes-connect/blob/master/connect.js#L130-L146

It seems like we are still referencing `props.data` there and it could explain why the refresh didn't work. I will try to use `resources` instead and see what happens.

mkuklis:
@zburke and @MikeTaylor I went back to what we had before in `LoanActionsHistory` based on `stripes-connect`. This is still not fixed but I realized this is not part of this work at all.  I'm going to fill a new issue for it and hopefully we can merge this in order to move forward.

mkuklis:
Here are the two new issues:

https://issues.folio.org/browse/STCON-38 
https://issues.folio.org/browse/UIU-248

mkuklis:
@zburke thank you again for reviewing this. These reviews are helpful!

MikeTaylor:
Sounds good. Thanks for responding so well to my pushback :-)

PR #71, opened 2017-10-02
Display count when loans section is collapsed. Fixes UIU-233

mkuklis:


mkuklis:
@Baroquem apologies for taking your time but I have one more for you to review.

PR #72, opened 2017-10-03
Display Number of Loans. Fixes UIU-218

mkuklis:


Baroquem:
Michal, the `renderSubHeader()` function in ClosedLoans.js doesn't seem to be used (unlike the one in OpenLoans, which is used). Is it still needed there?

mkuklis:
@Baroquem thank you for the review! You are right it was a left over. I just cleaned that up.

PR #73, opened 2017-10-04
Display proxy count and change the section name. Fixes UIU-232 and UIU-236.

mkuklis:


JohnC-80:
`<Badge` should probably be a shared component - what do you guys think?

mkuklis:
@JohnC-80 I was going to ask you about it. It could be helpful but I didn't want to push for it just yet. Let's merge this one and perhaps extract it back to stripes-components? Does that make sense?

PR #80, opened 2017-10-06
Fix renew date. Fixes UIU-252

mkuklis:


JohnC-80:
We use javascript Date objects in many places... One of the deps of stripes-components is [moment.js](http://momentjs.com/). As the subject of localization increases in popularity, I wonder if the Javascript Date objects will continue to stand up for us. I see we've invested some time in constructing utility functions, as small as they are.... do you guys have any thoughts on using moment.js instead?

mkuklis:
@JohnC-80 I didn't realize we have moment in stripes-components. Are there any helpers already provided? I think it would be helpful to have a reusable util just for dates. ui-users has this included:
https://github.com/felixge/node-dateformat so that's what I used but would be happy to switch to moment.

JohnC-80:
moment is prettymuch its own helper... you should just be able to import and use its API directly.

mkuklis:
@JohnC-80 sounds good. I was actually looking at moment but didn't want to bring another dependency to `ui-users`.  But since it's already part of `stripes-components` I will use it and get rid of the `dateformat`. Thanks John!

mkuklis:
Ok @JohnC-80 moment should be now in place.

PR #93, opened 2017-10-11
STCOM-40  Apply <Pane>'s 'noOverflow' prop to results pane

JohnC-80:
Hopefully alleviating the painfully annoying symptom that was noticeable on the projector at Montreal and occurs on almost everyone else's development/demo machine except for my own!


JohnC-80:
Thanks @Baroquem !

PR #100, opened 2017-10-17
Show loan's proxy borrower. Refs UIU-177.

zburke:


zburke:
@mkuklis Good idea to use `util.js:: getFullName()` to format the name. 

Re-reading the JIRA ticket, I realized the link to the proxy-user's profile was missing as well. That's been added, along with a click-handler to clean out the loan-related state so clicking the link actually displays the proxy-user's profile page rather than the proxy-user's loans modal.

PR #119, opened 2017-11-03
Refactor async user validator. Fixes UIU-302.

mkuklis:


MikeTaylor:
I  can scarcely express how happy this makes me.

PR #125, opened 2017-11-07
Prepare Autocomplete for Bootstrap global css removal

cherewaty:
## Purpose
Removing Bootstrap's global CSS from `stripes-core` (https://github.com/folio-org/stripes-core/pull/61) breaks the `Autocomplete` component used for country selection in addresses.

This should unblock https://github.com/folio-org/stripes-core/pull/61, so we can move toward the intended font stack and surface other CSS that was leaning too heavily on the global Bootstrap styles.

## Approach
I copied over bits and pieces of Bootstrap CSS into the `Autocomplete` component's CSS as a stop-gap until there's an autocomplete component available in `stripes-components` up to FOLIO standards. Now the component is still functional without Bootstrap CSS, even though it will show the "Close" text previously hidden by a global `.sr-only` class.

## Screenshot
The `Autocomplete` component, with the contents of this PR and no global Bootstrap styles:

thqoc2ibyc

### DIFF
--- a/lib/Autocomplete/Autocomplete.css
+++ b/lib/Autocomplete/Autocomplete.css
@@ -20,19 +20,67 @@
   border: 1px solid #acacac;
   font-size: 14px;
   color: #333;
+  width: 100%;
 }
 
 .autocomplete button {
   float: none;
   position: absolute;
-  top: -9px;
-  right: 0px;
+  top: 2px;
+  right: 5px;
   z-index: 100;
   opacity: 1;
+  -webkit-appearance: none;
+  padding: 0;
+  cursor: pointer;
+  background: 0 0;
+  border: 0;
 }
 
 .autocomplete button span {
-  font-size: 40px;
+  font-size: 20px;
   font-weight: 300;
   color: #999;
+}
+
+.autocomplete ul {
+  position: absolute;
+  top: 100%;
+  left: 0;
+  z-index: 1000;
+  display: none;
+  float: left;
+  min-width: 160px;
+  padding: 5px 0;
+  margin: 2px 0 0;
+  font-size: 14px;
+  text-align: left;
+  list-style: none;
+  background-color: #fff;
+  -webkit-background-clip: padding-box;
+  background-clip: padding-box;
+  border: 1px solid #ccc;
+  border: 1px solid rgba(0,0,0,.15);
+  border-radius: 4px;
+  -webkit-box-shadow: 0 6px 12px rgba(0,0,0,.175);
+  box-shadow: 0 6px 12px rgba(0,0,0,.175);
+}
+
+.autocomplete li a {
+  display: block;
+  padding: 3px 20px;
+  clear: both;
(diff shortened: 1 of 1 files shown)


cherewaty:
@JohnC-80 does merging this give you enough confidence for me to merge https://github.com/folio-org/stripes-core/pull/61?

JohnC-80:
Yes, this will work! Thanks for leading the charge here. `.sr-only` is a good general utility class to have anyways, it or something like it should exist in what little global css we have.  I can get a PR together for that in stripes-core (if you haven't done so already.)

MikeTaylor:
Could UIU-298 be related to this? In that issue, things go wrong when the country selector in the Address editor is set to a string that is not a recognised country name. I think the fix is probably for it simply not to allow that.

JohnC-80:
At the end of the day, the typeahead is still just a text field with suggestions. It doesn't do anything to stop the user from just setting 'a' as their country and leaving it at that... if we require a known value, a '<Select>' is a better option there.

MikeTaylor:
It _is_ a `<Select>`, isn't it? See https://github.com/folio-org/stripes-components/blob/master/lib/structures/AddressFieldGroup/AddressEdit/EmbeddedAddressForm.js#L59

JohnC-80:
It defaults to one, yes. The autocomplete is being passed in at UserForm: https://github.com/folio-org/ui-users/blob/master/UserForm.js#L115
The country list that we have will still need to be handed over to the Select somehow (I believe.)

MikeTaylor:
So is there a simple way to tell it "only accept valid countries from the list"?

JohnC-80:
Only way to really do it with the typeahead would be through the validation of the country field... wouldn't be able to submit until the value's a valid one.

MikeTaylor:
Could we not just have an `enforceValidSelection="true"` property on the `<Select>` component?

JohnC-80:
The ```<Select>``` would inherently work - it's either valid or nothing with it because that's the only choices the user has... with the <Autocomplete>, as I mentioned before, it's still just a textfield, only suggesting things and not enforcing them.. they have to type something invalid to start with in order to get a list of valid suggestions, so there's a weird circular relationship there. If we wanted best of both, maybe we could work up to something like a [Select2 ](https://select2.org/dropdown) bound to be a react version out there somewhere...

MikeTaylor:
I'm obviously not following what's going on here at all. (Which is not too surprising, as I've managed to avoid pretty much all the form code until now.) Is the address's country field a `<Select>` or an `<Autocomplete>`? Or is it somehow ... both?

JohnC-80:
Sorry, @cherewaty , we hijacked the heck out of this. Merge away! We do need a shared autocomplete component - so this improvement to could create a good baseline for what we could pull into stripes-components.

cherewaty:
@JohnC-80 could you hit merge? I don't have the permissions to on this repo.

MikeTaylor:
Done.

PR #137, opened 2017-11-13
patron group settings number of users and last updated

wwelling:


wwelling:
depends on https://github.com/folio-org/stripes-components/pull/87 and https://github.com/folio-org/stripes-smart-components/pull/26

wwelling:
The only limitation is that the order. VisibleFields will be first and then additionalFields. I felt this was somewhat out of scope for the feature I was assigned. Would not be a very difficult enhancement later.

wwelling:
@mkuklis yeah, I will can do that in the morning.

PR #138, opened 2017-11-13
Stripes 452

jeremythuff:
Resolves [UIU-135](https://issues.folio.org/browse/UIU-135), [UICHKOUT-7](https://issues.folio.org/browse/UICHKOUT-7), [UICHKOUT-8](https://issues.folio.org/browse/UICHKOUT-8)


Users.js
```diff
@@ -63,6 +64,13 @@ class Users extends React.Component {
   };
 
   static manifest = Object.freeze({
+    query: {
+      initialValue: {
+        search: '',
+        filters: 'active.Active',
+        sort: 'Name',
+      },
+    },
```
MikeTaylor: I'm not too happy about defining this statically. In the old version, the initial URL within the module was defined in the package file: and I see that in your revised stripes-core, that is still the case. Can't we make it that when you first click on the Users link and stripes-core send you to this application (specifically, to `/users/?filters=active.Active&sort=Name`) the value of the `query` resource gets set from that URL? In fact, wouldn't that happen _on its own_, in light of the code as you've written it?

Users.js
```diff
@@ -72,19 +80,56 @@ class Users extends React.Component {
...
-          query: makeQueryFunction(
-            'username=*',
-            'username="$QUERY*" or personal.firstName="$QUERY*" or personal.lastName="$QUERY*" or personal.email="$QUERY*" or barcode="$QUERY*" or id="$QUERY*" or externalSystemId="$QUERY*"',
-            {
-              Status: 'active',
+          query: (...args) => {
+            /*
+              This code is not DRY as it is copied from makeQueryFunction in stripes-components.
+              This is necessary, as makeQueryFunction only referneces query paramaters as a data source.
+              STRIPES-480 is intended to correct this and allow this query function to be replace with a call
```
MikeTaylor: I wonder if it would be better to do STRIPES-480 before merging this. Presumably the result would be a smaller and more pleasant change-set.

lib/SearchAndSort/SearchAndSort.js
```diff
@@ -5,8 +5,9 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import Route from 'react-router-dom/Route';
 import { withRouter } from 'react-router';
+// eslint-disable-next-line import/no-unresolved
```
MikeTaylor: Why the lint-disable?

MikeTaylor: (I know you didn't add it in this PR, but it bothers me.)

lib/SearchAndSort/SearchAndSort.js
```diff
@@ -152,6 +175,12 @@ class SearchAndSort extends React.Component {
       const resultAmount = nextProps.parentResources.records.other.totalRecords;
       this.SRStatus.sendMessage(`Search returned ${resultAmount} result${resultAmount !== 1 ? 's' : ''}`);
     }
+
+    if (this.state.filters.uninitialized && this.props.parentResources.query) {
+      this.setState({
+        filters: initialFilterState(this.props.filterConfig, this.props.parentResources.query.filters),
+      });
+    }
```
MikeTaylor: This seems kind of perverted. What is the reasoning?

package.json
```diff
@@ -15,8 +15,9 @@
     "type": "app",
     "displayName": "Users",
     "route": "/users",
-    "home": "/users?filters=active.Active&sort=Name",
+    "home": "/users",
```
MikeTaylor: Ah, no, this won't do at all. It relates to comment earlier in the file. We really want to specify this in the package file (which leaves open the possibility to configuring it elsewhere later), not hardwire it in code.

settings/permissions/PermissionSetForm.js
```diff
@@ -8,6 +8,7 @@ import Button from '@folio/stripes-components/lib/Button';
 import Paneset from '@folio/stripes-components/lib/Paneset';
 import PaneMenu from '@folio/stripes-components/lib/PaneMenu';
 import IfPermission from '@folio/stripes-components/lib/IfPermission';
+// eslint-disable-next-line import/no-unresolved
```
MikeTaylor: Do not want.

skomorokh:
Ah but now that we have a place for initial values do we even need this separate home setting rather than just using the root route for the app?

MikeTaylor:
> ... now that we have a place for initial values ...

What place? I hope you don't mean "hardwired into the code".

skomorokh:
...it's in the manifest where the resource is initialised, that's a fine place for it. In your package.json example it's no less hardcoded, it's just further away from its context. Or do you mean for it to be possible to override that home value from, eg. stripes.config.js so one can (perhaps via an app store ui) configure the initial state of the ui-users app on a per-tenant basis?

MikeTaylor:
> ...it's in the manifest where the resource is initialised

In other words, in the source code. That is bad.

And, yes, the end-game here is allowing alternative forms of configuration. Like I said in the comment above, "(which leaves open the possibility to configuring it elsewhere later)".

package.json
jeremythuff: I am adding this back, but the thing is, this will do nothing to effect the default values of users when running as an embedded module. Those values will still be derived from the manifest. The values expressed in this query string happen to coincide with values from the manifest, so it seemed to me to obfuscate the behavior, but it is true that for a top level module the values expressed here will be honored over the values expressed in the manifest.

Users.js
jeremythuff: See comment below...

lib/SearchAndSort/SearchAndSort.js
jeremythuff: I might not need to be disabled, as on my machine eslint was throwing similar errors for imports that were not throwing errors on the ci linting. Or then again, maybe that the best time to use such comments, as it prevents false positives for the presence of errors when running liniting locally.

lib/SearchAndSort/SearchAndSort.js
jeremythuff: To be DRY. I am not the author of initialFilterState, and at this point in the code we want to reset the filters to their initial state, whatever that was. This ensures that that happens in the same way it happened initially.

jeremythuff: Or at least I don't think I wrote initialFilterState...

settings/permissions/PermissionSetForm.js
jeremythuff: see comment above...

lib/SearchAndSort/SearchAndSort.js
jeremythuff: It could be that since the fix to stripes connect that this race condition goes away.

lib/SearchAndSort/SearchAndSort.js
MikeTaylor: OK. We have an ongoing issue with ESLint errors from inability to resolve imports from peer-dependencies, which we need to think about separately. Let's ignore this problem for now, and make a separate issue for it elsewhere.

MikeTaylor: (I wish there was a way for me to mark this conversation thread as Resolved.)

Users.js
MikeTaylor: I still feel this is worth fixing. At present, the inclusion of a vast chunk of modified makeQueryFunction overwhelms the rest of this change-set and makes it hard to see what's really changing.

lib/SearchAndSort/SearchAndSort.js
MikeTaylor: Can you please try it? It would be a shame to commit this workaround if we don't need it.

package.json
MikeTaylor: > This will do nothing to effect the default values of users when running as an embedded module.

That is an interesting and important point that I'd not thought of at all. In general, we are going to want different defaults when running as the top-level module and when running embedded. So I suspect the way to go might be to add a separate `InitialParamsWhenRunningAsEmbeddedModule` setting to the package file -- at which point it might also make sense to replace `home` with `InitialParamsWhenRunningAsTopLevelModule`. (I admit these names might need a bit of work.)

But come to think of it, perhaps the more honest thing is to have the invoking module specify the initial parameters that it wants from the embedded module?

Anyway, for now, let's put this back -- so that the package-file setting still determines what our initial settings are when running as a top-level module -- and we'll separately address how best to provide initial parameters when running embedded.

MikeTaylor:
How big is the STRIPES-480 work? If you think you can do it quickly and easily, I think it's worth doing that right before merging this. But if you think it's going to be more than (say) a day's work, then go ahead and merge this in its present form, and we'll return to that.

jeremythuff:
You are right that STRIPES-480 is about a day's worth of work or so, but unfortunately, as I am back on a TAMU sprint, it might be some time before I have that day to spend on it. I am happy to do it, but I am concerned that waiting on this merge will incur merge conflict (and addition work) between now and the time I am able to get to it.

MikeTaylor:
I agree, merging now was the right thing to do. Thanks for humouring me, though!

PR #149, opened 2017-11-20
UIU-305 Add tooltip on disabled renew button

clintbellanger:
Explain to the user why the Renew button is disabled (no loans selected) with a rollover tooltip.
Using plain html title= attribute for tooltips.


MikeTaylor:
On second thoughts, @clintbellanger, I'll sort these out myself, as I need to roll a release now (see UIU-322). But please keep this in mind for future commits. Thanks!

PR #154, opened 2017-11-22
Changlog for v2.11.0 is now complete ...

MikeTaylor:
... but still needs lots of editing to be comprehensible.


MikeTaylor:
Also, this PR fixes the lint errors we found earlier.

PR #187, opened 2017-12-11
replaced controlled vocab with editable list

wwelling:
adding a manifest to a parent component of ControlledVocab causes PUT and DELETE to break

### DIFF
--- a/settings/PatronGroupsSettings.js
+++ b/settings/PatronGroupsSettings.js
@@ -88,6 +88,32 @@ class PatronGroupsSettings extends React.Component {
     }
   }
 
+  onCreateType(type) {
+    console.log('ui-items - settings - onCreateType called', type);
+    this.props.mutator.groups.POST(type);
+  }
+
+  onUpdateType(type) {
+    console.log('ui-items - settings - onUpdateType called', type);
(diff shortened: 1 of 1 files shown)


zburke:
The refactoring you've done here looks fine, and indeed working directly with `<EditableList>` solves the problem of saving and removing PatronGroups, but I'd prefer to get to the bottom of why those actions fail when done via `<ControlledVocab>` since `<ControlledVocab>` is in fact implemented in terms of `<EditableList>`.  

It's my understanding from chatting with @wwelling that if a `manifest` is present on a component containing a `<ControlledVocab>` then the `<ControlledVocab>`'s update and remove actions do not function. @MikeTaylor does anything jump out at you as a likely cause? This smells like an issue with `stripes-connect`, but admittedly I have not done any prospecting on my own yet.

wwelling:
Anyone appose this being merged? If and when ControlledVocab is applicable for PatronGroupSettings we can simple switch back.

zburke:
No objections to merging. Filed https://issues.folio.org/browse/STSMACOM-37 so we don't lose track of the bug.

wwelling:
@zburke thanks

PR #197, opened 2017-12-27
UIU-248 Updated view with resourceShouldRefresh.

mbstacy:
UIU-248 Updated view with resourceShouldRefresh. Tested both with boolean and a function. The function should return a boolean. This allows for custom function on whether or not to refresh dataset.

@MikeTaylor @mkuklis  This is related to pull request with STCON-39. This allow for the detail loan action view to refresh. This code is dependent on STCON-39 pull request.


zburke:
ESLint should be installed as a dependency with each repo. Run `yarn lint` from the root of each repo.

PR #202, opened 2018-01-04
Use default filters for "Reset all" filter button

doytch:
Uses the new reset/clear button functionality introduced in:
 - https://github.com/folio-org/stripes-components/pull/164
 - https://github.com/folio-org/stripes-smart-components/pull/69

Also made the `filterConfig` updating not use hardcoded indices.


MikeTaylor:
See https://issues.folio.org/browse/STSMACOM-42. Not sure what is going on here.

PR #234, opened 2018-01-31
Move translations into translations directory

ttamj:
Depends on:
https://github.com/folio-org/stripes-core/pull/178

And a 2.9.0 release of stripes-core.

Related to STCOR-49


MikeTaylor:
As noted elsewhere (but repeated here on the belt-and-braces principle), please hold off on merging this until I have released stripes-core v2.9.0 -- see https://issues.folio.org/browse/STCOR-138

PR #253, opened 2018-02-26
Using makeQueryFunction in manifest

jeremythuff:
Resolves STRIPES-480.

Depends on [PR 34](https://github.com/folio-org/stripes-connect/pull/34) in stripes-connect and [PR 251](https://github.com/folio-org/stripes-components/pull/251) in stripes-components.

This replaces duplicate code from the makeQueryFunction of stripes components with a call to makeQueryFunction.


jeremythuff:
I think it was '$QUERY' with '?{query}', which should be right. The '%' indicates that the values should come from localResources, whereas '$QUERY' was used to get values from the query string, the ns for that being '?'

MikeTaylor:
That makes sense. In fact (I think?) they should both be the same. The anointed resource `query` gets populated with the parts of the URL query, so that `%{query.THING}` should always have the same value as `?{THING}` -- do you agree?

... if so, we should think about whether to gently deprecate `?{THING}` and make _all_ access via the anointed resource, just as we encourage in code. But that's for another day, and should not hold up _this_ PR.

jeremythuff:
I do agree that by using '?' you end up bypassing the functionality we built into the anointed resource. At this point we could not comprehensively say that '?' can be replaced with '%', since the local resource will only hold query string values if it has been configured to do so. 

another thing to note is that '%' is pretty powerful, as it can access any local resource, not just those designated as a query resource.

But it does seem like we might want to move towards a world where we are never grabbing things from url.

MikeTaylor:
> At this point we could not comprehensively say that '?' can be replaced with '%', since the local resource will only hold query string values if it has been configured to do so.

That's a good point.

Sure, `%` is more powerful than `?`; but I'm cool with its functionality being a superset of the latter. But the configurability of the anointed resource, the fact that it's not always going to be called `query`, is a deal-breaker. We do need to keep `?`, at least for now.

PR #256, opened 2018-02-27
Uiu 306 confirmation message

aditya-matukumalli:


lib/OpenLoans/OpenLoans.js
```diff
@@ -160,7 +196,10 @@ class OpenLoans extends React.Component {
     e.stopPropagation();
 
     if (key.action && this[key.action]) {
-      this[key.action](key.loan, e);
+      const renewedLoan = this.renew(key.loan);
+      Promise.resolve(renewedLoan).then(
+        () => this.showCalloutMessage(key, 1), // 1 represents the number of items checked
+      );
     }
```
zburke: There's a bug here: rather than calling the appropriate handler, `this[key.action](key.loan, e);`, this code always calls `this.renew(key.loan)`. I think lines 199-202 just need to be moved into`renew(loan)`; the other changes are likely OK.

PR #258, opened 2018-02-28
Uiu 306 confirmation message

aditya-matukumalli:


lib/OpenLoans/OpenLoans.js
```diff
@@ -160,7 +195,14 @@ class OpenLoans extends React.Component {
     e.stopPropagation();
 
     if (key.action && this[key.action]) {
-      this[key.action](key.loan, e);
+      switch (key.action) {
```
doytch: Might be cleaner as an `if` but 🤷‍♂️

aditya-matukumalli: Agreed. Will make that change. Thanks.

PR #262, opened 2018-03-05
Show list of results when a filter is selected

MikeTaylor:
Fixes part of UIU-399.

(Also, get rid of the unused and misleading initialValue for the anointed resource.)


zburke:
A side effect of passing `failOnCondition: 2` to `makeQueryFunction` is that the "reset all" button doesn't work quite right any more: if you do a search or filter, and then click 'reset all', the form fields reset but the results list does not change because `makeQueryFunction` returns immediately, no query is performed, and the existing results list stays intact.

MikeTaylor:
Should be fixed when I finally get https://issues.folio.org/browse/UISE-48 done.

PR #270, opened 2018-03-07
let SearchAndSort derive things from packageInfo. Refs STSMACOM-64

zburke:
Refs [STSMACOM-64](https://issues.folio.org/browse/STSMACOM-64)


zburke:
> But is there any reason to increment the patchlevel in the version number of ui-users?

I was going with the principle of "adding things changes things" but you are correct that nothing depends on the apps, unlike the libraries. I'll stop.

PR #278, opened 2018-03-07
don't use deprecated props for Typeahead

zburke:
The prop `name` was deprecated for `<Typeahead>` in favor of `inputProps`, generating a console warning. This update resolves the console warning, and also sets the field's `name` attribute based on `inputProps.name` rather than `label`, which must have been an unnoticed bug.

Refs [UIU-427](https://issues.folio.org/browse/UIU-427).

See also https://github.com/folio-org/stripes-components/pull/267


zburke:
Holding off per @JohnC-80. `<Typeahead>` will likely be replaced soon so there is no real gain in this change. 

See https://github.com/folio-org/stripes-components/pull/267.

PR #280, opened 2018-03-08
UIU-422

rasmuswoelk:
Fixed validating of username in edit user.

**Issue:** https://issues.folio.org/issues/

It needs to return a promise so I changed that. For some reason it wouldn't work with redux-form's own provided helper: `SubmissionError` when doing this: `reject(SubmissionError(error))`. Instead I'm just rejecting with the error like so: `reject(errors)` and that seems to work. See the gif below.

mar-08-2018 17-05-40


rasmuswoelk:
@MikeTaylor I'll update the changelog and merge this in. Thanks!

PR #293, opened 2018-03-13
UIU-298-typeahead-replacement for country field

JohnC-80:
More removal than an addition... the `<AddressFieldGroup>` in stripes-components now uses `<Selection>` by default with its own countries list (copied from the countries list that existed in users.)
As part of this PR, I've removed the Autocomplete component as well as the corresponding component override lines in the `AddressFields` prop in both `lib/UserAdresses/UserAdresses.js` and `lib/EditSections/EditContactInfo/EditContactInfo.js`.

Since `<Selection>` enforces a *known system entity (it's more of a `<select>` than a suggestion dropdown) this should clear up UIU-298 and keep unrecognized countries from being entered.


JohnC-80:
Holding off to catch up tests. Diligence!?!?!

PR #305, opened 2018-03-20
UIU-150: apply filters from URL on init

rjruizes:
https://issues.folio.org/browse/UIU-150

This PR toggles a local resource when filterConfig is loaded, thus triggering a refresh. Refreshes in stripes-connect are triggered when the URL changes or when a local resource changes [[1]](https://github.com/folio-org/stripes-connect/blob/13018b60b930e60e7e0eded193353564081099bf/connect.js#L117).
A fetch is actually performed if a resource's path has changed. The users resource path depends on the `query` resource and the `filterConfig`.

Some more background on how ui-users works with stripes-connect:
> When a user makes a change in SearchAndSort, `this.mutators.query.update()` happens. All resources in the manifest have their `componentWillReceiveProps` method triggered. The `users` resource path depends on `query` (`makeQueryFunction` refers to `resourceValues.query`), so a fetch is called for it. The other resources are not fetched because their path is unchanged.


rjruizes:
Thanks, corrected in https://github.com/folio-org/ui-users/pull/306
```
initializedFilterConfig: { initialValue: false },
```

PR #345, opened 2018-04-10
UIU-393 Filters switching place

rladdusaw:
Resolves UIU-393

[ViewUser.js](https://github.com/folio-org/ui-users/blob/master/ViewUser.js#L108) was not getting a sorted list of Patron Groups, but [Users.js](https://github.com/folio-org/ui-users/blob/master/Users.js#L119) was. This caused the list of filters to switch places depending on whichever was called most recently. 

Resolved by addling identical sorting to ViewUser.js so both requests for Patron Groups would give identical results.

### DIFF
--- a/ViewUser.js
+++ b/ViewUser.js
@@ -105,7 +105,7 @@ class ViewUser extends React.Component {
     },
     patronGroups: {
       type: 'okapi',
-      path: 'groups',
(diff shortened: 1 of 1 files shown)


funkymalc:
@zburke - This PR is several months old.  Can it be closed?

zburke:
I don't know why this PR was never merged, but it was handled instead in #460.

PR #350, opened 2018-04-16
UIU-202: Wire and Translate Users App Into German

julianladisch:
* add all translations to translations/de.json
* add i18n-lint.sh
* remove unused entries in translations/en.json
* remove statistical designation UMI "United States Minor Outlying Islands" from data/countries.js
* convert from FormattedMessage to SafeHTMLMessage where needed


zburke:
Should we consider moving `i18n-lint.sh` to stripes-core so all repos can take advantage of it? It looks quite handy.

MikeTaylor:
> Should we consider moving `i18n-lint.sh` to stripes-core so all repos can take advantage of it? It looks quite handy.

Definitely!

Let's get the present PR merged first. Then we can make a new Jira for moving 18n-link into `stripes-core/util` and renaming it `i18n-lint` (without the `.sh` suffix). From there, we can legitimately hope that it will develop into a very useful tool.

PR #364, opened 2018-04-24
Uiu 461 openloans refinement

aditya-matukumalli:
refs UIU-461


lib/Loans/OpenLoans/OpenLoans.js
```diff
@@ -126,6 +195,15 @@ class OpenLoans extends React.Component {
     return contributorsList;
   }
 
+  fetchLoanPolicy(item) {
```
mkuklis: @aditya-matukumalli do we still need this function?

PR #381, opened 2018-05-04
UIU-202: Replaced hardcoded labels by i18n translation keys

julianladisch:
in ProxyItem.js.


mkuklis:
@julianladisch ah great!

PR #382, opened 2018-05-07
UIU-202: Another review of the German translations.

julianladisch:


julianladisch:
Reviewed by Birgit Gruber: https://issues.folio.org/browse/UIU-202

PR #403, opened 2018-06-06
STRIPES-539 Remove yarn warnings

cherewaty:
https://issues.folio.org/browse/STRIPES-539

Several dependencies weren't getting appropriate `peerDependencies`. The solution: add the lacking ones to `devDependencies`.


zburke:
The downside of adding these version-specific deps back into each repo under `devDependencies` is that we lose the ability to, say, upgrade a shared-dependency (like `"react": "^16.4.0"`) in a _single_ place (`stripes-core/package.json`). 

@doytch, I believe you headed up the React 15 to 16 effort; any thoughts on whether there is a middle ground here?

package.json
```diff
@@ -204,28 +204,33 @@
...
   "devDependencies": {
-    "@folio/eslint-config-stripes": "^1.1.0",
+    "@folio/eslint-config-stripes": "^2.0.0",
+    "@folio/stripes-connect": "^3.1.1",
+    "@folio/stripes-core": "^2.9.0",
+    "@folio/stripes-logger": "^0.0.2",
     "babel-core": "^6.17.0",
     "babel-eslint": "^7.0.0",
-    "babel-preset-es2015": "^6.18.0",
+    "babel-preset-env": "^1.7.0",
```
ttamj: We shouldn't need to maintain any babel dependencies within individual ui modules. These are all managed within stripes-core.  I think these can be removed entirely.

cherewaty: Yep, that's right 👍 Removed.

doytch:
Yeah, what @zburke is true. I'd prefer to keep things as peer dependencies unless we explicitly and actively **want** to support multiple versions of a package. For React, there's really no scenario where that'll be the case, and I'm not really sure I see the usefulness of specifying separate webpack and redux versions.

I mean, even the eslint dependency which is hanging out there from previously feels odd now and I'd think it makes more sense to have it defined just in stripes-core.

cherewaty:
Stale, closing.

PR #417, opened 2018-06-19
Uiu 526 bulk renew loans

aditya-matukumalli:


withRenew.js
```diff
@@ -45,44 +45,49 @@ const withRenew = WrappedComponent =>
...
     hideModal() {
       this.setState({ errors: [] });
     }
 
-    getMessage() {
-      const { errors } = this.state;
-      if (!errors.length) return '';
-      return errors.reduce((msg, error) => ((msg) ? `${msg}, ${error}` : error), '');
+    getMessage(errors) {
+      this.errors = errors;
```
mkuklis: @aditya-matukumalli let's avoid creating an instance var here. The lint will complain about `getMessage` but we can add ignore or just move getMessage out of the component.

lib/Loans/OpenLoans/OpenLoans.js
```diff
@@ -288,6 +269,23 @@ class OpenLoans extends React.Component {
...
+    const q = this.state.loans.map(loan => {
+      return `itemId==${loan.itemId}`;
+    }).join(' or ');
+
+    const query = `(${q}) and status==("Open - Awaiting pickup" or "Open - Not yet filled") sortby requestDate desc`;
+    this.props.mutator.requests.reset();
+    this.props.mutator.requests.GET({ params: { query } }).then((requestRecords) => {
+      let requestCount = 0;
+      const requestCountObject = requestRecords.reduce((map, record) => {
+        map[record.itemId] = map[record.itemId] ? map[record.itemId]++ : ++requestCount;
```
mkuklis: @aditya-matukumalli I'm probably missing something here but what is the purpose of `requestCount`? Why would we want to assign `requestCount` to  `map[record.itemId]`?

lib/BulkRenewalDialog/BulkRenewInfo.js
```diff
@@ -0,0 +1,78 @@
...
+  render() {
+    const { successRenewals, failedRenewals } = this.props;
+    const iconAlignStyle = { display: 'flex', alignItems: 'center' };
+
+    return (
+      <div>
+        <div style={iconAlignStyle}>
+          { failedRenewals.length > 0 ?
+            <span style={iconAlignStyle}>
+              <Icon size="medium" icon="validation-error" color="red" />
```
doytch: Can you use the new `status` prop for `Icon` here instead of `color`? Technically, errors aren't in red but `#900` and I went ahead and did the work of selecting the correct colours from `variables.css` with that `status` prop.

The PR is folio-org/stripes-components#427. You'll probably have to bump your stripes-components dependency version.

lib/BulkRenewalDialog/BulkRenewInfo.js
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  constructor(props) {
+    super(props);
+    this.connectedLoanList = props.stripes.connect(BulkRenewedLoansList);
+  }
+
+  render() {
+    const { successRenewals, failedRenewals } = this.props;
+    const iconAlignStyle = { display: 'flex', alignItems: 'center' };
```
doytch: Since you're already importing `Layout`, could you use the `flex` style it has to avoid inlining styles?

lib/BulkRenewalDialog/BulkRenewalDialog.js
```diff
@@ -0,0 +1,85 @@
...
+
+    if (Object.keys(newState).length) return newState;
+    return null;
+  }
+
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      open: false, // eslint-disable-line react/no-unused-state
```
doytch: I can't tell where you're using this besides setting it in `getDerivedStateFromProps`. Is this a refactor artifact? 🤔

translations/ui-users/en.json
```diff
@@ -3,6 +3,13 @@
   "resultCount": "{count, number} {count, plural, one {Record found} other {Records found}}",
   "closedLoansCount": "{count, number} {count, plural, one {closed loan} other {closed loans}}",
 
+  "brd.renewConfirmation": "Renew Confirmation",
+  "brd.header.renewalStatus": "Renewal status",
+  "brd.header.title": "Title",
+  "brd.failedRenewal": "Item not renewed",
+  "brd.successfulRenewal": "Item successfully renewed",
+  "brd.itemNotRenewed": "<strong>{count}</strong> {count, plural, one {item} other {items}} <b>not</b> renewed. ",
```
doytch: Does this work now? I used to have issues with SafeHTMLMessage rendering `<strong>` tags. I had to switch to `<b>` for everything despite not wanting to 😆

withRenew.js
```diff
@@ -45,44 +45,49 @@ const withRenew = WrappedComponent =>
...
+          const contentType = resp.headers.get('Content-Type');
+          if (contentType && contentType.startsWith('application/json')) {
+            resp.json().then((error) => {
+              const errors = this.handleErrors(error);
+              reject(this.getMessage(errors));
+            });
+          } else {
+            resp.text().then((error) => {
+              reject(error);
+              alert(error); // eslint-disable-line no-alert
```
doytch: IIRC, failed renewals will throw a 422 Unprocessable Entity, or may start throwing them as I've seen the backend starting to standardise around that return code. So a failed renewal would throw up an `alert()` which doesn't seem great, especially since you're already showing failed renewals in the UI, right?

lib/BulkRenewalDialog/BulkRenewedLoansList.js
```diff
@@ -0,0 +1,102 @@
...
+        itemStatus: formatMessage({ id: 'ui-users.loans.columns.itemStatus' }),
+        currentDueDate: formatMessage({ id: 'ui-users.loans.columns.dueDate' }),
+        requestQueue: formatMessage({ id: 'ui-users.loans.details.requests' }),
+        barcode: formatMessage({ id: 'ui-users.information.barcode' }),
+        callNumber: formatMessage({ id: 'ui-users.loans.details.callNumber' }),
+        loanPolicy: formatMessage({ id: 'ui-users.loans.details.loanPolicy' }),
+      }}
+      formatter={{
+        renewalStatus: loan => {
+          if (failedRenewals.filter(loanObject => loanObject.id === loan.id).length > 0) {
```
doytch: If we move the rendering of these Popovers and components into a parent, can we reuse the `LoanList` component? So much of this is similar that it feels off to have something so functionally similar and basically rewrite the whole thing.

doytch: The only other thing I can see is the ordering of failedRenewals and successRenewals but again, the parent could just do that instead.

withRenew.js
aditya-matukumalli: Sure, will do

lib/BulkRenewalDialog/BulkRenewInfo.js
aditya-matukumalli: @doytch Ya, was following that PR. Will do the needful.

lib/BulkRenewalDialog/BulkRenewInfo.js
aditya-matukumalli: @doytch oh! was there one? missed it then. Will do!

lib/BulkRenewalDialog/BulkRenewalDialog.js
aditya-matukumalli: @doytch Yes, must have been a mistake from porting over from one of the CDDD files.

translations/ui-users/en.json
aditya-matukumalli: @doytch Yup, it works. Translations guy @rjruizes helped me with that.

withRenew.js
aditya-matukumalli: @doytch You can find partly relevant discussion here https://github.com/folio-org/ui-checkout/pull/131#issuecomment-393168957. So basically, yes, the 422 validation responses are sent as a JSON responses and they are shown in the modal popups, but the other errors such as statusCode >= 500, are sent as plain text which is shown as an alert box. I am not sure if there has been discussion somewhere where it has been decided otherwise.

lib/BulkRenewalDialog/BulkRenewedLoansList.js
aditya-matukumalli: @doytch Sure, let me look into this

lib/Loans/OpenLoans/OpenLoans.js
aditya-matukumalli: @mkuklis requestCount here represents the number of requests open against individual item. I am storing a map of the itemId and the corresponding number of requests open against them. I guess I could just assign 1 instead of `++requestCount` as  `map[record.itemId]++` should take care of the count after the initial increment. That way I could just get rid of that variable. Does this address your concern?

mkuklis: @aditya-matukumalli yes replacing it with just `1` and getting rid of that var was what I was thinking.

aditya-matukumalli: @mkuklis yeah, added that in. Working on @doytch's change requests

lib/BulkRenewalDialog/BulkRenewedLoansList.js
aditya-matukumalli: @doytch Looks like refactoring would add more complexity than what it is now. Because, in that case, all the properties like `visibleColumns` (the columns are different in both the scenarios), `columnMapping`, `formatter` would need to be sent down as props from the parent. Also, the `LoanList` component has a manifest whereas the `bulkRenewal` doesn't have one (basically a dumb component).  I feel its better to keep it this way as both of them are different and simple enough. Let me know your thoughts.

withRenew.js
doytch: Ahhh weird. Way to sleuth 🔍

translations/ui-users/en.json
doytch: This is good to know! :D

lib/BulkRenewalDialog/BulkRenewedLoansList.js
doytch: Fair nuff! Thanks for checking! 👍

PR #426, opened 2018-06-25
case of search term and corpus must match

zburke:
The previous implementation lower-cased the search term but not the text
being searched, making it impossible to search against any of the words
in a permission's display-name that were capitalized.

Refs [UIORG-76](https://issues.folio.org/browse/UIORG-76)


MikeTaylor:
I'm really glad you found this. I've been aware for a while that something wasn't right with searching in the permissions, but never got a firm handle on what it was.

PR #432, opened 2018-07-03
Uiu effective location

aditya-matukumalli:


aditya-matukumalli:
The location property on the item object of the loan record represents the effectiveLocation. This logic was already in place. That makes this PR redundant.

PR #447, opened 2018-07-31
lint

zburke:
Yabba-dabba-doo!

Lintstones, meet the Lintstones
They'll clean up your repository
Stopping all your code rot
They're give you some consistency

Let's put the braces on one line
Let's do the same things all the time

When you're with the Lintstones
Have a yabba-dabba-do-time
The same-for-you-time
We'll have a gay old time!


zburke:
The codebase has moved ahead of these changes and they are no longer relevant. RIP, Fred and Barney.

PR #449, opened 2018-08-01
Disable log statement- it was messing with .on().

cledvina:


MikeTaylor:
Why did you close instead of merging?

PR #451, opened 2018-08-01
UIU-546: Assign Service Points to users

doytch:
[UIU-546](https://issues.folio.org/browse/UIU-546): Added ability to CRUD the service points a user can work from and their preferred service point.

Would appreciate feedback on the following two points.

**Double Fetch**
I was hoping to avoid having to fetching `service-points` in `ViewUser` and `EditServicePoints`, but that would need me to start passing `ViewUser.props.resources` down through `UserForm` to (for consistency) `UserForm`'s children (which includes `EditServicePoints`). I _can_ do this if we want. 

Could also merge `ViewUser`'s `resources` into the `parentResources` that already gets passed down. Thoughts?

**Preferred Service Point of "None"**
The "None" option for preferred service points is a bit sticky. The story requirements are that the `Select` has no default value but a value must be manually selected and validated against, and that None is an option that can be selected. The [implementation of the field](https://github.com/folio-org/mod-inventory-storage/pull/133/files#diff-a928212a5778cd39db6dcd04e066a475R18) allows for undefined or null entries, but values must be a UUID.

Storing `null` and treating it separately from `undefined` doesn't really work since when we're later _retrieving_ the user's data, the JSON transformation at the backend ignores `null` values coming out of the entry. 

Since users need to have a defined a preferred service point (per the story and validation checks), I've set the the `None` selection to have a value of `-` and then applied logic to convert this to `undefined` before POST/PUTting and converting `undefined` to `-` after GETting.

I don't love this, but I couldn't figure out a better way of accomplishing this. I think we're seeing or may see varieties of this issue pop up here and there so it'd be nice to have a good, solid pattern for them.


mkuklis:
@doytch this looks very nice. Please see my comments below.

**Double Fetch**
It would be nice to avoid double fetching. Perhaps we could just merge `service-points` into  `parentResources`? In general I feel it's worth passing couple extra params over fetching the same data twice. What do you think?

**Preferred Service Point of "None"**
Could we use an empty string `''` instead of `-` (I apologize if I'm missing something here).

**ViewUser**
It looks like the `ViewUser` component is getting bigger. Perhaps it would be possible to extract the service points work into a separate HOC in a similar fashion to what has been done with proxies and renewals. What do you think?

doytch:
@mkuklis thanks for the ideas!

- Yeah, I was leaning towards merging it into `parentResources` and like it more now that you're seconding it. :) Just wanted to make sure I wasn't missing anything.
- I thought about an empty string but the issue is that it'll need the same translation process at the edge of getting from/sending to the backend, and it increases the risk of a comparison like  `!preferredServicePoint` (which is how the validation checks generally look like) catching it and complaining that the value is unset. I didn't see any benefit to it over a token like `-` that won't get slurped up by a lazy equality check.
- It is. I think part of the issue is that `ViewUser` is handling all this form-y stuff. It took a while to get my bearings in this code since I kept seeing work that I expected to be done in `UserForm` inside `ViewUser`. I think the HOC is a good suggestion. Personally I'd also like to see all the form-y updating things moved into a middle component, like `ViewUser` > `EditUser` > `UserForm` but perhaps that's a discussion for later?

mkuklis:
@doytch sounds good about `parentResources` and `-`. 
I like the `ViewUser` > `EditUser` > `UserForm`! That's something we can discuss later for sure.

doytch:
Refactored into HOC w/ @mkuklis's suggestion. Definitely cleaner 👌

PR #459, opened 2018-08-10
Test New Perms: Added waitUntilNetworkIdle before deletion check

doytch:
The deletion of the permission set is not instantaneous but is only reflected in the UI once the `DELETE` call returns successfully. Frequently, the subsequent `evaluate()` call would fire before the `DELETE` call returned, erroneously indicating a failed deletion.

### DIFF
--- a/test/ui-testing/new_permission_set.js
+++ b/test/ui-testing/new_permission_set.js
@@ -87,7 +87,7 @@ module.exports.test = function foo(uiTestCtx) {
           .click('a[href^="/settings/users/groups"]')
           .wait('a[href="/settings/users/perms"]')
           .click('a[href="/settings/users/perms"]')
-          .wait(222)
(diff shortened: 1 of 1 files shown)


MikeTaylor:
Why can't we wait for the relevant UI element to update, rather then waiting for the network to become idle?

doytch:
@MikeTaylor We could. I just figured it was clearer and simpler what we were doing this way. I don't really feel strongly though about one approach or another though.

MikeTaylor:
Well, you've done way more work on testing than I have, so I should butt out really. But my feeling is that it's more honest to wait for the specific thing we want to test than to wait for a proxy for it.

doytch:
Yeah I don't mind one way or another. I see your point, but the counter in my head was that it makes it clearer _why_ we're waiting in this particular manner. But I guess the actions we're doing leading up to it kinda help with that.

PR #463, opened 2018-08-13
Test changes proposed in FOLIO-1344. 

funkymalc:


funkymalc:
The major change here is that module integration tests are re-enabled however they are invoked via stripes-cli in "app context" mode.   The platform bundle is not pre-built.

PR #466, opened 2018-08-14
UIU-585: Updated required interfaces

doytch:
[UIU-585](https://issues.folio.org/browse/UIU-585): We're using `configuration` to get information about whether to show profile pictures.

While I was in the code I also noticed we weren't announcing our dependency on the interface for loan-policy-storage that we use to view a users loans and change their due date and renew them. So I included that in the `package.json` and wrapped the affected components in an `IfInterface` as well.

### DIFF
--- a/ViewUser.js
+++ b/ViewUser.js
@@ -543,15 +543,17 @@ class ViewUser extends React.Component {
 
         <IfPermission perm="circulation.loans.collection.get">
           <IfInterface name="circulation" version="3.0">
-            <this.connectedUserLoans
-              onClickViewLoanActionsHistory={this.onClickViewLoanActionsHistory}
-              onClickViewOpenLoans={this.onClickViewOpenLoans}
-              onClickViewClosedLoans={this.onClickViewClosedLoans}
-              expanded={this.state.sections.loansSection}
-              onToggle={this.handleSectionToggle}
-              accordionId="loansSection"
-              {...this.props}
-            />
+            <IfInterface name="loan-policy-storage" version="1.0">
+              <this.connectedUserLoans
+                onClickViewLoanActionsHistory={this.onClickViewLoanActionsHistory}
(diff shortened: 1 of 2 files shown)


MikeTaylor:
I wonder if there's a way to automatically determine when a UI module makes use of an Okapi interface without declaring it? If all network access goes through stripes-connect -- as it presently does -- it should be doable. stripes-core would assemble an exhaustive list of all the WSAPI endpoints defined by the declared interfaces, and tell stripes-connect about them. It could then warn on the console whenever it's asked to access a non-registered path.

Not sure how that would work in GraphQL-world, though.

CC @skomorokh, who usually has an opinion on this kind of thing.

doytch:
I was thinking about that as I was taking this on @MikeTaylor. Even a static analysis tool would be helpful, IMO. Could run it manually or as part of CI.

MikeTaylor:
In fact, we could even call stripes-connect at the end of the CI run and get it tell us which of the registered endpoints was _not_ used at least once.

doytch:
Wouldn't we need 100% code coverage for that to be truthful, though?

MikeTaylor:
Yeah, sure. It could only be a hint to the implementer, so you could think "Hmm, I wonder whether we actually use that interface at all?" -- a prod to investigate further.

skomorokh:
LOL yes, I do have opinions about this kind of thing: 
* we shouldn't worry about stripes-connect and focus on thinking of improvements in terms of graphql
* dependency resolution should be transpile-time not runtime eg. the bundle should be rebuilt when tenant configuration changes
* IfInterface is only  if your dependency on an interface is _optional_, if it's mandatory you should expect it and just bail with an error when the interface isn't found, same as if it suddenly went offline partway through the session

MikeTaylor:
Ah, you're a couple of steps ahead of @skomorokh and me!

PR #484, opened 2018-09-04
remove prepublishOnly npm script

funkymalc:
The npm script, prepublishOnly, fails.   It assumes there is a stripes-core checkout one directory-level higher than this module's root directory.   Really bad assumption.


MikeTaylor:
Yep, this can go. It was scaffolding for early work that led ultimately to the way our CI generates module-descriptors for UI modules, and is no longer in use. (Should have been removed long ago.)

PR #489, opened 2018-09-07
[WIP] STRIPES-547 stripes framework example

ttamj:
**For reference only.  Do not merge.**

This demonstrates the simplest implementation of stripes-framework in which the dependencies brought in by `stripes` are defined as peer dependencies within `ui-users`.


cherewaty:
Love where this is going 👏

PR #496, opened 2018-09-10
Clean up dependencies

cherewaty:
Removes unnecessary `babel` packages, and adds needed `devDependencies`.


zburke:
Agreed. I suspect this is the result of updating eslint-config-stripes from 2.0.3 to 3.2.0. Bringing the code into compliance with the new config is A Good Thing, don't get me wrong, but perhaps best handled in a commit independent of the other dependency changes.

PR #504, opened 2018-09-13
Release v2.15.0

MikeTaylor:
(Second try at a PR for this. The first one was rejected because of a testing failure that seems to be nothing to do with ui-users -- some kind of transitory problem with Ansible.)


MikeTaylor:
Well, bugger me -- seems GitHub is too sneaky for me: it's realised that the present PR is identical to the previous one, and tossed the same Jenkins results into it. I'll have to try deleting the branch, reinstating it, and _then_ making a third identical PR.

MikeTaylor:
Bugger. Same thing.

PR #516, opened 2018-09-20
avoid limit=0 in queries

zburke:
Due to a regression in RAML from 0.8 to 1.0, limit=0 in queries results
in an error. While it should be syntactically valid, the simplest thing
for the release is to use limit=1 instead of re-releasing a bajillion
server-side modules.

Refs [FOLIO-1517](https://issues.folio.org/browse/FOLIO-1517)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change history for ui-users
 
+## 2.15.3 (IN PROGRESS)
+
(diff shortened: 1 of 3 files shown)


zburke:
Agreed, @MikeTaylor, which is why it's only being patched on the release branch, and then only if @adamdickmeiss tells me it's the most expedient thing to do. It's not going into master.

MikeTaylor:
Gotcha!

zburke:
Not going to patch the Q3 branch at this late stage.

PR #517, opened 2018-09-20
Clean up user edit form layout

cherewaty:
## Purpose 
The user edit form doesn't do the flexing it needs to for different screen widths.

## Approach
Set a max-width of `50em` for the entire form. Adjust `react-flexbox-grid` usage as necessary.

## Tips
1. Don't use `react-flexbox-grid` to set a faux max-width. Set the max-width on the container, so the `<Row>` fills the container.
2. Don't use `react-flexbox-grid` when you don't need to.
```
<Row>
  <Col xs={12}>
    Content
  </Col>
</Row>
```
does exactly the same thing as 
```
<div>
  Content
</div>
```

## Next Steps
- `react-flexbox-grid` is effective when you can guarantee the parent container is 100% of the width of the browser. Some day we might want this form to be renderable in the third pane of a `SearchAndSort`-esque layout. Then we'll want to write a little more custom CSS in this repo instead of using `react-flexbox-grid`.
- Four inputs per row is a lot. UX researchers are pretty consistent in advocating for single-column forms: https://baymard.com/blog/avoid-multi-column-forms

### Before
folio-q3 aws indexdata com_users_view_05afe1c7-efcb-4312-b0d5-7076cf695875_layer edit query e ipad

folio-q3 aws indexdata com_users_view_05afe1c7-efcb-4312-b0d5-7076cf695875_layer edit query e ipad 1

### After
localhost_3000_users_view_05afe1c7-efcb-4312-b0d5-7076cf695875_layer edit query e ipad 1

localhost_3000_users_view_05afe1c7-efcb-4312-b0d5-7076cf695875_layer edit query e ipad

### DIFF
--- a/src/components/EditSections/EditServicePoints/EditServicePoints.js
+++ b/src/components/EditSections/EditServicePoints/EditServicePoints.js
@@ -174,14 +174,10 @@ class EditServicePoints extends React.Component {
             onToggle={this.props.onToggle}
             displayWhenClosed={<Badge>{(this.userServicePoints && this.userServicePoints.length) || 0}</Badge>}
           >
-            <Row>
-              <Col xs={8}>
-                { this.renderAddServicePointButton() }
-                { this.renderPreferredServicePointSelect() }
-                { this.renderServicePoints() }
-                { this.renderAddServicePointModal() }
-              </Col>
-            </Row>
+            <div>{ this.renderAddServicePointButton() }</div>
+            { this.renderPreferredServicePointSelect() }
+            { this.renderServicePoints() }
+            { this.renderAddServicePointModal() }
           </Accordion>
         </IfInterface>
       </IfPermission>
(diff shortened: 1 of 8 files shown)


src/components/EditSections/EditServicePoints/EditServicePoints.js
```diff
@@ -174,14 +174,10 @@ class EditServicePoints extends React.Component {
...
           >
-            <Row>
-              <Col xs={8}>
-                { this.renderAddServicePointButton() }
-                { this.renderPreferredServicePointSelect() }
-                { this.renderServicePoints() }
-                { this.renderAddServicePointModal() }
-              </Col>
-            </Row>
+            <div>{ this.renderAddServicePointButton() }</div>
```
wwilsman: Before, this whole section was had a full-width wrapper. I'm assuming these other methods render their own wrapper, or they don't need it at all?

cherewaty:
Double-checked service points section. All good.

screen shot 2018-10-01 at 12 08 09 pm

PR #519, opened 2018-09-21
Added app icon to the first column of the results list

rasmuswoelk:
Added AppIcon to the first column of the results list to make it match the design of this prototype:
https://ux.folio.org/universal-navigation/uxi-3/

### DIFF
--- a/Users.js
+++ b/Users.js
@@ -3,6 +3,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 
 import { makeQueryFunction, SearchAndSort } from '@folio/stripes-smart-components';
+import AppIcon from '@folio/stripes-components/lib/AppIcon';
 
 import uuid from 'uuid';
 import ViewUser from './ViewUser';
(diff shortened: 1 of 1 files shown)


cherewaty:
What design problem are we trying to solve by adding this icon to every item in the search results list? I'm unsure of the rationale.

rasmuswoelk:
By having icons in these lists, users will quickly be able to scan what kind of data they are looking at without having to further read the content of the row.

While it seems redundant to have icons in the list in an app like the "Users"-app when there only is one kind of content, it does have a great advantage in apps like "Inventory" or "Search" that contains multiple types of content in the various lists.

I think the rationale for adding it here is the consistency aspect as well as recognizability. This is a design pattern that will be repeated all over the application whenever there is some kind of data list.

rasmuswoelk:
I see your points but you'll have to take it up with @filipjakobsen.

It's a small change and it can easily be changed back. I'm merging now. We can revisit later if needed 👍

PR #522, opened 2018-09-24
Update number of tags on the user's view

mkuklis:
This PR adds a new `shouldRefresh` function from https://github.com/folio-org/stripes-connect/pull/62  to `selUser` resource in order to control when the user should be refreshed.

### DIFF
--- a/ViewUser.js
+++ b/ViewUser.js
@@ -47,8 +47,7 @@ class ViewUser extends React.Component {
       clear: false,
       shouldRefresh: (resource, action) => {
         const { path } = action.meta;
-        const options = resource.optionsTemplate;
-        return path.match(/link/) || (options.path || '').startsWith(path);
+        return path.match(/link/);
       },
     },
     loansHistory: {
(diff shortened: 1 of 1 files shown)


ViewUser.js
```diff
@@ -45,6 +45,10 @@ class ViewUser extends React.Component {
       type: 'okapi',
       path: 'users/:{id}',
       clear: false,
+      shouldRefresh: (resource, action) => {
```
zburke: I get that this _does_ work, but I'm not entirely sure how. What does the `action` object typically look like? If we're refreshing `selUser`, why not also `loansHistory` and `permissions`, which also include `:{id}` in their paths?

mkuklis: Thanks for looking into this @zburke! 

The action is being built here:

https://github.com/folio-org/stripes-connect/blob/master/epics/mutations.js#L19

and we only really care about the `action.meta` (I may change it to just pass that to `shouldRefresh` instead of a full action object).

the `meta` contains some basic info from the resource which executed the mutation (currently `path` and `name`).

The match by path currently only checks the beginning of the string with `startsWith` so the `permissions` and `loansHistory` won't match even though they include `:{id}`.

PR #538, opened 2018-10-09
Uiu 570

evaluk:
Updated branch by merging master.  Key changes are in test/ui-testing/test.js and other test files.  Merge was massive due to project progress since branch was created.


evaluk:
I noticed in my merging of this branch that there was a new file for a new set of tests that is not currently called.  I left it as is in my merge, since I do not know the status of this test set.

PR #541, opened 2018-10-16
Load proxies/sponsors on initial load. Fixes UIU-661.

mkuklis:
https://issues.folio.org/browse/UIU-661


zburke:
Ah; so we were only loading things on-update, not on-mount?

mkuklis:
@zburke yes I think this was some sort of regression because I'm sure we used `componentDidMount` in the past but then it disappeared for some reason.

PR #559, opened 2018-11-02
STSMACOM-143: Update user record buttons

MaksymDryha:
According to [this story](https://issues.folio.org/browse/STSMACOM-143), "Add [...]" buttons for Proxy/Sponsor, and User permissions should be aligned to the left so that the label text and position for this type of button follows the latest general FOLIO style guidelines:

image


src/components/ProxyGroup/ProxyEditList/ProxyEditList.js
```diff
@@ -114,14 +114,25 @@ export default class ProxyEditList extends React.Component {
...
-            <Layout className="right">
+            <Layout>
               <Pluggable
                 aria-haspopup="true"
                 type="find-user"
                 {...this.props}
                 dataKey={name}
-                searchLabel={intl.formatMessage({ id: 'stripes-components.addNew' })}
+                searchLabel={name === 'proxies' ?
+                  intl.formatMessage({ id: 'ui-users.sponsor.addSponsor' }) :
```
Godlevskyi: Please use \<FormattedMessage\>

PR #563, opened 2018-11-09
include requests accordion

zburke:
The "create new request" button isn't working correctly, but the other features are in place. 

Refs [UIU-677](https://issues.folio.org/browse/UIU-677)


zburke:
Geez if this isn't the messiest failed attempt at a rebase, I don't know what it. Let's just start all over and try again.

PR #574, opened 2018-11-13
UIU-679: Users App: Use documented react-intl patterns instead of stripes.intl

Godlevskyi:
Updated this repo to only use react-intl constructs outlined at 
https://github.com/folio-org/stripes-core/blob/master/doc/i18n.md

### DIFF
--- a/src/LoanActionsHistory.js
+++ b/src/LoanActionsHistory.js
@@ -290,14 +290,16 @@ class LoanActionsHistory extends React.Component {
 
   showContributors(list, listTodisplay, contributorsLength) {
     this.list = list;
+    const contributorsList = <KeyValue
+      label={<FormattedMessage id="ui-users.loans.columns.contributors" />}
+      value={listTodisplay}
+    />;
+
     return (contributorsLength >= 77) ?
       (
         <Popover>
           <div data-role="target" style={{ cursor: 'pointer' }}>
-            <KeyValue
-              label={<FormattedMessage id="ui-users.loans.columns.contributors" />}
-              value={listTodisplay}
-            />
+            {contributorsList}
           </div>
           <div data-role="popover">
             {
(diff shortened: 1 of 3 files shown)


src/LoanActionsHistory.js
```diff
@@ -432,11 +518,11 @@ class LoanActionsHistory extends React.Component {
               formatter={loanActionsFormatter}
               visibleColumns={['actionDate', 'action', 'dueDate', 'itemStatus', 'Source']}
               columnMapping={{
```
cherewaty: I think `columnMapping` right now has to receive strings - see https://github.com/folio-org/ui-data-import/pull/20 and https://issues.folio.org/browse/STCOM-399

Godlevskyi: Oh.. I see to calculate the width;

src/AccountActionsHistory.js
```diff
@@ -237,58 +250,109 @@ class AccountActionsHistory extends React.Component {
 
           <Row>
             <Col xs={1.5}>
-              <KeyValue label={this.props.stripes.intl.formatMessage({ id: 'ui-users.details.field.feetype' })} value={_.get(account, ['feeFineType'], '-')} />
+              <KeyValue
+                label={<FormattedMessage id="ui-users.details.field.feetype" />}
+                value={_.get(account, ['feeFineType'], '-')}
```
aliaksei-chumakou: maybe it makes sense to import only `get` from `lodash/get`

VictorSoroka1: @aliaksei-chumakou I think it is not an issue anymore. @ttamj, could you please confirm that `wepback` plugin will be added to tree-shake `lodash` dependencies?

ttamj: @VictorSoroka1, yes that is correct.

src/settings/WaiveSettings.js
```diff
@@ -17,21 +21,26 @@ class WaiveSettings extends React.Component {
...
     return (
       <this.connectedControlledVocab
-        stripes={this.props.stripes}
+        stripes={stripes}
         validate={(item, index, items) => validate(item, index, items, 'nameReason', label)}
         baseUrl="waives"
-        records="waiver"
-        label={this.props.stripes.intl.formatMessage({ id: 'ui-users.waives.label' })}
-        labelSingular={this.props.stripes.intl.formatMessage({ id: 'ui-users.waives.singular' })}
+        records="waives"
```
Yurii-Danylenko: Please, replace `waives` with `waiver`

PR #586, opened 2018-11-20
refine csv export logic, refs STUTL-1

aditya-matukumalli:


src/components/Loans/ClosedLoans/ClosedLoans.js
```diff
@@ -73,6 +73,23 @@ class ClosedLoans extends React.Component {
...
+    // Map to pass into exportCsv
+    this.columnHeadersMap = this.headers.map(item => {
+      return {
+        label: stripes.intl.formatMessage({ id: `ui-users.${item}` }),
+        value: item
+      };
+    });
 
     this.sortMap = {
       [stripes.intl.formatMessage({ id: 'ui-users.loans.columns.title' })]: loan => _.get(loan, ['item', 'title']),
```
Godlevskyi: No need for the key to be internationalized. 
We need to add this column internationalization to columnMapping

src/components/Loans/ClosedLoans/ClosedLoans.js
```diff
@@ -73,6 +73,23 @@ class ClosedLoans extends React.Component {
     this.handleOptionsChange = this.handleOptionsChange.bind(this);
     this.getFeeFine = this.getFeeFine.bind(this);
     this.anonymizeLoans = this.anonymizeLoans.bind(this);
+    this.buildRecords = this.buildRecords.bind(this);
+
+    this.headers = ['action', 'dueDate', 'loanDate', 'returnDate', 'systemReturnDate', 'item.barcode', 'item.callNumber', 'item.contributors',
+      'item.holdingsRecordId', 'item.instanceId', 'item.status.name', 'item.title', 'item.materialType.name',
+      'item.location.name', 'metaData.createdByUserId', 'metadata.updatedDate', 'metadata.updatedByUserId', 'loanPolicyId'];
+
+    this.headers = ['action', 'dueDate', 'loanDate', 'returnDate', 'systemReturnDate', 'item.barcode', 'item.callNumber', 'item.contributors',
```
Godlevskyi: This seems to duplicate lines 78-80

src/components/Loans/ClosedLoans/ClosedLoans.js
aditya-matukumalli: @Godlevskyi  We need the key to be internationalized here as we need the value from the translations file to be the label for our csv columns here. Not sure if you are referring to MCL when you say columnMapping here? as we use json2csv module for the csv export and it would need a label, value pair as can be seen  in this example: https://www.npmjs.com/package/json2csv#example-3

src/components/Loans/ClosedLoans/ClosedLoans.js
aditya-matukumalli: Thanks, made the change

src/components/Loans/ClosedLoans/ClosedLoans.js
```diff
@@ -73,6 +73,23 @@ class ClosedLoans extends React.Component {
...
+      'item.holdingsRecordId', 'item.instanceId', 'item.status.name', 'item.title', 'item.materialType.name',
+      'item.location.name', 'metaData.createdByUserId', 'metadata.updatedDate', 'metadata.updatedByUserId', 'loanPolicyId'];
+
+    // Map to pass into exportCsv
+    this.columnHeadersMap = this.headers.map(item => {
+      return {
+        label: stripes.intl.formatMessage({ id: `ui-users.${item}` }),
```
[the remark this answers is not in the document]
aditya-matukumalli: @Godlevskyi Made the changes to use props.intl from injectIntl over stripes.intl. Thanks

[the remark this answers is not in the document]
Godlevskyi: Thanks!

src/components/Loans/ClosedLoans/ClosedLoans.js
Godlevskyi: Thanks!

src/components/Loans/ClosedLoans/ClosedLoans.js
Godlevskyi: I thing it we do not need to internathionalize the key values as we have all the translations for the MCL in columnMapping. you can move it from the render up and use it for the header lables
```
    const columnMapping = {
       'title': this.props.stripes.intl.formatMessage({ id: 'ui-users.loans.columns.title' }),
       'barcode': this.props.stripes.intl.formatMessage({ id: 'ui-users.loans.columns.barcode' }),
       'Fee/Fine': this.props.stripes.intl.formatMessage({ id: 'ui-users.loans.columns.feefine' }),
       'loanDate': this.props.stripes.intl.formatMessage({ id: 'ui-users.loans.columns.loanDate' }),
       'dueDate': this.props.stripes.intl.formatMessage({ id: 'ui-users.loans.columns.dueDate' }),
       'returnDate': this.props.stripes.intl.formatMessage({ id: 'ui-users.loans.columns.returnDate' }),
       'renewals': this.props.stripes.intl.formatMessage({ id: 'ui-users.loans.columns.renewals' }),
       'Call Number': this.props.stripes.intl.formatMessage({ id: 'ui-users.loans.details.callNumber' }),
       'Contributors': this.props.stripes.intl.formatMessage({ id: 'ui-users.loans.columns.contributors' }),
     };
```
Can you use it for the csv headers also, like: columnMapping['title']?

aditya-matukumalli: @Godlevskyi Then thing here is the headers that export uses don't match with the ones in columnMapping. We have more headers in the export than in the columnMapping.(FYI: we have 16 columns in export vs 9 in columnMapping). But, yes we can definitely move the columnMapping into the constructor and add in all the additional translations there and use something like columnMapping['title']. But, I am not sure if that would make the code look cleaner? That would mean almost 20 to 25 lines of translations for the columnMapping variable.
Is that what you are suggesting here?

src/components/Loans/ClosedLoans/ClosedLoans.js
```diff
@@ -302,6 +330,10 @@ class ClosedLoans extends React.Component {
     const anonymizeString = <FormattedMessage id="ui-users.anonymize" />;
     const loansSorted = _.orderBy(loans,
       [this.sortMap[sortOrder[0]], this.sortMap[sortOrder[1]]], sortDirection);
+    const columnHeadersMap = this.columnHeadersMap;
+    const clonedLoans = JSON.parse(JSON.stringify(this.props.loans)); // Do not mutate the actual resource
+    const recordsToCSV = this.buildRecords(clonedLoans);
+    const onlyFields = columnHeadersMap;
```
VictorSoroka1:
```
const columnHeadersMap = this.columnHeadersMap;
const onlyFields = columnHeadersMap;
```

seems like excess for me as `this.columnHeadersMap` can be passed straight away to the component.

```
<ExportCsv data={recordsToCSV} onlyFields={this.columnHeadersMap} />
```

src/components/Loans/ClosedLoans/ClosedLoans.js
```diff
@@ -147,6 +160,21 @@ class ClosedLoans extends React.Component {
     return contributorsList;
   }
 
+  buildRecords(recordsLoaded) {
```
VictorSoroka1: Seems like the `buildRecords` method is exactly the same as the one in `OpenLoans.js` file. Could you please extract in order and use in both files in order to avoid duplication?

translations/ui-users/en.json
```diff
@@ -269,6 +269,26 @@
...
+  "item.status.name": "Item status",
+  "item.title": "Item title",
+  "item.materialType.name": "Material type",
+  "item.location.name": "location",
+  "metaData.createdByUserId": "createdByUserId",
+  "metadata.updatedDate": "Updated date",
+  "metadata.updatedByUserId": "updatedByUserId",
+  "loanPolicyId": "Loan policy Id",
+
+
```
VictorSoroka1: Please remove excess empty line.

src/components/Loans/ClosedLoans/ClosedLoans.js
Godlevskyi: @aditya-matukumalli, thanks for the clarification regarding the export.
But my point was more about the MCL implementation. 
I do not think that it is correct to have translations in the sortMap and sortOrder, as we have mappings in the columnMapping.

Sorry, this comment was not related to the current PR but more to the component itself.
Thanks

src/ViewUser.js
```diff
@@ -262,6 +263,21 @@ class ViewUser extends React.Component {
     });
   }
 
+  buildRecords(recordsLoaded) {
```
Yurii-Danylenko: Please, avoid a mutation of the input props. This makes the code difficult to debug. Also, I'd like to propose a solution for your consideration:
```
recordsLoaded.map((record) => {
  const {
    item,
    item: {contributors},
  } = record;

  if (isArray(contributors)) {
    return {
      ...record,
      item: {
        ...item,
        contributors: contributors
          .map((currentContributor) => currentContributor.name)
          .join('; ')
      }
    };
  }
})
```

src/components/Loans/ClosedLoans/ClosedLoans.js
```diff
@@ -324,7 +337,7 @@ class ClosedLoans extends React.Component {
               >
                 {anonymizeString}
               </Button>
-              <ExportCsv data={loans} excludeKeys={['id', 'userId', 'itemId']} />
+              <ExportCsv data={recordsToCSV} onlyFields={this.columnHeadersMap} />
```
Yurii-Danylenko: Could you probably place props on separate lines?

src/components/Loans/ClosedLoans/ClosedLoans.js
```diff
@@ -71,6 +72,17 @@ class ClosedLoans extends React.Component {
...
     this.getFeeFine = this.getFeeFine.bind(this);
     this.anonymizeLoans = this.anonymizeLoans.bind(this);
+    this.headers = ['action', 'dueDate', 'loanDate', 'returnDate', 'systemReturnDate', 'item.barcode', 'item.callNumber', 'item.contributors',
+      'item.holdingsRecordId', 'item.instanceId', 'item.status.name', 'item.title', 'item.materialType.name',
+      'item.location.name', 'metaData.createdByUserId', 'metadata.updatedDate', 'metadata.updatedByUserId', 'loanPolicyId'];
+
+    // Map to pass into exportCsv
+    this.columnHeadersMap = this.headers.map(item => {
+      return {
+        label: this.props.intl.formatMessage({ id: `ui-users.${item}` }),
```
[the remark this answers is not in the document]
aditya-matukumalli: @Yurii-Danylenko As we require a plain string there for the label, the only way to do that is using intl? The last line here: https://github.com/folio-org/stripes/blob/master/doc/i18n.md#tldr

src/ViewUser.js
aditya-matukumalli: @Yurii-Danylenko Thats a great solution. Added in the new logic

src/components/Loans/ClosedLoans/ClosedLoans.js
aditya-matukumalli: @Yurii-Danylenko Done

src/components/Loans/OpenLoans/OpenLoans.js
```diff
@@ -686,7 +698,8 @@ class OpenLoans extends React.Component {
     const bulkActionsTooltip = <FormattedMessage id="ui-users.bulkActions.tooltip" />;
     const renewString = <FormattedMessage id="ui-users.renew" />;
     const changeDueDateString = <FormattedMessage id="stripes-smart-components.cddd.changeDueDate" />;
-
+    const clonedLoans = JSON.parse(JSON.stringify(this.props.loans)); // Do not mutate the actual resource
```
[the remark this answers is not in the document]
aditya-matukumalli: @Yurii-Danylenko Done

src/components/Loans/ClosedLoans/ClosedLoans.js
```diff
@@ -302,7 +314,8 @@ class ClosedLoans extends React.Component {
     const anonymizeString = <FormattedMessage id="ui-users.anonymize" />;
     const loansSorted = _.orderBy(loans,
       [this.sortMap[sortOrder[0]], this.sortMap[sortOrder[1]]], sortDirection);
-
+    const clonedLoans = JSON.parse(JSON.stringify(this.props.loans)); // Do not mutate the actual resource
```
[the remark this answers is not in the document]
aditya-matukumalli: @Yurii-Danylenko Done

src/components/Loans/ClosedLoans/ClosedLoans.js
aditya-matukumalli: @Godlevskyi I think its better if I file a new JIRA to remove the translations from sortMap and sortOrder and possibly other places and work on it. Thanks

Godlevskyi: @aditya-matukumalli Great! thank you!

PR #594, opened 2018-11-27
don't use FormattedMessage for options

zburke:
The label for an `<option>` must be a string so we need to use
`formatMessage()` instead of `<FormattedMessage>`. This reverts that
portion of the work in #574.

Additionally, it sets the values for the `<option>`s in the
proxy-related fields to hard-coded strings consistent with the immutable
values on the backend.

Fixes [UIU-736](https://issues.folio.org/browse/UIU-736)

### DIFF
--- a/src/components/ProxyGroup/ProxyEditItem/ProxyEditItem.js
+++ b/src/components/ProxyGroup/ProxyEditItem/ProxyEditItem.js
@@ -2,8 +2,6 @@ import React from 'react';
 import {
   FormattedMessage,
   FormattedTime,
-  injectIntl,
-  intlShape
 } from 'react-intl';
 import moment from 'moment'; // eslint-disable-line import/no-extraneous-dependencies
 import {
@@ -34,7 +32,6 @@ import css from './ProxyEditItem.css';
 class ProxyEditItem extends React.Component {
   static propTypes = {
     index: PropTypes.number,
-    intl: intlShape,
     record: PropTypes.object,
     namespace: PropTypes.string, // sponsors or proxies
     name: PropTypes.string,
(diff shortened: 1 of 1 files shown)


aliaksei-chumakou:
@zburke in ui-orders we use `<FormattedMessage>` for options, you might check it here https://github.com/folio-org/ui-orders/blob/master/src/components/POLine/Vendor/FieldRefNumberType.js
I would like to know your thoughts about this way

src/components/ProxyGroup/ProxyEditItem/ProxyEditItem.js
```diff
@@ -136,43 +139,31 @@ class ProxyEditItem extends React.Component {
...
-    const yesNoValues = [
-      <FormattedMessage id="ui-users.yes" />,
-      <FormattedMessage id="ui-users.no" />,
-    ];
-
-    const relationStatusOptions = relationStatusValues.map(val => ({
-      label: val,
-      value: val,
-      selected: record.proxy && record.proxy.status === val
+    const relationStatusOptions = ['active', 'inactive'].map(option => ({
```
Godlevskyi: @zburke, it is not necessary to inject `intl` for the options of `Select` and we can use `<FormattedMessage>` in the following way:

```
const requestForSponsorOptions = ['yes', 'no'].map(option => (
    <FormattedMessage id={`ui-users.${option}`} >
    {(optionTranslated) => <option value={option}> {optionTranslated}</option>}
    </FormattedMessage>
));
```
and then in the Select we do not need the dataOptions prop:

```
<Field
  disabled={this.state.statusDisabled}
  label={<FormattedMessage id="ui-users.proxy.relationshipStatus" />}
  name={`${name}.proxy.status`}
  component={Select}
  fullWidth
  defaultValue={record.proxy && record.proxy.status}
>
  {requestForSponsorOptions}
</Field>
```

zburke: Good point; thanks for the example. I'll refactor.

PR #596, opened 2018-11-27
I18n-friendly shortcuts

JohnC-80:
Turns out that `[` and `]` aren't so friendly to foreign keyboards since the `alt` key must be pressed to access them. This can also confuse javascript since bracket key will always have the `alt` key pressed (it may assume it's a different, unhandled shortcut.)

### DIFF
--- a/src/UserForm.js
+++ b/src/UserForm.js
@@ -152,12 +152,12 @@ class UserForm extends React.Component {
       {
         name: 'expandAllSections',
         handler: this.expandAllSections,
-        shortcut: 'mod+['
+        shortcut: 'mod+alt+b'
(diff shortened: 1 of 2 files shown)


translations/ui-users/en.json
```diff
@@ -2,6 +2,8 @@
   "meta.title": "Users",
 
   "search": "Search",
+  "expandAll": "Expand All",
+  "collapseAll": "Collapse All",
```
zburke: "Expand all" and "Collapse all"

JohnC-80: Good catch! Thanks!

PR #599, opened 2018-11-29
let strings be strings

zburke:
Some attributes really need to be strings, not `<FormattedMessage>`s. Most of the time you can use a render prop to `<FormattedMessage>` to accomplish that, but not always, e.g. a `<DropdownMenu>` won't render correctly when coded that way so we have to fall back to `injectIntl` and `props.intl.formatMessage()`.

### DIFF
--- a/src/ViewUser.js
+++ b/src/ViewUser.js
@@ -559,25 +559,33 @@ class ViewUser extends React.Component {
       (
         <PaneMenu>
           {
-            tagsEnabled && <IconButton
-              icon="tag"
-              title={<FormattedMessage id="ui-users.showTags" />}
-              id="clickable-show-tags"
-              onClick={this.props.tagsToggle}
-              badgeCount={tags.length}
-              aria-label={<FormattedMessage id="ui-users.showTags" />}
-            />
+            tagsEnabled &&
+            <FormattedMessage id="ui-users.showTags">
+              { ariaLabel => (
+                <IconButton
+                  icon="tag"
+                  id="clickable-show-tags"
+                  onClick={this.props.tagsToggle}
+                  badgeCount={tags.length}
+                  aria-label={ariaLabel}
+                />
+              )}
+            </FormattedMessage>
           }
           <IfPermission perm="users.item.put">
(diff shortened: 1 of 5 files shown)


src/ViewUser.js
```diff
@@ -559,25 +559,33 @@ class ViewUser extends React.Component {
...
-            />
+            tagsEnabled &&
+            <FormattedMessage id="ui-users.showTags">
+              { ariaLabel => (
+                <IconButton
+                  icon="tag"
+                  id="clickable-show-tags"
+                  onClick={this.props.tagsToggle}
+                  badgeCount={tags.length}
+                  aria-label={ariaLabel}
```
Godlevskyi: @zburke, IconButton has an ariaLabel prop which than passed as aria-label to the Button.

src/ViewUser.js
```diff
@@ -559,25 +559,33 @@ class ViewUser extends React.Component {
...
+            <FormattedMessage id="ui-users.crud.editUser">
+              { ariaLabel => (
+                <IconButton
+                  icon="edit"
+                  id="clickable-edituser"
+                  style={{ visibility: !user ? 'hidden' : 'visible' }}
+                  onClick={this.props.onEdit}
+                  href={this.props.editLink}
+                  ref={this.editButton}
+                  aria-label={ariaLabel}
```
Godlevskyi: same, here

src/components/EditablePermissions/EditablePermissions.js
```diff
@@ -164,7 +169,7 @@ class EditablePermissions extends React.Component {
           <DropdownMenu
             data-role="menu"
             width="40em"
-            aria-label={<FormattedMessage id="ui-users.permissions.availablePermissions" />}
+            aria-label={this.props.intl.formatMessage({ id: 'ui-users.permissions.availablePermissions' })}
```
Godlevskyi: @zburke As we already destructure props above, why not to add intl also to destructuring.

zburke:
Prop-names and destructuring are now more consistent.

PR #600, opened 2018-11-29
UIU-691: Apply internationalization to all hardcoded strings for Settings

anastasiaHahanova:
## Purpose
According to [UIU-691](https://issues.folio.org/browse/UIU-691) all the hardcoded strings should be internationalized using the react-intl pattern.
react-intl constructs outlined at [here](https://github.com/folio-org/stripes-core/blob/master/doc/i18n.md) 

## Approach
Use` <FormattedMessage>`, `<FormattedDate>`, and `<FormattedTime>` as often as possible, only falling back to formatMessage(), formatDate(), and formatTime() (accompanied by injectIntl) when a bare string is absolutely necessary.

## Additional
This PR resolve [UIU-640](https://issues.folio.org/browse/UIU-640)

### DIFF
--- a/src/AccountActionsHistory.js
+++ b/src/AccountActionsHistory.js
@@ -3,11 +3,10 @@ import React from 'react';
 import {
   FormattedMessage,
   FormattedTime,
-  intlShape,
-  injectIntl,
 } from 'react-intl';
 import Link from 'react-router-dom/Link';
 import PropTypes from 'prop-types';
+
 import {
   Paneset,
   Pane,
@@ -17,8 +16,32 @@ import {
   KeyValue,
   MultiColumnList,
 } from '@folio/stripes/components';
+
 import { Actions } from './components/Accounts/Actions';
 import { getFullName } from './util';
+
+import css from './AccountsHistory.css';
+
+const columnWidths = {
+  'action': 250,
+  'amount': 100,
+  'balance': 100,
+  'transactioninfo': 200,
+  'created': 100,
+  'source': 200,
+  'comments': 700
+};
+
+const columns = [
+  'date',
+  'action',
+  'amount',
+  'balance',
+  'transactioninfo',
+  'created',
+  'source',
+  'comments',
+];
 
 class AccountActionsHistory extends React.Component {
   static manifest = Object.freeze({
@@ -62,13 +85,12 @@ class AccountActionsHistory extends React.Component {
     onCancel: PropTypes.func.isRequired,
     onClickViewLoanActionsHistory: PropTypes.func.isRequired,
     handleAddRecords: PropTypes.func.isRequired,
-    intl: intlShape.isRequired,
   };
 
   constructor(props) {
     super(props);
     const {
-      stripes : { connect },
+      stripes: { connect },
     } = props;
     this.onSort = this.onSort.bind(this);
     this.onChangeActions = this.onChangeActions.bind(this);
@@ -97,16 +119,7 @@ class AccountActionsHistory extends React.Component {
         comment: false,
         regular: false,
       },
-      sortOrder: [
-        'date',
-        'action',
-        'amount',
-        'balance',
-        'transactioninfo',
-        'created',
-        'source',
-        'comments',
-      ],
+      sortOrder: [...columns],
       sortDirection: ['desc', 'desc']
     };
   }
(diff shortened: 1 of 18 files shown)


src/AccountsHistory.js
```diff
@@ -422,28 +420,40 @@ class AccountsHistory extends React.Component {
 
     const closeMenu = (
       <PaneMenu>
-        <button onClick={this.props.onCancel} type="button">
+        <button
+          onClick={this.props.onCancel}
```
MaksymDryha: onCancel can be destructured

src/ViewUser.js
```diff
@@ -818,7 +823,10 @@ class ViewUser extends React.Component {
               handleAddRecords={this.handleAddRecords}
             />
           </Layer>
-          <Layer isOpen={query.layer ? query.layer === 'account' : false} label="Account Actions History">
+          <Layer
+            isOpen={query.layer ? query.layer === 'account' : false}
```
VictorSoroka1: I see that this logic is repeated. Let's move it the method.

`isLayerOpen('account')`

src/index.js
```diff
@@ -54,10 +55,10 @@ class UsersRouting extends React.Component {
       <CommandList commands={commands}>
         <Switch>
           <Route
-            path={`${this.props.match.path}`}
+            path={this.props.match.path}
```
VictorSoroka1: Please destructure this as well.

src/settings/index.js
```diff
@@ -13,80 +19,105 @@ import PaymentSettings from './PaymentSettings';
...
+        paneTitle={<FormattedMessage id="ui-users.settings.label" />}
+      />
+    );
+  }
+
+  getSections() {
+    return [
+      {
+        label: <FormattedMessage id="ui-users.settings.general" />,
+        pages: _.sortBy(this.getGeneral(), ['label']),
```
VictorSoroka1: Not sure the sorting will continue to work as `label` prop becomes `FormattedMessage` component now instead of string.

src/AccountActionsHistory.js
```diff
@@ -373,22 +440,13 @@ class AccountActionsHistory extends React.Component {
...
-              'action',
-              'amount',
-              'balance',
-              'transactioninfo',
-              'created',
-              'source',
-              'comments',
-            ]}
+            visibleColumns={columns}
             contentData={(account.id === (actions[0] || {}).accountId) ? actionsSort : []}
```
MaksymDryha: I'm not sure that it's a good way of making conditions. Here we are creating an empty object, which obviously won't have the `accountId` property. All further comparisons do not make sense. Maybe it's better to rewrite it like this:
```suggestion
    let contentData = [];
    if (actions[0] && actions[0].accountId === account.id) {
      contentData = actionsSort
    }
```

More lines, but it's easier to read and hos no unneeded steps. What do you think?

src/AccountActionsHistory.js
```diff
@@ -97,16 +119,7 @@ class AccountActionsHistory extends React.Component {
...
-        'date',
-        'action',
-        'amount',
-        'balance',
-        'transactioninfo',
-        'created',
-        'source',
-        'comments',
-      ],
+      sortOrder: [...columns],
```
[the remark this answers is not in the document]
anastasiaHahanova: Firstly, 2 same array have been using for different values. I've removed one, but one of the values (sortOrder) changes this array afterwards.

src/ViewUser.js
```diff
@@ -796,7 +802,7 @@ class ViewUser extends React.Component {
             )}
           </FormattedMessage>
           <Layer
-            isOpen={query.layer ? query.layer === 'open-accounts' || query.layer === 'closed-accounts' || query.layer === 'all-accounts' : false}
+            isOpen={this.isLayerOpen('open-accounts') || this.isLayerOpen('closed-accounts') || this.isLayerOpen('all-accounts')}
```
MaksymDryha: I would suggest placing each of them on a new line

src/ViewUser.js
```diff
@@ -534,6 +534,12 @@ class ViewUser extends React.Component {
     }
   }
 
+  isLayerOpen = value => {
+    const { layer } = this.props.resources.query;
+
+    return layer && layer === value;
```
MaksymDryha: Is there a reason to check if `layer` exists? If it's `undefined`, then `layer === value` returns `false`

PR #607, opened 2018-11-30
Fix renew issue on loan details screen. Part of UIU-686

mkuklis:
This PR refactors action history component in order to address renew issues found in UIU-686.
https://issues.folio.org/browse/UIU-686


mkuklis:
thank you @aditya-matukumalli

PR #609, opened 2018-12-03
Hide request count when request related permissions are not present. Part of UIU-741

mkuklis:
This PR hides request count on the open loan view and loan details view in cases when `ui-users.requests.all` permission is not present.

### DIFF
--- a/src/LoanActionsHistory.js
+++ b/src/LoanActionsHistory.js
@@ -21,6 +21,7 @@ import {
   KeyValue,
   Row,
   Col,
+  IfPermission,
 } from '@folio/stripes/components';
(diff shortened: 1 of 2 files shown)


src/LoanActionsHistory.js
```diff
@@ -21,6 +21,7 @@ import {
   KeyValue,
   Row,
   Col,
+  IfPermission,
```
zburke: `IfPermission` is deprecated in `stripes v1.1` and has moved to `stripes-core`. Otherwise, this looks good.

mkuklis:
Oh crap thanks for catching this @zburke!

PR #613, opened 2018-12-04
Do not use <FormattedMessage ../> for the keys to sort the columns, r…

aditya-matukumalli:
…efs UIU-750

- We are currently using <FormattedMessage .. /> to sort the columns, but currently the logic works with the keys being just strings. This is currently breaking the sorting functionality on the UI.
  Hence, as we already have the mappings defined in the columnMapping, its easier to pull in the mappings from there as the keys for the sortMap and sortOrder.


cherewaty:
Related to https://issues.folio.org/browse/STCOM-399

PR #615, opened 2018-12-05
import IfInterface, IfPermission from core

zburke:
Yahoo! This is your deprecation!
Yahoo! This is your deprecation!
Deprecate stripes things, come on! Let's deprecate!
Deprecate stripes things, come on! LET'S DEPRECATE!

There's an import goin' on right here
Checks interfaces, checks access for yer,
Was in components, it's deprecated from there
We gonna get it now right straight from the core


zburke:
Dupes #597

cherewaty:
🕺

PR #619, opened 2018-12-06
Fix minor errors

DGB-UNAM-FOLIO:
* Fix minor errors of fee/fine history and details. Fixes [UIU-238](https://issues.folio.org/browse/UIU-238), [UIU-239](https://issues.folio.org/browse/UIU-239).

### DIFF
--- a/src/ViewUser.js
+++ b/src/ViewUser.js
@@ -695,7 +695,9 @@ class ViewUser extends React.Component {
 
     if (this.isLayerOpen('add-block') || this.isLayerOpen('edit-block')) {
       return (
-        <Layer label={query.layer === 'add-block' ? 'Add Block' : 'Edit Block'}>
+        <Layer label={query.layer === 'add-block'
+          ? <FormattedMessage id="ui-users.blocks.layer.add" />
+          : <FormattedMessage id="ui-users.blocks.layer.edit" />}>
           <this.connectedPatronBlockLayer
             {...this.props}
             query={query}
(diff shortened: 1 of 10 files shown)


src/ViewUser.js
```diff
@@ -690,7 +690,116 @@ class ViewUser extends React.Component {
         onClickViewAccountActionsHistory={this.onClickViewAccountActionsHistory}
         onClickViewClosedAccounts={this.onClickViewClosedAccounts}
         onClickViewAllAccounts={this.onClickViewAllAccounts}
-      />);
+      />
+    );
+
+    if (this.isLayerOpen('add-block') || this.isLayerOpen('edit-block')) {
+      return (
+        <Layer label={query.layer === 'add-block' ? 'Add Block' : 'Edit Block'}>
```
mkuklis: @DGB-UNAM-FOLIO could you please translate "Add Block" and "Edit Block"?

zburke: Also, I believe the field is `contentLabel` which must be a string, so you probably need to wrap the whole `<Layer>` in a `<FormattedMessage>` with a render prop: ```
<FormattedMessage id="ui-users.blocks.addBlock">
  {contentLabel => (
    <Layer contentLabel={contentLabel} ...> ... </Layer>
  )}
</FormattedMessage>
```

DGB-UNAM-FOLIO:
Thank you very much for the comments, @mkuklis and @zburke, we modified them.

PR #623, opened 2018-12-06
Prohibit adding user as a proxy or sponsor for themselves. Fixes UIU-734.

mkuklis:
https://issues.folio.org/browse/UIU-734


aditya-matukumalli:
Good work! @mkuklis

PR #624, opened 2018-12-07
UUI-678: BigTest initial setup

Godlevskyi:
## Purpose
https://issues.folio.org/browse/UIU-678

We want to add both unit test coverage to the users application, and also add the ability to start the application in mirage mode so that development can be done without having to actually be connected to any kind of server environment.

## Approach

This adds BigTest and Mirage to the users app using the emerging standard bigTest setup of models/serializers/factories going into the tests/bigTest directory.

Endpoints were being setup against an existing API.

In addition to the endpoints, the default scenario was populated with about 20 user instance records 

## Strategy

There is a lot of work to be done in creating the initial stub that lets you achieve the very basic functionality. After that, we can proceed incrementally to implement features of the harness in order to facilitate more comprehensive tests.

As such, the strategy was to stand up just enough of the endpoints to be able to both view the application and write a basic test that loaded a list of users over the network, made sure that all the instances rendered properly, and that clicking on a item in the list loaded the corresponding user record.

In this PR only the _read only_ endpoints for most entities have been implemented.


test/bigtest/network/serializers/service-point.js
```diff
@@ -0,0 +1,19 @@
...
+
+  serialize(...args) {
+    const json = ApplicationSerializer.prototype.serialize.apply(this, args);
+    if (isArray(json.servicePoints)) {
+      return assign({}, { servicepoints: json.servicePoints }, {
+        totalRecords: json.servicePoints.length,
+      });
+    } else {
+      return json.servicePoints;
+    }
```
anastasiaHahanova: `else` can be removed from this statement
```
if() {}

return json.servicePoints;
```

test/bigtest/network/serializers/sevice-points-user.js
```diff
@@ -0,0 +1,19 @@
...
+
+  serialize(...args) {
+    const json = ApplicationSerializer.prototype.serialize.apply(this, args);
+    if (isArray(json.servicePointsUsers)) {
+      return assign({}, json, {
+        totalRecords: json.servicePointsUsers.length,
+      });
+    } else {
+      return json.servicePointsUsers;
+    }
```
anastasiaHahanova: and here

test/bigtest/network/serializers/user.js
```diff
@@ -0,0 +1,24 @@
...
+        totalRecords: json.users.length,
+        resultsInfo: {
+          totalRecords: json.users.length,
+          facets: [],
+          diagnostics: []
+        }
+      });
+    } else {
+      return json.users;
+    }
```
anastasiaHahanova: and here

.stripesclirc.js
```diff
@@ -0,0 +1,46 @@
+const webpack = require('webpack');
+
+const environment = process.env.NODE_ENV;
+
+function mirage(config, enabled = false) {
+  if (enabled) {
+    console.info('Using Mirage Server'); // eslint-disable-line no-console
+    return Object.assign({}, config, {
```
OleksandrAntonenko1:
```suggestion

    return Object.assign({}, config, {
```
Please add empty line after const and before return. Check other places.

aliaksei-chumakou:
@Godlevskyi could you please check comment from Matt? It might be helpful https://github.com/folio-org/ui-orders/pull/35#issuecomment-431482325

karma.conf.js
```diff
@@ -0,0 +1,26 @@
+module.exports = (config) => {
```
aliaksei-chumakou: Are you sure you need this config? I suppose it works with common config from core package

package.json
```diff
@@ -527,19 +527,28 @@
...
   "devDependencies": {
+    "@bigtest/interactor": "^0.9.1",
+    "@bigtest/mirage": "^0.0.1",
+    "@bigtest/mocha": "^0.5.2",
+    "@bigtest/react": "^0.1.2",
     "@folio/eslint-config-stripes": "^3.2.1",
     "@folio/stripes": "^1.1.0",
-    "@folio/stripes-cli": "^1.5.0",
+    "@folio/stripes-cli": "^1.6.0",
+    "@folio/stripes-core": "^2.16.1000442",
```
aliaksei-chumakou: I made the same mistake, you should use stripes core like `^2.16.1` or something, from `npm-folio` registry.
https://github.com/folio-org/ui-orders/pull/35#discussion_r227049002

karma.conf.js
Godlevskyi: Good point! seems it is same. Will remove it

package.json
Godlevskyi: Done

PR #631, opened 2018-12-13
Hide request buttons and disable request links. Fixes UIU-757.

mkuklis:
Hide request buttons and disable request links when user doesn't have permissions. 
https://issues.folio.org/browse/UIU-757

### DIFF
--- a/src/LoanActionsHistory.js
+++ b/src/LoanActionsHistory.js
@@ -354,7 +354,7 @@ class LoanActionsHistory extends React.Component {
 
     const loanPolicyName = get(loanPolicies, 'records[0].name', '-');
     const requestCount = this.state.requestsCount[this.props.loan.itemId] || 0;
-    const requestQueueValue = (requestCount && stripes.hasPerm('ui-users.requests.all'))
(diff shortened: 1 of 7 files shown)


src/LoanActionsHistory.js
```diff
@@ -354,7 +354,7 @@ class LoanActionsHistory extends React.Component {
 
     const loanPolicyName = get(loanPolicies, 'records[0].name', '-');
     const requestCount = this.state.requestsCount[this.props.loan.itemId] || 0;
-    const requestQueueValue = (requestCount && stripes.hasPerm('ui-users.requests.all'))
+    const requestQueueValue = (requestCount && stripes.hasPerm('ui-users.requests.all,ui-requests.all'))
```
zburke: Hmmmmmm. This implicitly makes `ui-users` dependent on `ui-requests`. I guess that's not strictly speaking true since you won't get errors in users without requests installed. Just, er, huh. This causes exactly the right behavior, and it may in fact be the right implementation, but it still makes me cringe a little.

mkuklis: @zburke agree!

PR #638, opened 2018-12-17
UIU-754: Users App: Implement Pane header dropdown on Detail Records

anastasiaHahanova:
## Purpose
According to [UIU-754](https://issues.folio.org/browse/UIU-754), should be possible to cancel / edit user details using pane header dropdown. 

- Actions in 'Create User': Cancel
- Actions in 'Show User': Edit
- Actions in 'Edit User': Cancel

## Approach
- Add actionMenu attribute to `<Pane>` with necessary action.
- Cover actions with tests.

### DIFF
--- a/src/UserForm.js
+++ b/src/UserForm.js
@@ -2,10 +2,12 @@ import cloneDeep from 'lodash/cloneDeep';
 import React from 'react';
 import { FormattedMessage } from 'react-intl';
 import PropTypes from 'prop-types';
+
 import {
   Paneset,
   Pane,
   PaneMenu,
+  Icon,
   IconButton,
   Button,
   ExpandAllButton,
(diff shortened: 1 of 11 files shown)


src/UserForm.js
```diff
@@ -246,11 +253,36 @@ class UserForm extends React.Component {
...
+    } = this.props;
+
     const submitter = handleSubmit(onSubmit);
+
     submitter();
   }
 
+  getActionMenu = ({ onToggle }) => {
+    const { onCancel } = this.props;
+    const handleClick = () => {
```
MaksymDryha: Isn't it better to move this function to a separate method?

src/ViewUser.js
```diff
@@ -869,6 +870,26 @@ class ViewUser extends React.Component {
     );
   }
 
+  getActionMenu = ({ onToggle }) => {
+    const { onEdit } = this.props;
+    const handleClick = () => {
```
MaksymDryha: Same here

src/UserForm.js
anastasiaHahanova: anyway it's impossible to avoid creating of a new function

MaksymDryha: Ok, I see

PR #643, opened 2018-12-19
UIU-780: fix broken closed loans sorting

MaksymDryha:
Currently, translated strings from `columnMapping` are not used in `sortMap` and `sortOrder`. This breaks sorting funtionality


aditya-matukumalli:
Good catch @MaksymDryha

PR #707, opened 2019-02-01
Rely on url changes instead of query resource. Fixes UIU-878

mkuklis:
This PR fixes a bug where unsaved changes modal is not shown after a record is being changed and closed.

Here are couple more details about what is going on:

1. We click on an edit link and the query resource is updated with the `{ layer: edit }` which eventually propagates to `history.push` https://github.com/folio-org/stripes-core/blob/master/src/locationService.js#L74

2. The form is now visible. The stripes-form registers a block: https://github.com/folio-org/stripes-form/blob/master/lib/StripesFormWrapper.js#L25 

3. We make a change on the form and hit close. 

4. The query resource is updated to `{ layer: null }`. This triggers re-render of the `ViewUser`. Since the previous implementation relies on the state of the query resource the `ViewUser` will hide the user form: https://github.com/folio-org/ui-users/blob/master/src/ViewUser.js#L783 and eventually call `componentWillUnmount` on stripes-form: https://github.com/folio-org/stripes-form/blob/master/lib/StripesFormWrapper.js#L38-L42 which causes to unblock the previously registered block in 2) (the unsaved changes modal is now gone).

The change in the PR will rely on URL (browser history) changes instead of query resource.

### DIFF
--- a/src/components/PatronBlock/PatronBlockForm.js
+++ b/src/components/PatronBlock/PatronBlockForm.js
@@ -14,8 +14,9 @@ import {
   TextArea,
   Checkbox,
   Datepicker,
-  AppIcon
 } from '@folio/stripes/components';
+
+import { AppIcon } from '@folio/stripes-core/src/components';
 import { TitleManager } from '@folio/stripes/core';
 import { Field } from 'redux-form';
 import stripesForm from '@folio/stripes/form';
--- a/src/components/PatronBlock/PatronBlockForm.js
+++ b/src/components/PatronBlock/PatronBlockForm.js
@@ -14,9 +14,9 @@ import {
   TextArea,
   Checkbox,
   Datepicker,
+  AppIcon,
 } from '@folio/stripes/components';
 
-import { AppIcon } from '@folio/stripes-core/src/components';
 import { TitleManager } from '@folio/stripes/core';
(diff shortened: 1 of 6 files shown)


zburke:
Man we have _got_ to stop relying on the query string for layers and move to page-based routes.

mkuklis:
Yeah I think the first step should be to make sure we don't depend on query resource for navigation at all. It has to get done as a part of the routing rework but I think it's a smaller change and it would allow us to fix a lot of these form/navigation related issues.

src/components/PatronBlock/PatronBlockForm.js
```diff
@@ -14,8 +14,9 @@ import {
   TextArea,
   Checkbox,
   Datepicker,
-  AppIcon
 } from '@folio/stripes/components';
+
+import { AppIcon } from '@folio/stripes-core/src/components';
```
zburke: That `AppIcon` deprecation warning in stripes-components is _super_ noisy, but I'd rather handle it in a separate PR after we get folio-org/stripes-core/pull/591 merged, assuming it looks correct to you. Then we can import `AppIcon` from `@folio/stripes/core` along with `TitleManager` below. Same for the import in `src/Users.js`. Sound good?

PR #702, opened 2019-01-31
loan-policy-storage interface up to support CIRCSTORE-96

kostyantyn-kh:

PR #709, opened 2019-02-04
Loan policy storage interface up rebase

zburke:
rebasing #702 to nudge Jenkins toward passing tests 🤞


zburke:
Merged #702 instead.

PR #715, opened 2019-02-07
Fix failing users test

mkuklis:
It looks like selecting filter by entering a specific URL doesn't work. The test has been changed and the inactive users filter checkbox is clicked manually now.


zburke:
> It looks like selecting filter by entering a specific URL doesn't work.

Why? It works fine IRL.

mkuklis:
@zburke good question and I don't have the answer :(. I think we need to investigate it more but let's move forward with this one.

PR #748, opened 2019-02-15
test cleanup

zburke:
* define selectors for all user filters in users interactor
* wait for elements before interacting with them in patron-group test
* rename all tests to kebab-case for consistency with platform-core


mkuklis:
@zburke looks great!

PR #767, opened 2019-03-08
Add CRUD Fee/Fine Transfer Accounts

DGB-UNAM-FOLIO:
We are adding:

* Add CRUD Fee/Fine Transfer Accounts Settings. Ref [UIU-544](https://issues.folio.org/browse/UIU-544).

### DIFF
--- a/src/settings/TransferAccountsSettings.js
+++ b/src/settings/TransferAccountsSettings.js
@@ -236,7 +236,7 @@ class TransferAccountsSettings extends React.Component {
             validate={this.validate}
             visibleFields={['accountName', 'desc', 'lastUpdated']}
             columnMapping={{
-              accountName: formatMessage({ id: 'ui-users.transfers.columns.name' }),
+              accountName: formatMessage({ id: 'ui-users.transfers.columns.name' })+'*',
               desc: formatMessage({ id: 'ui-users.transfers.columns.desc' }),
               lastUpdated: formatMessage({ id: 'stripes-smart-components.cv.lastUpdated' })
             }}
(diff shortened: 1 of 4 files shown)


translations/ui-users/en.json
```diff
@@ -827,5 +828,11 @@
   "accounts.actions.warning.waiveAction": "waived",
   "accounts.actions.warning.deselect": "Deselect to continue",
   "accounts.actions.cancellation.success": "<b>{count}</b> {count, plural, one {fee/fine} other {fees/fines}} for <b>{amount}</b> has been successfully <b>{action}</b> for <b>{user}</b>",
-  "accounts.status.outstanding": "Outstanding"
+  "accounts.status.outstanding": "Outstanding",
+  "transfers.label":"Fee/fine: Transfer accounts",
+  "transfers.columns.name":"Name*",
```
zburke: Asterisks to label required fields should not be present in translations. They will be applied automatically by `stripes-components` form elements when given the prop `required`.

src/settings/TransferAccountsSettings.js
```diff
@@ -236,7 +236,7 @@ class TransferAccountsSettings extends React.Component {
             validate={this.validate}
             visibleFields={['accountName', 'desc', 'lastUpdated']}
             columnMapping={{
-              accountName: formatMessage({ id: 'ui-users.transfers.columns.name' }),
+              accountName: formatMessage({ id: 'ui-users.transfers.columns.name' })+'*',
```
zburke: @JohnC-80 interesting question here: how (if at all) should a required field be labeled in the `<EditableList>` component since it's displayed as a column header rather than as a form-field label?

PR #775, opened 2019-03-18
Fix outdated comment

MikeTaylor:
Emtpy list of values on filterConfig for patron-groups claimed that
the list would be filled in by componentWillUpdate. In fact,
componentDidUpdate is responsible for this, and the comment now
reflects that.


MikeTaylor:
/me wipes a tear from his eye.

PR #776, opened 2019-03-20
Fix navigation issue on clicking  View Item Details in Ellipses Menu,…

aditya-matukumalli:
… refs UIU-871

- As we need conditional rendering in case the user does not have the required permission, we need to use the `stripes` object's `hasPerm()` function directly instead of the `<IfPermission>` component in the JSX

### DIFF
--- a/src/components/Loans/OpenLoans/components/ActionsDropdown/ActionsDropdown.js
+++ b/src/components/Loans/OpenLoans/components/ActionsDropdown/ActionsDropdown.js
@@ -11,7 +11,6 @@ import {
   IconButton,
 } from '@folio/stripes/components';
 import {
-  IfPermission,
   stripesShape,
 } from '@folio/stripes/core';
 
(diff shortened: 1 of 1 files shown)


zburke:
I don't follow. `<IfPermission>` is written [in terms of `stripes.hasPermission()`](https://github.com/folio-org/stripes-core/blob/1565db8bd2a14641be18eb7f3097217d8e55169c/src/components/IfPermission/IfPermission.js): 
```
const IfPermission = ({ children, perm }) => (
  <StripesContext.Consumer>
    {stripes => (
      stripes.hasPerm(perm) ? children : null
    )}
  </StripesContext.Consumer>
);
```
Why does this change work?

aditya-matukumalli:
@zburke Should have been more clear with my commit comment there. The culprit here is **<DropDownMenu>** component which only loops through its direct descendants as can be seen here https://github.com/folio-org/stripes-components/blob/master/lib/DropdownMenu/DropdownMenu.js#L83. 
So, **MenuItem** in this case is the child of **<IfPermission>** which doesn't get picked up, and hence using **hasPerm** makes it the direct child of **<DropDownMenu>** and that solves the problem here.

zburke:
OK, that makes sense now, and I understand why this fix works -- I'm impressed you figured it out, in fact! -- but it feels like we could/should also improve upon the `<DropdownMenu>` as the logic there feels fragile.

PR #791, opened 2019-04-08
Eject from settings smart component UIU-895

skomorokh:
As part of the overall effort to reduce routes and other "glue" code in shared components (see STRIPES-589 / FOLIO-1694), we're moving away from the <Settings> smart component.


skomorokh:
{folio-org}/{ui-circulation}#{314} should be incorporated into the `SettingsLayout` component.

PR #810, opened 2019-04-16
UIU-909: Change HasCommand position.

OleksandrAntonenko1:
### Steps to repro

1. On bugfest.folio.ebsco.com, with a user who has permission set folio-admin
2. Go to Check Out app
3. Bring up user
4. Check out an item to user
5. Using action menu, navigate to loan details
6. Exit loan details
7. Exit open loans
### Expected: 
User info pane is wide enough to see key information about user

### Actual: 
The resulting user info pane is overly narrow.
### Additional info
Commands:
* 		collapseAllSections:  "mod+alt+g"
* 		edit:  "mod+alt+e"
* 		expandAllSections:  "mod+alt+b"
* 		new:  "alt+n"
* 		save:  "mod+s"
* 		search:  "mod+alt+h"

### DIFF
--- a/src/ViewUser.js
+++ b/src/ViewUser.js
@@ -938,143 +938,150 @@ class ViewUser extends React.Component {
         onClose={onClose}
         actionMenu={this.getActionMenu}
       >
-        <TitleManager record={getFullName(user)} />
-        <Headline
-          size="xx-large"
-          tag="h2"
+        <HasCommand
+          commands={this.keyboardCommands}
+          display={['new', 'edit', 'search']}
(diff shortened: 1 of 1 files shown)


zburke:
Interesting. Any inclination about why this change works?

OleksandrAntonenko1:
There was a problem with "HasCommand" component especially with "FocusTrap" component(which is a part of stripes-react-hotkeys). It can not pass check presented in the screenshot but putting it into the "Pane" resolves this problem. Cause it will be displayed for sure.
Screen Shot 2019-04-23 at 4 40 20 PM

PR #824, opened 2019-04-26
Test: Comments settings section

DGB-UNAM-FOLIO:
We adding:

Test to Comments section.


mkuklis:
Nice work @DGB-UNAM-FOLIO!

PR #837, opened 2019-05-14
Add ability to search by first and last name. Part of UIU-1068.

mkuklis:
This PR adds ability to search by both first and last name. This is a temporary fix until https://issues.folio.org/browse/RMB-385 is done on the server.

### DIFF
--- a/src/Users.js
+++ b/src/Users.js
@@ -1,4 +1,4 @@
-import { get } from 'lodash';
+import { get, template } from 'lodash';
 import React from 'react';
 import PropTypes from 'prop-types';
 import {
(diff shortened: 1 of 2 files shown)


src/Users.js
```diff
@@ -57,7 +62,8 @@ class Users extends React.Component {
         params: {
           query: makeQueryFunction(
             'cql.allRecords=1',
-            '(username="%{query.query}*" or personal.firstName="%{query.query}*" or personal.lastName="%{query.query}*" or personal.email="%{query.query}*" or barcode="%{query.query}*" or id="%{query.query}*" or externalSystemId="%{query.query}*")',
+            // TODO: Refactor/remove this after work on FOLIO-2066 and RMB-385 is done
+            (parsedQuery, props, localProps) => localProps.query.query.split(' ').map(query => compileQuery({ query })).join(' or '),
```
zburke: I thought we wanted to join clauses with `AND` since the terms within each clause are `ORed`.

mkuklis: @zburke I think you are right here! It will narrow our results. Let me test this first.

PR #839, opened 2019-05-16
UIU-897 Routing refactor

JohnC-80:
Pushing this code up for some public review... will improve on this description with details about the implementation/transition for existing modules.

Refs [UIU-897](https://issues.folio.org/browse/UIU-897)

## Highlights:

- Removed all `<Layer>` instances
- Allowed `react-router` to perform heavy lifting, making use of the `match.params` API to direct things where query-string-checking existed before.
- Moved high-level manifests to 'route' container components within a `routes` folder. These components function as 'data-getters' and pass their data down to presentational 'view' components.
- Workflows (Edit, Loans, Fees-and-Fines, PatronBlocks) all have their own unique routes.
- Reduced the user detail preview to less than 500 lines (vs 1000 + in the old version)
  - All `<Layer/>` changing code removed, state-maintenance and conditional layer rendering removed.
- Used `<SandSQuery>` for the user search view.
- Modified directory structure. Cleared the top level of 'src' except for 'index' and 'commands'. 
  - `src/components` holds `views`, `util`, and `data` since the presentational components therein make the most use of these files.
- A few name changes throughout the components to improve understanding of the code.
- Passing tests.

## Future work
- Push manifests as high as they can go (to the parent route, if possible) so that they aren't buried within the components of the module. This would mean cleaning up the HOC's which do odd things like spreading their wrapped component's manifest (assuming that a manifest exists is a bad thing.)
- Moving to `<Link/>`s over buttons across the board.
- Handling back-links in a better way (history.goBack() for now)
- Saving the place of a user on the search result list for back-tracking situations.
- Updating the hardcoded URL's used within other modules for displaying user details.
- See if `<SearchAndSort>` itself can be made to allow modules to do their own routing.


doytch:
How do all of the uses of `navigationHandlers` look from a DOM POV? Would it be possible to take this opportunity to move some of those buttons etc to `Link`s? Does it even make sense to do so if it is possible?

JohnC-80:
@doytch When I used a `navigationHandler`, it was usually for easier refactor where a function was already being used to mutate the query resource.
I would definitely prefer using Links... and once there's a good way to replace the functional `history.goBack()` with a direct URL, that'd for sure be the way to go...
In a couple of places, the handler that it would use would be determined by some logic (whether there were any loans or any accounts associated with the particular item that was clicked, what the type was, etc...). In one attempt, I did replace it with a `<Link>` and the function was used to get the appropriate href... but then it got really weird when I had to make `history.goBack()` a branch, so now instead of just a `to` prop I'm returning a 'navigation props' object that's either `{ to }` or `{ onClick }`...

src/views/UserDetail/UserDetail.js
```diff
@@ -0,0 +1,556 @@
...
+                />
+                <IfPermission perm="manualblocks.collection.get">
+                  <PatronBlock
+                    accordionId="patronBlocksSection"
+                    user={user}
+                    hasPatronBlocks={(hasPatronBlocks === 1 && totalPatronBlocks > 0)}
+                    patronBlocks={patronBlocks}
+                    expanded={sections.patronBlocksSection}
+                    onToggle={this.handleSectionToggle}
+                    onClickViewPatronBlock={this.onClickViewPatronBlock}
```
VictorSoroka1: There is no longer `onClickViewPatronBlock` method in the file.

src/views/UserDetail/UserDetail.js
```diff
@@ -0,0 +1,556 @@
...
+
+                <IfPermission perm="ui-users.loans.view">
+                  <IfInterface name="loan-policy-storage">
+                    { /* Check without version, so can support either of multiple versions.
+              Replace with specific check when facility for providing
+              multiple versions is available */ }
+                    <IfInterface name="circulation">
+                      <UserLoans
+                        onClickViewLoanActionsHistory={this.onClickViewLoanActionsHistory}
+                        onClickViewOpenLoans={this.onClickViewOpenLoans}
```
VictorSoroka1: There is no longer `onClickViewOpenLoans ` method in the file.

src/views/UserDetail/UserDetail.js
```diff
@@ -0,0 +1,556 @@
...
+                        onClickViewClosedLoans={this.onClickViewClosedLoans}
```
VictorSoroka1: There is no longer `onClickViewClosedLoans ` method in the file.

src/views/UserDetail/UserDetail.js
```diff
@@ -0,0 +1,556 @@
...
+  checkScope = () => {
+    return document.getElementById('ModuleContainer').contains(document.activeElement);
+  };
+
+  // This is a helper function for the "last updated" date element. Since the
+  // date/time is actually set on the server when the record is updated, the
+  // lastUpdated element of the record on the client side might contain a stale
+  // value. If so, this returns a locally stored update date until the data
+  // is refreshed.
+  dateLastUpdated(user) {
```
VictorSoroka1: There are no usages for the function. I wonder if it can be deleted.

src/views/UserDetail/UserDetail.js
```diff
@@ -0,0 +1,556 @@
...
+    this.toggleAllSections(true);
+  }
+
+  collapseAllSections = (e) => {
+    e.preventDefault();
+    this.toggleAllSections(false);
+  }
+
+  // focus management when edit layer closes (refocus edit button)
+  afterCloseEdit = () => {
```
VictorSoroka1: There are no usages for the function. I wonder if it can be deleted.

VictorSoroka1:
Additionally, I found the issue with the user details forms. When a new user is created with minimal setup:
- Last name
- Patron group
- Username
- Email
- Preferred contact

and then upon editing it is unable to add permissions as the submit button is always disabled. I have reproduced it on folio-testing/folio-snapshot.

https://folio-snapshot.aws.indexdata.com/users/preview/e11c4a8b-796a-4b00-9d8c-48eac17961e8

PR #879, opened 2019-06-26
UIU-912: Handle address type validation issue after adding permission…

VictorSoroka1:
… to user

### DIFF
--- a/src/ViewUser.js
+++ b/src/ViewUser.js
@@ -525,12 +525,19 @@ class ViewUser extends React.Component {
 
   update(user) {
     const addressTypes = (this.props.parentResources.addressTypes || {}).records || [];
+    const userData = cloneDeep(user);
 
-    if (user.personal.addresses) {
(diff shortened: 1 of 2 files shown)


src/ViewUser.js
```diff
@@ -525,12 +525,19 @@ class ViewUser extends React.Component {
...
+    const userData = cloneDeep(user);
 
-    if (user.personal.addresses) {
-      user.personal.addresses = toUserAddresses(user.personal.addresses, addressTypes); // eslint-disable-line no-param-reassign
+    if (userData.personal.addresses) {
+      userData.personal.addresses = toUserAddresses(userData.personal.addresses, addressTypes); // eslint-disable-line no-param-reassign
     }
 
-    const { proxies, sponsors, permissions, servicePoints, preferredServicePoint } = user;
+    const {
```
YevheniiMaltsev: It looks like this is a good chance for us to follow our code conventions. Thanks for this update

VictorSoroka1: Agree. Done.

PR #882, opened 2019-07-02
Optimize fee/fine settings

DGB-UNAM-FOLIO:
We are adding:
* Optimize fee/fine settings.


zburke:
Sorry, SonarCloud Code Analysis "Duplicated lines on new code" warning, you're wrong again. I wish you would stop complaining about every component that uses `<ControlledVocab>` and passes an entirely different set of values in. Isn't the whole point of writing reusable components that we want to, you know, REUSE them? Sheesh.

PR #886, opened 2019-07-05
UIU-1092 use new loan-history shape

zburke:
The shape of the response to queries against `loan-storage/loan-history`
has changed. Queries used to look like this:
```
GET loan-storage/loan-history?query=id=$LOAN_ID

{
    "loans" : [ {
        "id" : "$LOAN_ID",
        ...
    } ],
    "totalRecords" : 1
}
```
Then now look like this:
```
GET loan-storage/loan-history?query=loan.id=$LOAN_ID

{
    "loansHistory": [ {
        "id": "f1f72ec5-2e59-4607-8d24-52de0f8e56f9",
        "operation": "I",
        "loan": {
            "id": "$LOAN_ID",
            ...
        }
    } ],
    "totalRecords" : 1
}
```
Note the changes to both the query and the extra level of nesting of the
`loan` element in the response.

Refs [UIU-1092](https://issues.folio.org/browse/UIU-1092)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 * Fix Payment Status from "Charge and pay now". Fixes UIU-931.
 * Fix UX Consistency owners. Fixes UIU-1050.
 * Handle address type validation issue after adding permission to user. Fixes UIU-912.
+* Accomodate changes in the shape of `/loan-storage/loan-history` data. Refs UIU-1092.
 
 ## [2.23.0](https://github.com/folio-org/ui-users/tree/v2.23.0) (2019-06-12)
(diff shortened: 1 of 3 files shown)


zburke:
Argh; all three test failures are not actually failures but timeouts that ultimately resolved, just not within the 2-second limit. I'll attempt rewriting the tests with better `await` conditions so they converge more quickly.

mkuklis:
@zburke I was running into similar issues. I'm trying to address them under https://github.com/folio-org/ui-users/pull/884 maybe it will help here too.

zburke:
@mkuklis, your changes bumping the convergence window from 2000ms to 5000ms did the trick!

PR #895, opened 2019-07-11
address edit enhancement

AlenYao:
Currently, there are three value 'State/Prov./Region', 'city' and 'Zip/Postal Code', now, i want to change the component to 'Select', and implement cascade relationship for the three values, i mean, first select 'Prov', then get 'city' dataOptions, after select 'city', get the 'code' dataOptions, I use onChange() event and already works fine when create action(the case in create a user), but 'city' and 'code' value cannot be display for update action(edit a existing user). I know, the reason is onChange() event is not be triggered, dataOptions for 'city' and 'code' are null when update action

### DIFF
--- a/src/components/EditSections/EditContactInfo/AddressEdit.css
+++ b/src/components/EditSections/EditContactInfo/AddressEdit.css
@@ -1,0 +1,110 @@
+@import '@folio/stripes-components/lib/variables';
+
+.addressEditList {
+  margin-bottom: 1rem;
+}
+
+.legend {
+  display: block;
+  width: 100%;
+  font-size: var(--font-size-large);
+  color: #333;
+  border: 0;
+  margin-bottom: 6px;
+  padding-bottom: 2px;
+}
+
+.subLegend {
+  composes: legend;
+  font-size: var(--font-size-medium);
+  color: #777;
+}
+
+.indent {
+  padding: 0 6px;
+}
+
+.removeAddress {
+  padding: 0;
+  text-align: center;
+  -webkit-appearance: none;
+  cursor: pointer;
+  color: #555;
+  border-radius: 6px;
+  transition: all 0.25s ease;
+  background-color: transparent;
+  border: 1px solid transparent;
+  width: auto;
+  height: auto;
+  margin: 0;
+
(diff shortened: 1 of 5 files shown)


JohnC-80:
@AlenYao The way this is set up, within your `<AddressSel>` component, you may need to access the existing values within context(provided by redux-form): `this.context._reduxForm.values.personal.addresses[this.props.fieldKey].city` etc...

You might be able to avoid keeping the lists in state and the internal `onChange` handlers altogether, if you refer to this context in the `render()` function and grab your list then, passing the variable into the appropriate `<Field>`. The value in `context` will change via `redux-form`'s internal onChange handlers... your component just needs to read those values and adjust accordingly.

You can see what these values are through React Developer Tools (highly recommended browser extension that you should use if you don't already.)

Note that this version of redux-form uses old React context API. Near future, we would like to switch to React-Final-Form in which this will be accessible in a different way.

enettifee:
> Currently, there are three value 'State/Prov./Region', 'city' and 'Zip/Postal Code', now, i want to change the component to 'Select', and implement cascade relationship for the three values, i mean, first select 'Prov', then get 'city' dataOptions, after select 'city', get the 'code' dataOptions, I use onChange() event and already works fine when create action(the case in create a user), but 'city' and 'code' value cannot be display for update action(edit a existing user). I know, the reason is onChange() event is not be triggered, dataOptions for 'city' and 'code' are null when update action

=====

Alen, this could potentially have major effects on the UI - the intention has been that the address fields would remain free text to accommodate a variety of different address types. If I'm understanding you correctly, this would require settings for those fields, and mean that addresses with different formats (like those in many non-US countries) would not be enterable. If you think this is needed, I think it needs to go back to the Users PO (who is now Patty Wanninger - you can find her easily on Slack) for discussion.

zburke:
This is an interesting proof of concept, but given the requirements from the User Management SIG at present, it cannot be merged. 

It may be worth exploring whether address handling could be factored out of users into separate plugins. This would give each tenant the choice of whether to use the current free-text implementation or a different chained-fields implementation such as the one you have developed here.

PR #901, opened 2019-07-17
Increase interface version 'users-bl'

OleksiiMaksymov:
**Purpose**
The interface version was raised due to a change in the API contract in the 'mod-users-bl' module

Resolves: [MODUSERBL-72](https://issues.folio.org/browse/MODUSERBL-72) and [UIU-1099](https://issues.folio.org/browse/UIU-1099)


OleksiiMaksymov:
Changes to the `ui-users` module will need to be merged with the changes: MODUSERBL-72 and UIU-1099

PR #910, opened 2019-07-26
UIU-1134 retrieve 10k permissions

zburke:
Retrieve more permissions. 1k wasn't enough; hopefully 10k is. If not,
640k ought to be enough for everybody said nobody, ever. 

Fixes [UIU-1134](https://issues.folio.org/browse/UIU-1134)


zburke:
Conclusion from a discussion with @marcjohnson-kint is that this is a pargmatic quick fix but we should explore refactoring the UI as well.

PR #918, opened 2019-08-01
UIU-1112: add notes functionality

MaksymDryha:
This PR adds notes functionality to users app.

## Approach
* add notes accordion to ViewUser component
* create route components for creating, editing and viewing notes
* configure mirage to fully support notes functionality
* cover notes functionality with tests


src/util.js
```diff
@@ -38,3 +38,15 @@ export function validate(item, index, items, field, label) {
...
+export function retrieveNoteReferredEntityDataFromLocationState(state) {
+  if (state) {
+    return {
+      name: state.entityName,
+      type: state.entityType,
+      id: state.entityId,
+    };
+  }
+
+  return false;
```
mkuklis: @MaksymDryha could we return `null` here instead?

src/NoteEditPage.js
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  render() {
+    const {
+      match,
+      location: { state },
+    } = this.props;
+
+    const noteId = match.params.id;
+    const referredEntityData = retrieveNoteReferredEntityDataFromLocationState(state);
```
mkuklis: @MaksymDryha it looks like every time you want to use `retrieveNoteReferredEntityDataFromLocationState` you have to first extract state from the location object. Perhaps it would be cleaner to do that inside `retrieveNoteReferredEntityDataFromLocationState` and just pass the `location` as a param to `retrieveNoteReferredEntityDataFromLocationState`?

src/index.js
```diff
@@ -159,6 +162,9 @@ class UsersRouting extends React.Component {
     return (
       <CommandList commands={commands}>
         <Switch>
+          <Route path="/users/notes/new" exact component={NoteCreatePage} />
+          <Route path="/users/notes/:id" exact component={NoteViewPage} />
+          <Route path="/users/notes/:id/edit" exact component={NoteEditPage} />
```
mkuklis: @MaksymDryha this setup looks great!

src/NoteEditPage.js
mkuklis: Hmm maybe it's not a big deal since you still need to extract location from props anyway.

MaksymDryha: @mkuklis, I think it's better to pass just `state`, because if we pass the whole `location` object, then `retrieveNoteReferredEntityDataFromLocationState` will receive a lot of things which it doesn't want to care about

src/util.js
MaksymDryha: @mkuklis, makes sense, done!

package.json
```diff
@@ -645,6 +645,7 @@
   },
   "dependencies": {
     "@folio/react-intl-safe-html": "^1.0.0",
+    "faker": "^4.1.0",
```
VictorSoroka1: Should faker be in dev-deps?

MaksymDryha: @VictorSoroka1, good catch, thank you!

PR #923, opened 2019-08-06
Add fee/fine tasks

DGB-UNAM-FOLIO:
We are adding:

* Finish implementing patron notices. Refs [UIU-1113](https://issues.folio.org/browse/UIU-1113), [UIU-1114](https://issues.folio.org/browse/UIU-1114), [UIU-1115](https://issues.folio.org/browse/UIU-1115).
* Clean up duplicate code. Refs [UIU-932](https://issues.folio.org/browse/UIU-932).


src/components/Accounts/Actions/Actions.js
```diff
@@ -379,22 +377,28 @@ class Actions extends React.Component {
       type.status.name = 'Closed';
     }
     const balance = type.remaining - parseFloat(amount);
-    const comment = values.comment ? (tagStaff + ': ' + values.comment) : '';
+    let c = '';
```
VictorSoroka1: Please rename to `comment`. `c` is not informative naming.

src/components/Accounts/Actions/Actions.js
```diff
@@ -379,22 +377,28 @@ class Actions extends React.Component {
...
+    if (values.comment) {
+      c = tagStaff + ': ' + values.comment;
```
VictorSoroka1:
```suggestion
      comment = `${tagStaff}: ${values.comment}`;
```

src/components/Accounts/Actions/Actions.js
```diff
@@ -379,22 +377,28 @@ class Actions extends React.Component {
...
+    }
+    if (values.patronInfo && values.notify) {
+      c = c + '\n' + tagPatron + ': ' + values.patronInfo;
```
VictorSoroka1: Use [es6 template strings](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) as I used above.

src/components/Accounts/Actions/ActionModal.js
```diff
@@ -0,0 +1,509 @@
...
+  }
+
+  onClose = () => {
+    const {
+      accounts,
+      onClose,
+      reset,
+    } = this.props;
+
+    const selected = calculateSelectedAmount(accounts); // Por que?
```
VictorSoroka1: `// Por que?`
Please write comments in English. As I see this comment is not informative. If it is not needed at all, please remove it.

src/components/Accounts/Actions/ActionModal.js
```diff
@@ -0,0 +1,509 @@
...
+                <Col xs={7}>
+                  <FormattedMessage id="ui-users.accounts.remainingAmount" />
+                  {':'}
+                </Col>
+                <Col xs={4}>
+                  {remaining}
+                </Col>
+              </Row>
+            </Col>
+            {(action === 'payment' && accounts.length > 1) &&
```
VictorSoroka1: Move this check to a method `isPaymentAction(action)` and use it all over the file.

src/components/Accounts/Actions/ActionModal.js
```diff
@@ -0,0 +1,513 @@
...
+  return errors;
+};
+
+const onChange = (values, dispatch, props, prevValues) => {
+  if (values.ownerId !== prevValues.ownerId) {
+    dispatch(change('payment-many-modal', 'method', null));
+  }
+};
+
+const asyncValidate = (values, dispatch, props, blurredField) => {
```
mkuklis: Why is the asyncValidate used here? It doesn't looks like anything is validated asynchronously here?

src/components/Accounts/Actions/ActionModal.js
```diff
@@ -0,0 +1,513 @@
...
+    super(props);
+
+    this.state = {
+      amount: '0.00',
+      showNotify: false,
+      notify: false,
+    };
+  }
+
+  shouldComponentUpdate(nextProps, nextState) {
```
mkuklis: This looks like an anti pattern. Please avoid using `initialize` and also avoid using local state. The state should be already maintained by the `redux-form`. Instead please use `initialValues` when initializing the form from the higher level component. There is also a setting in `redux-form` called `enableReinitialize` which should help here. Also this PR may help more with the setup: https://github.com/folio-org/ui-users/commit/9444e9fede2b8e2c45857cc424122f7d67842a56

In general please avoid using `shouldComponentUpdate` and `componentDidUpdate`.

src/components/Accounts/Actions/ActionModal.js
DGB-UNAM-FOLIO: AsyncValidate is used to round the amount field after the focus is lost.

src/components/Accounts/Actions/ActionModal.js
DGB-UNAM-FOLIO: We made this correction in the code.

src/components/Accounts/Actions/ActionModal.js
mkuklis: @DGB-UNAM-FOLIO please use `onBlur` directly on the amount input field.

zburke: I was thinking the same thing, @mkuklis. Validation functions should not be used to implement side-effects.

src/components/Accounts/Actions/Actions.js
```diff
@@ -438,9 +443,15 @@ class Actions extends React.Component {
       type.status.name = 'Closed';
     }
     const balance = type.remaining - parseFloat(waive);
-    const comment = values.comment ? (tagStaff + ': ' + values.comment) : '';
```
VictorSoroka1: As I see [transfer](https://github.com/folio-org/ui-users/pull/923/files#diff-556cff218755ef7b8f01dba9665e8432R366) method has a lot of duplication stuff as [waive](https://github.com/folio-org/ui-users/pull/923/files#diff-556cff218755ef7b8f01dba9665e8432L428). That is why `commentInfo` stuff is duplicated in both of them. Please come with code without duplication logic in two places.

src/components/Accounts/Actions/ActionModal.js
DGB-UNAM-FOLIO: We had used onBlur, but it didn't work, because of this we handled it this way.

src/components/Accounts/Actions/ActionModal.js
DGB-UNAM-FOLIO: We made the suggested corrections, @mkuklis checked the operation of patron blocks after the changes you added, we realized that it is not working normally, when trying to deactivate the checkboxes, they are not deactivated.
Could you please check this part of the code?

src/components/Accounts/Actions/Actions.js
DGB-UNAM-FOLIO: We add the ``validateComment()`` function, to reduce the duplicate code of ``pay``, ``waive`` and ``transfer`` relative to ``commentInfo``. Does this solve the suggestion you made to us?

src/components/Accounts/Actions/ActionModal.js
mkuklis: Could you describe in more details what didn't work? Did `onBlur` not executed?

src/components/Accounts/Actions/ActionModal.js
mkuklis: Thank you for catching this @DGB-UNAM-FOLIO. This is my fault. I will try to address it today.

mkuklis: Ok @DGB-UNAM-FOLIO this has been fixed here: https://github.com/folio-org/ui-users/pull/927/files

src/components/Accounts/Actions/ActionModal.js
DGB-UNAM-FOLIO: We use ``onBlur``, but it does not perform the operations as we expected.
At first it worked, but one day it stopped working well, so we had to find another way to make it work.

src/components/Accounts/Actions/ActionModal.js
DGB-UNAM-FOLIO: Do not worry @mkuklis! Thank you for reviewing it.

src/components/Accounts/Actions/Actions.js
VictorSoroka1: I suggest using the following approach to get rid of duplicated logic.

```
  transfer = (type, amount, values) => {
    this.props.mutator.activeRecord.update({ id: type.id });

    const paymentMeta = this.preparePaymentMeta({
      paymentMethod: values.account,
      action: 'transfer',
      payment: amount,
      type,
    });

    return this.editAccount(type, paymentMeta.paymentStatus, type.status.name, paymentMeta.balance)
      .then(() => this.newAction(paymentMeta.action, type.id, paymentMeta.paymentStatus, amount, this.validateComment(values), paymentMeta.balance, 0, type.feeFineOwner));
  }

  pay = (type, amount, values) => {
    this.props.mutator.activeRecord.update({ id: type.id });

    const paymentMeta = this.preparePaymentMeta({
      paymentMethod: values.method,
      action: 'pay',
      payment: amount,
      type,
    });

    const owners = _.get(this.props.resources, ['owners', 'records'], []);
    const createdAt = (owners.find(o => o.id === values.ownerId) || {}).owner;
    return this.editAccount(type, paymentMeta.paymentStatus, type.status.name, paymentMeta.balance)
      .then(() => this.newAction(paymentMeta.action, type.id, paymentMeta.paymentStatus, amount, this.validateComment(values), paymentMeta.balance, values.transaction, createdAt || type.feeFineOwner));
  }

  preparePaymentMeta = ({ paymentMethod, payment, type, action }) => {
    const isFullPayment = payment >= type.remaining;

    this.markPaymentAsClosed(type, isFullPayment);

    return {
      action: { paymentMethod },
      paymentStatus: this.getPaymentStatusMessage(action, isFullPayment),
      balance: type.remaining - parseFloat(payment);
    }
  };

  markPaymentAsClosed = (type, isFullPayment) => {
    if (isFullPayment) {
      type.status.name = 'Closed';
    }
  }

  getPaymentStatusMessage = (action, isFullPayment) => {
    const { intl: { formatMessage } } = this.props;
    const paymentStatusBeginning = _.capitalize(formatMessage({ id: `ui-users.accounts.actions.warning.${action}Action` }));

    return isFullPayment
     ? `${paymentStatusBeginning} ${formatMessage({ id: 'ui-users.accounts.status.fully' })}
     : `${paymentStatusBeginning} ${formatMessage({ id: 'ui-users.accounts.status.partially' })}`
  }
```

DGB2018: SonarQube is not reporting duplicity, so what would be the purpose of your suggestion?

VictorSoroka1: Even though Sonar does not if you compare both methods they look almost the same.

PR #925, opened 2019-08-07
STCOM-429: Fixed footer for Save and Cancel buttons

SergiyVSergiyenko:
### Purpose
Implement footer with _Save & close_ and _Canclel_ buttons on the edit user form according to [story](https://issues.folio.org/browse/STCOM-429).

Image

PR #928, opened 2019-08-09
STCOM-429: Prevent disabling user form submit button when form is `inva…

VictorSoroka1:
…lid`

This PR reverts back the previous implementation which was [mistakenly](https://github.com/folio-org/ui-users/pull/925/files#diff-9c19cebbdcaf24b46db12e89c86ed050L231) changed in the scope of fixed footer feature.

### DIFF
--- a/src/UserForm.js
+++ b/src/UserForm.js
@@ -275,11 +275,10 @@ class UserForm extends React.Component {
     const {
       pristine,
       submitting,
-      invalid,
       onCancel,
     } = this.props;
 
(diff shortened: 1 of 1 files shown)


zburke:
OK ... but why? If the form data is invalid, why should submit be enabled?

VictorSoroka1:
The main idea was to revert the previous behavior as was not a part of the story. Moreover, maybe some user wants to hit submit and see all validation errors. I think such changes should be raised as part of another story.

PR #932, opened 2019-08-14
Use stripes.actsAs instead-of deprecated stripes.type

MikeTaylor:


doytch:
> Do we want to replace `type` with `actsAs` or add `actsAs`?

I believe replacing it makes the most sense. And I believe the _existence_ of `type` is really what Stripes CLI complains about.


> Wouldn't the former require a major version bump?

I don't see why it would. Technically, Stripes [continues to provide the `showSettings` prop to modules](https://github.com/folio-org/stripes-core/blob/master/src/components/Settings/Settings.js#L71) in addition to `actAs="settings"`, so there's really nothing else to be done.

Even if it _only_ provided `actAs` now, ui-users would then just update itself to use `actAs` instead of `showSettings` and could avoid bumping its own major version.

zburke:
You're right, @doytch; I was thinking about it inside out.

MikeTaylor:
@doytch is correct. _Replacing_ `type` with `actsAs` is the way to go here, because (1) all remotely recent versions of Stripes understand it as being identical in meaning, and (2) the presence of `type` is the probem we're fixing here, not the absence of `actsAs`.

(BTW., it's `actsAs`, present continuous — not `actAs`, imperative. That has caught me out, too!)

PR #933, opened 2019-08-15
STCOM-429 restore element IDs

zburke:
The element ID for the save button was inadvertently removed in #925. Now
it's back!

Refs [STCOM-429](https://issues.folio.org/browse/STCOM-429)

### DIFF
--- a/src/UserForm.js
+++ b/src/UserForm.js
@@ -285,6 +285,7 @@ class UserForm extends React.Component {
       <PaneFooter>
         <Button
           data-test-user-form-cancel-button
+          id="clickable-cancel"
           buttonStyle="default mega"
           onClick={onCancel}
         >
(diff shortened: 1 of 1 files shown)


VictorSoroka1:
@zburke, is this for integration tests?

zburke:
> is this for integration tests?

Yes. Integration tests run against folio-snapshot which runs in production mode when the `data-test-*` elements have been stripped. 

I've noticed this pattern of removing IDs from buttons when code is refactored. In most cases we could get by without IDs, but selectors that rely on DOM traversal tend to be more brittle, not to mention much longer. Is there a particular reason folks remove them or is it just assumed "nothing here is using the ID so let's toss the extra baggage"?

VictorSoroka1:
> Is there a particular reason folks remove them or is it just assumed "nothing here is using the ID so let's toss the extra baggage"?

I think the second, I think not everyone is aware that ids are used in integration tests, thus remove them since do not find their usages in css. Though we now aware of it and it makes sense for us.

PR #935, opened 2019-08-19
UIU-1129: Fix multiple renewal behaviour.

OleksandrAntonenko1:
# Description
Changed renew errors handling according to the requirements.
## Story 
https://issues.folio.org/browse/UIU-1129


OleksiiMaksymov:
@zburke Yes, that's right when attempting to recall a non-loanable item fails because a non-loanable policy has no due date.
But in the scope of the current task, we have an open recall request on the item and this item in the loan.
After that, we can try to change the loan policy (not loanable, not renewable or change number of renewals) and get the corresponding failures in the renewal.
This means that the loan policy has not been changed at the time of check out of items and creating an open recall request.

About the fail with a 500 it is a bug. I will create a separate story for this.

zburke:
> About the fail with a 500 it is a bug. I will create a separate story for this.

Already filed [CIRC-424](https://issues.folio.org/browse/CIRC-424)

PR #943, opened 2019-08-27
UIU-820 allow proxies with empty expiration dates

zburke:
When checking whether a proxy relationship should be valid, an account
with no expiration date should be considered valid. The previous version
fed the empty value to `moment`, which I think resulted in moment
deciding that an empty date was "before" the current date, which caused
a validation error.

Fixes [UIU-820](https://issues.folio.org/browse/UIU-820)


src/components/ProxyGroup/ProxyEditItem/ProxyEditItem.js
```diff
@@ -108,12 +108,12 @@ class ProxyEditItem extends React.Component {
     let error = '';
 
     // proxy user expired
-    if (moment(proxyRel.user.expirationDate).endOf('day').isSameOrBefore(today)) {
+    if (proxyRel.user.expirationDate && moment(proxyRel.user.expirationDate).endOf('day').isSameOrBefore(today)) {
```
aditya-matukumalli: This would fail I believe with the error _Cannot read property 'expirationDate' of undefined_ if `proxyRel` is `{}`. 

Might be better to use something like lodash `get` here:

`if( get(proxyRel, 'user.expirationDate') && ... )` 

which would result to `undefined` and fail the condition even if `proxyRel` is `{}`

PR #950, opened 2019-09-02
STCOM-571: Make PaneFooter flexible enough

SergiyVSergiyenko:
### Purpose
Update footer with _Save & close_ and _Cancel_ buttons on the edit user form.
Support arbitrary rendering of the content on two sides according the [story](https://issues.folio.org/browse/STCOM-571).

**Screenshots**
View_1
View_2
View_3
View_4
View_5

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 * Show all reasons when a loan renewal fails. Refs UIU-1129.
 * Pass correct props to `<AppIcon>`.
 * Add new permissions modal component. Refs UIU-629, UIU-631.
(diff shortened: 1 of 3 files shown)


zburke:
What none of us caught here was that [STCOM-571](https://issues.folio.org/browse/STCOM-571) fix this PR takes advantage of *did not make it* into the `stripes-components v5.7.0`/`stripes v2.1.0` release. We all must have tested this against the stripes-components tip-of-master/npm-folioci instead of npm-folio. 

I don't know if it's good luck or bad luck that STCOM-571 will be fixed in [`v5.7.1`](https://github.com/folio-org/stripes-components/releases/tag/v5.7.1), but there it is.

PR #954, opened 2019-09-10
UIU-1209, CIRC-200, CIRC-201: Update renewal validation error message.

bohdan-suprun:
In scope of this https://github.com/folio-org/mod-circulation/pull/337 PR we've updated validation error message for item renewal. The reason is following: 

In case we have an item with pending hold requests and use Loan policy with fixed profile. Than we're trying to renew the item and in this case it should behave as rolling renew (as per Sean's comment https://issues.folio.org/browse/CIRC-200?focusedCommentId=59100&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-59100).

CC: @dmytrotur, @marcjohnson-kint.


marcjohnson-kint:
@bohdan-suprun Pull requests are intended to be associated with JIRA issues within the project for that module, in order for to associate the change with version management.

Please raise a JIRA issue for this change and associate any commits / pull requests with that issue

bohdan-suprun:
> @bohdan-suprun Pull requests are intended to be associated with JIRA issues within the project for that module, in order for to associate the change with version management.
> 
> Please raise a JIRA issue for this change and associate any commits / pull requests with that issue

done @marcjohnson-kint UIU-1209 has raised.

PR #961, opened 2019-09-13
Add fix fees/fines

DGB-UNAM-FOLIO:
We are adding:

Fix "Shared" Fee/Fine Owner not allowed to have associated Service Points. Fixes [UIU-1084](https://issues.folio.org/browse/UIU-1084).


src/settings/OwnerSettings.js
```diff
@@ -43,6 +43,16 @@ class OwnerSettings extends React.Component {
     this.connectedControlledVocab = props.stripes.connect(ControlledVocab);
   }
 
+  validate = (item, index, items) => {
```
zburke: Rather confusing to have a function and a method both named `validate`. Maybe call this one `validateItem`, or rename the import?

DGB-UNAM-FOLIO: I understand, we will verify.

src/settings/OwnerSettings.js
```diff
@@ -43,6 +43,16 @@ class OwnerSettings extends React.Component {
...
+    const { intl: { formatMessage } } = this.props;
+    const label = formatMessage({ id: 'ui-users.owners.singular' });
+    const itemErrors = validate(item, index, items, 'owner', label);
+    if (item.owner === 'Shared' && item.servicePointOwner.length > 0) {
```
zburke: This works for editing an existing owner named "Shared" but it'll blow up when creating one because the `item.servicePointOwner` array isn't defined yet: 
```
Uncaught TypeError: Cannot read property 'length' of undefined
    at Object.OwnerSettings._this.validate (OwnerSettings.js:50)
    at ControlledVocab.js:314
    ...
```

DGB-UNAM-FOLIO: I noticed that, we hope to fix soon.

DGB-UNAM-FOLIO: We have made the change.

zburke: Unfortunately, this does not fix the issue. An empty array evaluates to `true` so, for example, it is not possible to change the description for the `Shared` owner, even when the list of Associated service points is empty.

DGB-UNAM-FOLIO: We did the corresponding evaluation of the service point list, it seems to us that this time it does it correctly.

PR #963, opened 2019-09-17
CHAL-30 dedupe loan policy list before retrieval

zburke:
Dedupe the list of loan policies before retrieving it. Previously, a
long list of loans, even if they all shared the same policy, would
generate a long query with one entry per loan. The query now contains
one entry per unique policy.

Refs [CHAL-30](https://issues.folio.org/browse/CHAL-30)


zburke:
Both branch unit tests failures are false positives; the tests were actually successful but exceeded the 2000ms timeout.

PR #964, opened 2019-09-17
fix 'policy' typo

zburke:


src/components/Loans/OpenLoans/OpenLoansControl.js
```diff
@@ -143,7 +143,7 @@ class OpenLoansControl extends React.Component {
     }));
   };
 
-  getLoanPolicie = (policieId) => {
+  getLoanPolicy = (policieId) => {
```
aditya-matukumalli:
```suggestion
  getLoanPolicy = (policyId) => {
```

src/components/Loans/OpenLoans/OpenLoansControl.js
```diff
@@ -143,7 +143,7 @@ class OpenLoansControl extends React.Component {
...
     const { loanPolicies } = this.props;
 
     return loanPolicies[policieId];
```
aditya-matukumalli:
```suggestion
    return loanPolicies[policyId];
```

zburke:
Good catch, @aditya-matukumalli! I made the same policieId -> policyId change in a separate commit, rather than accepting the GitHub suggestion, in order to change both instances atomically.

PR #974, opened 2019-09-25
CHAL-30 retrieve requests-against-patron in batches

zburke:
The structure of the requests API is such that there is no way to
generate a query that retrieves open requests against a user given only
that user's ID. Instead, we must provide the item-IDs of the user's open
loans and retrieve requests based on those item-IDs. For a user with
many loans, this generates a very long querystring, one that is, in
fact, _too_ long, causing Okapi to return a "414 Request URI Too Long"
response.

The approach here is to implement paging, retrieving potential requests
in batches of 50. It's a bit of a hack, but it's not egregious. I picked
50 because I know a user with 100 loans causes the error and this was
halfway. We could probably bisect further and arrive at a higher number
between 50 and 100 but I'm not sure it's worth the effort.

Refs [CHAL-30](https://issues.folio.org/browse/CHAL-30)


src/withRenew.js
```diff
@@ -263,25 +263,32 @@ const withRenew = WrappedComponent => class WithRenewComponent extends React.Com
...
+    // step through the loans list in small batches in order to create a
+    // short-enough query string that we can avoid a "414 Request URI Too Long"
+    // response from Okapi. Details at CHAL-30
+    const step = 50;
+    for (let i = 0; i < loans.length; i += step) {
+      const loansSlice = loans.slice(i, i + step);
+      const q = loansSlice.map(loan => loan.itemId).join(' or ');
+      const query = `(itemId==(${q})) and status==("Open - Awaiting pickup" or "Open - Not yet filled") sortby requestDate desc`;
+      reset();
+      GET({ params: { query } })
```
VictorSoroka1: In the current implementation, it is possible that request from later step will be returned first from the server. If the order does not matter than it is OK.

zburke: The order doesn't matter since we're generating a map from `item.id` to request count. Do you think an implementation with Promise.all() would be cleaner/more clear? This was implemented along the lines of "I wonder if I can get this working in five minutes..." rather than "Let this code be a shining example of Good Design..."

VictorSoroka1: Sounds good.

PR #977, opened 2019-09-25
Fix UIU-1221

DGB-UNAM-FOLIO:
We are adding:

* Fix additional information for patron. Refs [UIU-1221](https://issues.folio.org/browse/UIU-1221).


src/components/Accounts/Actions/Actions.js
```diff
@@ -201,15 +201,28 @@ class Actions extends React.Component {
     return this.props.mutator.accounts.PUT(Object.assign(account, newAccount));
   }
 
+  isTagInfo = (values) => {
```
zburke: Conventionally, a function with the prefix `is` returns a boolean rather than a string, e.g. `isActive(someUser)`, `isError(someResponse)`, `isJohnMalcovich(JohnCusack)`. A good name for this function might be `tagInfo`, `assembleTagInfo`, etc.

PR #984, opened 2019-09-30
Release v2.25.4

zburke:


zburke:
The `continuous-integration/jenkins/branch` check keeps bombing with a unit test that is successful but exceeds its 2000ms timeout window. I'm writing that off as a false positive, slow performance due to a busy CI server rather than a legitimate failure.

PR #992, opened 2019-10-11
UIU-1286: Fix the accumulation of overdue loans in CSV report.

SergiyVSergiyenko:
### Purpose
Fix the mechanism for the accumulation of overdue loans in the CSV report according to the [story](https://issues.folio.org/browse/UIU-1286).

### DIFF
--- a/src/Users.js
+++ b/src/Users.js
@@ -120,13 +120,14 @@ class Users extends React.Component {
     uniquenessValidator: {
       type: 'okapi',
       records: 'users',
-      accumulate: 'true',
+      accumulate: true,
       path: 'users',
       fetch: false,
     },
     loans: {
       type: 'okapi',
       records: 'loans',
+      accumulate: true,
       path: () => `circulation/loans?query=(status="Open" and dueDate < ${getLoansOverdueDate()})`,
       permissionsRequired: 'circulation.loans.collection.get,accounts.collection.get',
     }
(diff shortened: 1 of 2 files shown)


src/Users.js
```diff
@@ -127,6 +127,7 @@ class Users extends React.Component {
     loans: {
       type: 'okapi',
       records: 'loans',
+      accumulate: 'true',
```
VictorSoroka1: This should be just `boolean`.

src/Users.js
```diff
@@ -253,13 +254,23 @@ class Users extends React.Component {
     }
   }
 
+  fetchLoans = props => {
+    const {
+      reset,
+      GET,
+    } = props.mutator.loans;
+
+    reset();
```
VictorSoroka1: I would also add some mechanism to prevent loans fetching when there is one already in progress.

src/Users.js
```diff
@@ -253,13 +254,23 @@ class Users extends React.Component {
     }
   }
 
+  fetchLoans = props => {
```
VictorSoroka1: I think the method is about to generate overdue loans report where fetching of loans is a kind of implementation detail. Please come up with another name.

SergiyVSergiyenko: Thanks for the comment. What about 'loansToReport'?

YevheniiMaltsev: @SergiyVSergiyenko usually the function should contains an action name - call, find, get, etc. So it looks like 'loansToReport' can be improved

PR #993, opened 2019-10-11
UIU-1289: trim whitespaces of username, when userdata is creating

annamelnyk:
### Description
If username contains whitespaces at the beginning or end, it will be trimmed before saving userdata


src/Users.js
```diff
@@ -215,32 +215,35 @@ class Users extends React.Component {
...
-
-      const newUser = await mutator.records.POST(user);
-      const perms = await (value => mutator.perms.POST({ userId: value.id, permissions: [] }))(newUser);
-      const updatedQuery = await (value => {
-        mutator.query.update({ _path: `/users/view/${value.userId}`, layer: null });
-      })(perms);
-      return (newUser, perms, updatedQuery);
-    }
+    username = username.trim();
+    const createdUserdata = {
```
OleksandrAntonenko1: Do we really need this object? Can we use `user` instead?

src/Users.js
```diff
@@ -215,32 +215,35 @@ class Users extends React.Component {
...
+    if (user.creds) delete user.creds;
+    const newUser = await mutator.records.POST(user);
+    const newCreds = await mutator.creds.POST({
+      ...creds,
+      userId: newUser.id
+    });
+    const newPerms = await mutator.perms.POST({
+      userId: newCreds.userId,
+      permissions: []
+    });
```
OleksandrAntonenko1: It might be a good idea to use destructuring here, cause we are using only `id` field. And `userId` is probably is the same during requests.

src/Users.js
```diff
@@ -215,32 +215,35 @@ class Users extends React.Component {
...
+    const updatedQuery = await mutator.query.update({
+      _path: `/users/view/${newPerms.userId}`,
+      layer: null
+    });
+    return updatedQuery;
```
OleksandrAntonenko1: Do we need to return something from this function?

src/Users.js
```diff
@@ -212,29 +212,38 @@ class Users extends React.Component {
...
+      ...createdUserdata.creds,
+      username: createdUserdata.username,
+      password: userdata.creds.password
+    };
+    const user = {
+      ...createdUserdata,
+      id: uuid()
+    };
+    if (user.creds) delete user.creds;
+    const newUser = await mutator.records.POST(user);
```
OleksandrAntonenko1: You should get `POST` from destructuring props. Also, it might be a good idea to rename it during destructuring.

PR #996, opened 2019-10-18
UIU-1083: Implement view loans permission

VictorSoroka1:
## Purpose

Implement view loans permission as part of [UIU-1083](https://issues.folio.org/browse/UIU-1083). This permission is only for view purposes and does not give the user the possibility to edit loans.

## Screenshot

Screenshot


VictorSoroka1:
@zburke, it is not so far, there is a UIU-1177 story just for that. I am wondering if it does make sense to set "visible" to "false" as it is not fully completed yet.

zburke:
Setting `visible = false` sounds like a good approach.

PR #997, opened 2019-10-18
UIU-1175: Finish work on user loans view permission

VictorSoroka1:
## Purpose

Move the showing of fees/fines history under "ui-users.accounts" permission in order for the case when a user has only "ui-users.loans.view" one to make this data read-only (and do not show fees/fines value as a link to the fees/fines history).


src/views/LoanDetails/LoanDetails.js
```diff
@@ -143,15 +143,23 @@ class LoanDetails extends React.Component {
...
+    const { stripes } = this.props;
+
+    if (total === 0) return '-';
+
+    const value = parseFloat(total).toFixed(2);
+
+    return stripes.hasPerm('ui-users.accounts')
+      ? (
+        <button
+          style={{ color: '#2b75bb' }}
```
YevheniiMaltsev: I know, this is not about your changes, but it looks like this is a challenge to support the values like '#2b75bb' in the code

VictorSoroka1: Makes sense. Done.

PR #1011, opened 2019-10-23
UIU-1320 Updating fee/fine column/display after charge

JohnC-80:
## Problem
See [UIU-1320](https://issues.folio.org/browse/UIU-1320)

## Approach
Logic for fetching loans/choosing the selected loan was checking the URL for a loan id in 2 different ways: the query string AND dynamic params picked up via react-router. I opted for the params route and removed the extra query string logic.
The charge fee/fine form should now have what it needs to apply the fine to the loan correctly.
The amounts in the Loans listing/detail will be updated, no refresh/navigation necessary.


VictorSoroka1:
@JohnC-80, thanks for managing this.

PR #1017, opened 2019-10-25
Add proxy behavior

JohnC-80:
# Problem(s)
Proxy edit workflow wasn't really testable.
Adding a single sponsor caused multiple to show up when navigating back to the user detail page upon success.

# Approach
Added facilities to test proxy edit workflow.
Ended up removing the `componentDidMount` method of `withProxy` and adjusting its `cDU` method.

It's possible that loading proxies/sponsors was okay on mount, and the save process needs to be refactored so that we only get transferred to the page without, I believe, an updated resource in tow...

Curious to see if this passes CI tests also...

### DIFF
--- a/test/bigtest/tests/user-proxy-edit-test.js
+++ b/test/bigtest/tests/user-proxy-edit-test.js
@@ -1,0 +1,103 @@
+import {
+  beforeEach,
+  describe,
+  it,
+} from '@bigtest/mocha';
+import {
+  Interactor
+} from '@bigtest/interactor';
+import { expect } from 'chai';
+
+import findUser from '@folio/plugin-find-user';
+
+import setupApplication from '../helpers/setup-application';
+import UserFormPage from '../interactors/user-form-page';
+import InstanceViewPage from '../interactors/user-view-page';
+import UsersInteractor from '../interactors/users';
+import FindUserInteractor from '@folio/plugin-find-user/test/bigtest/interactors/findUser';
+
+describe.only('User Edit: Proxy/Sponsor', () => {
+  setupApplication({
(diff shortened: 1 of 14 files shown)


test/bigtest/tests/user-proxy-edit-test.js
```diff
@@ -0,0 +1,100 @@
+import {
+  beforeEach,
+  describe,
+  it,
+} from '@bigtest/mocha';
+import { expect } from 'chai';
+
+import findUser from '@folio/plugin-find-user';
+import FindUserInteractor from '@folio/plugin-find-user/test/bigtest/interactors/findUser';
```
zburke: I _love_ that you wrote these unit tests for managing proxies ... but I don't think ui-users has a direct dependency on `@folio/plugin-find-user` so I'm not sure about this :/ There is [a user-proxy integration test](https://github.com/folio-org/platform-core/blob/9423464f9720330b0494178f870a86d837d5d52d/test/ui-testing/new-proxy.js) in platform-core that handles proxy management. 

What do you think?

zburke:
The simplification of `<Popdown>` via hooks is 🔥, but maybe best as a separate PR?

JohnC-80:
Yeah, it snuck in due to a tad of laziness...

Re: the plugin - since it's an `optional` dependency, it works. If the `Pluggable` instance that was there had some use-able fallback code, I could do that... this was just the path of least resistance. I'm not completely happy with this.

zburke:
1. Laziness is a virtue.
2. The path of least resistance is a virtuous path.
3. ???
4. Profit!

PR #1019, opened 2019-10-25
UIU-1323 restore overdue loans report

zburke:
This was inadvertently removed in #839.

Fixes [UIU-1323](https://issues.folio.org/browse/UIU-1323)

### DIFF
--- a/src/views/UserSearch/UserSearch.js
+++ b/src/views/UserSearch/UserSearch.js
@@ -58,6 +58,9 @@ class UserSearch extends React.Component {
       records: PropTypes.object,
       patronGroups: PropTypes.object,
     }).isRequired,
+    mutator: PropTypes.shape({
+      loans: PropTypes.object,
+    }).isRequired,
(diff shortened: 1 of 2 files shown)


src/views/UserSearch/UserSearch.js
```diff
@@ -106,13 +110,32 @@ class UserSearch extends React.Component {
...
+
+    if (exportInProgress) {
+      return;
+    }
+
+    this.setState({ exportInProgress: true }, () => {
+      reset();
+      GET()
+        .then(loans => this.overdueLoanReport.toCSV(loans))
+        .then(this.setState({ exportInProgress: false }));
```
VictorSoroka1:
```suggestion
        .then(() => this.setState({ exportInProgress: false }));
```
Please change it the way I suggested. I don't know how it turned out to work just that way.

zburke: 🤦‍♂ Good catch. Thank you!

PR #1020, opened 2019-10-25
UIU-1297 provide 10k limit for overdue loans

zburke:
Previously, there was no `limit` clause on the overdue loans report,
causing it to retrieve only the default row-count of 10. Sadly, there
are likely to be more than 10 overdue loans at any given time. Of
course, I have _never_ returned something past its due date, nor had my
borrowing privileges suspended at libraries in four different states I
don't know why you'd make that suggestion what kind of a person do you
think I am?

Now there's a limit clause of 10k.

Refs [UIU-1297](https://issues.folio.org/browse/UIU-1297)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change history for ui-users
 
+## 2.25.5 (IN PROGRESS)
+
+* Retrieve up to 10k loans in the overdue-loans report. Refs UIU-1297.
+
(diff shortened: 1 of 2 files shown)


zburke:
Note: this PR is against the `v2.25-round2` branch (should've been named `b2.25`; long story; don't ask) so it can be published in the `v2.25` series and deployed as a bug-fix to customers on Q3.2-2019 (daisy). 

#1019 implements the same 10k limit clause for the `master` branch.

zburke:
Updating the query is the right fix but the value in limit clause has not yet been decided.

PR #1027, opened 2019-10-30
UIU-1297: Retrieve up to max amount of overdue loans instead of 10 for CSV report

YevheniiMaltsev:
## Purposes

Retrieve up to max amount (2147483647 according to the [doc](https://s3.amazonaws.com/foliodocs/api/mod-circulation/p/circulation.html#circulation_loans_get)) of overdue loans instead of 10 to add information about all records to the CSV report. The value 2147483647 was selected after discussion with PO and backend devs. Fix [UIU-1297 ](https://issues.folio.org/browse/UIU-1297)

### DIFF
--- a/src/routes/UserSearchContainer.js
+++ b/src/routes/UserSearchContainer.js
@@ -17,6 +17,7 @@ import { UserSearch } from '../views';
 
 const INITIAL_RESULT_COUNT = 30;
 const RESULT_COUNT_INCREMENT = 30;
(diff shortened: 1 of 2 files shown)


src/routes/UserSearchContainer.js
```diff
@@ -17,6 +17,7 @@ import { UserSearch } from '../views';
 
 const INITIAL_RESULT_COUNT = 30;
 const RESULT_COUNT_INCREMENT = 30;
+const MAX_LIMIT = 2147483647;
```
VictorSoroka1: I am thinking about addind comment here in order to make it clear why this exact value was used.

PR #1035, opened 2019-11-04
UIU-820 sync expired-status with expired-date

zburke:
If the relationship between a proxy and sponsor has an expiration date
in the past, the relationship status must be set to inactive.

Refs [UIU-820](https://issues.folio.org/browse/UIU-820)


zburke:
@JohnC-80, I'm seeing values from the date-picker that I don't expect: given a proxy-sponsor relationship, setting an expiration-date shows that date to be one day earlier than expected, e.g. selecting the expiration date of 2019-12-01 logs 2019-11-30 in the console. Do you think there is a bug here in ui-users' (mis)use of `DatePicker` or maybe there is a bug in `DatePicker` itself?

zburke:
The problem in the previous implementation was `moment` taking a midnight-UTC timestamp and converting it to the day before for display in the local locale (midnight on Monday a.m. in London is 7 p.m. on Sunday in New York). We just want to compare dates here, so now that's what we do.

zburke:
I rather hate the fact that I had to move the proxy/status validation code from `ProxyEditItem` all the up to the top-level form. Unfortunately, without a warning function analogous to `setSubmitFailed()`, I couldn't figure out any way to cause a change on one field (e.g. the proxy-relationship-expiration field) to cause a warning on another field (e.g. the proxy status field). 

I'm open to suggestions if you have any.

PR #1050, opened 2019-11-14
UIU-820 better current-user-expired messaging

zburke:
The warning message displayed on the proxy-relationship-status field
when the current user record is expired was inadequate. Now, it clearly
states that the source of the warning is the current user.

Refs [UIU-820](https://issues.folio.org/browse/UIU-820)

### DIFF
--- a/src/views/UserEdit/UserForm.js
+++ b/src/views/UserEdit/UserForm.js
@@ -108,14 +108,14 @@ function getProxySponsorWarning(values, namespace, index) {
   const today = moment().endOf('day');
   let warning = '';
 
-  // proxy user expired
+  // proxy/sponsor user expired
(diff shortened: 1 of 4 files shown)


zburke:
> I may be overthinking this, but is there any potential ambiguity in the message? I.e., any difference in interpretation of " 'Current user' record expired" vs "Current 'user record' expired"?

@Baroquem, on the ticket, I asked if we need "record" in these error messages but did not get a response. 

Personally, I don't think there is any ambiguity we need to worry about.  I can sort of see where you are going with this, i.e. is the record itself expired (as though it came from a cache) vs does the record indicate the user's account has expired. We never reveal the abstraction of the-thing-itself vs the-thing's-record, however, so I'm not worried about the language here tripping somebody up.

PR #1053, opened 2019-11-18
UIU-1380 don't search items by barcode wildcard

zburke:
Searching for items by barcode wildcard is not performant on a
production system with millions of entries. The intent here is that a
barcode will be scanned in so there is really no need for a wildcard
search anyway; this is essentially "lookup by barcode" not "search by
barcode" and the query now reflects that.

Refs [UIU-1380](https://issues.folio.org/browse/UIU-1380)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 * Add service points to user. Refs UIU-1334.
 * Warn, but do not invalidate, when a proxy relationship is expired for any reason. Refs UIU-820.
 * Show requests information when user has view loans permission only. Refs UIU-1184.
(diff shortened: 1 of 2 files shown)


VictorSoroka1:
@zburke, will query 0923 will find the barcode 092353392?

zburke:
> @zburke, will query 0923 find the barcode 092353392?

No. This changes what was effectively a search to an exact match, which is the desired behavior in this circumstance. In fact, ["barcode search" in ui-inventory is implemented the same way](https://github.com/folio-org/ui-inventory/blob/ba291f81d7e9e043d0102f4f4a342176e7142e7f/src/constants.js#L56): 
```
item.barcode=="%{query.query}"
```

VictorSoroka1:
Fair enough.

PR #1055, opened 2019-11-18
four-second test-timeouts

zburke:
There have been a lot of test failures on PRs in this repository that seem completely unrelated to the PRs themselves. It's maddening. In many cases, the failure is not even a test failure but complaint that the "convergence exceeded the 2000ms timeout". 

This change bumps Mocha's default timeout to 4000ms, doubling the default 2000ms timeout. Maybe it will help and I can feel like I accomplished at least one thing today.

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -1,2 +1,11 @@
 /** @param {import('karma').Config} config */
-module.exports = config => config.set({ client: { captureConsole: false } });
+module.exports = config => {
+  config.set({
+    client: {
+      captureConsole: false,
+      mocha: {
+        timeout: 4000,
+      },
+    }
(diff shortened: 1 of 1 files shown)


mkuklis:
@zburke this looks good but I worry that the default timeout is hardcoded in `Convergence` currently: 

https://github.com/bigtestjs/convergence/blob/master/src/convergence.js#L104

Were you able to confirm that the timeout from the config works? If it does that would be sweet!

zburke:
Ugh. It doesn't work. This PR failed like everything else in ui-users for the last few days with three instances of convergences in the test-setup (i.e. `before` blocks) failing.

PR #1057, opened 2019-11-20
FOLIO-2360 use before blocks

zburke:
Unit tests recently fixed in `stripes-smart-components`, e.g.
/folio-org/stripes-smart-components/pull/665, were failing, it appears,
because of bad initialization practices: values were initialized outside
of `before` and `beforeAll` blocks, allowing tests to proceed before the
data they rely on was properly set up.

This PR tests whether the same issue may be behind some of the
mysterious failures here in ui-users.

Refs [FOLIO-2360](https://issues.folio.org/browse/FOLIO-23602)

### DIFF
--- a/test/bigtest/tests/note-flow-test.js
+++ b/test/bigtest/tests/note-flow-test.js
@@ -13,7 +13,7 @@ const notesAccordion = new NotesAccordion();
 const noteForm = new NoteForm();
 const noteView = new NoteView();
 
-describe('User notes flow', function () {
+describe.skip('User notes flow', function () {
   setupApplication();
 
   let user;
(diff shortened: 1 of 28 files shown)


zburke:
Fascinating, or not, that a test failed on the first run but succeeded on the second run.

VictorSoroka1:
This is a good start)

test/bigtest/tests/note-flow-test.js
```diff
@@ -13,7 +13,7 @@ const notesAccordion = new NotesAccordion();
 const noteForm = new NoteForm();
 const noteView = new NoteView();
 
-describe('User notes flow', function () {
+describe.skip('User notes flow', function () {
```
MaksymDryha: Should these tests be skipped for now, or is it unintentionally?

PR #1067, opened 2019-11-24
Request preferences tests

Yurii-Danylenko:
https://issues.folio.org/browse/UIU-1328

## Purpose
Cover with tests request preferences section.

Requires PR https://github.com/folio-org/stripes-smart-components/pull/670

### DIFF
--- a/test/bigtest/network/config.js
+++ b/test/bigtest/network/config.js
@@ -489,4 +489,16 @@ export default function config() {
 
   this.get('/perms/users/:id', { id: 'test' });
   this.put('/perms/users/:id', { id: 'test' });
+
+  this.get('/request-preference-storage/request-preference', ({ requestPreferences }, request) => {
+    if (request.queryParams.query) {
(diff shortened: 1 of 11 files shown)


test/bigtest/network/config.js
```diff
@@ -489,4 +489,16 @@ export default function config() {
...
+  this.get('/request-preference-storage/request-preference', {
+    requestPreferences: [{
+      id: 'dfd43782-fe91-4612-81c6-b72e3154966f',
+      holdShelf: true,
+      delivery: true,
+      defaultServicePointId: 'servicepointId1',
+      defaultDeliveryAddressTypeId: 'Type1',
+      fulfillment: 'Delivery',
+    }],
+    totalRecords: 1
```
MaksymDryha: I think it would be better to configure mirage models and factories instead of returning hardcoded response. The same configuration has already been done in ui-requests module.

PR #1074, opened 2019-12-05
Release v2.26.0

zburke:

PR #1077, opened 2019-12-06
Release 2.26.0 -temp

VictorSoroka1:


zburke:
Replaces #1076, which replaced #1074. All hail @VictorSoroka1, a man big enough to stand up to BigTest and send it home with its most recent version tucked between its legs.

PR #1083, opened 2019-12-12
UIU-1359: Prevent loan details opening upon click in loans action menu without selecting link

VictorSoroka1:
## Purpose

Prevent loan details opening upon click in loans action menu without selecting link in the scope of [UIU-1359](https://issues.folio.org/browse/UIU-1359).

### Previous behavior
prev

### Current behavior
current


VictorSoroka1:
The migration part is not that obvious. As a part of the issue, I don't plan to introduce higher risks and break stuff. I gave it a try, but it seems more time and knowledge are needed right now for it.

PR #1087, opened 2019-12-17
UIU-1375: Announce search results to screen readers and move focus when results is found

rasmuswoelk:
Updated `<UserSearch>` to announce the number of results found when searching to screen readers. If there are results, the focus will move to the pane header in the results pane to reduce the tabbing distance.

**Ticket:** 
https://issues.folio.org/browse/UIU-1375


rasmuswoelk:
The test error is not related to the code in this PR.

PR #1088, opened 2019-12-17
UIU-1420: Add record last updated and created back to manual patron block. 

mkuklis:
https://issues.folio.org/browse/UIU-1420


zburke:
p.s. SonarCloud's complaint here is bunk as far as I am concerned and I personally would ignore it.

mkuklis:
No idea why `props.selectedItem` exists there. If it's not used I think we can just remove it. I'm afraid of changing more though since this entire implementation is shaky.

mkuklis:
@zburke I took another look at this and you are right that `selectedItem` is not being used. I just removed it.

PR #1095, opened 2019-12-19
UIU-1404: Fix bug with uncorrect displaying of address type

Vladyslav-Velytskyi:
## Purpose
Fix bud with wrong address type displaying on User Details page, now it shows UUID

## Link
https://issues.folio.org/browse/UIU-1404

### Before
screenshot-1

### After
image

### DIFF
--- a/src/views/UserDetail/UserDetail.js
+++ b/src/views/UserDetail/UserDetail.js
@@ -1,7 +1,10 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import get from 'lodash/get';
-import cloneDeep from 'lodash/cloneDeep';
+import {
+  get,
+  keyBy,
+  cloneDeep,
+} from 'lodash';
 import { FormattedMessage } from 'react-intl';
 import {
   AppIcon,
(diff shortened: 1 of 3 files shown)


src/views/UserDetail/UserDetail.js
```diff
@@ -331,6 +334,15 @@ class UserDetail extends React.Component {
     }
   ]
 
+  getAddressesList(addresses, addressTypes) {
```
mkuklis: I know this is not a big list but this is still doing n^2 first mapping over `addresses` and then for each address looping through `addressTypes`. 

Could you try to turn `addressTypes` into a hash outside of map with `keyBy`: https://lodash.com/docs/4.17.15#keyBy 

for example:

`const addressTypesById = keyBy(addressTypes, 'id')`

then inside the address map you could just do:

`const addressTypeOption  = addressTypesById[address.addressType]`

Vladyslav-Velytskyi: Thanks, updated according to your comment

PR #1101, opened 2019-12-23
UIU-1413: add state `notify` value

annamelnyk:
### Description
Passing `notify` field when ActionModal of fee/fine payment is submitted, or `Charge`/`Charge with pay` button is pressed - `notify` value is also passing to BE

---

### Issue
https://issues.folio.org/browse/UIU-1413

### DIFF
--- a/src/components/Accounts/ChargeFeeFine/ChargeFeeFine.js
+++ b/src/components/Accounts/ChargeFeeFine/ChargeFeeFine.js
@@ -64,6 +64,7 @@ class ChargeFeeFine extends React.Component {
       lookup: false,
       pay: false,
       showConfirmDialog: false,
+      notify: null,
     };
     this.onClickCharge = this.onClickCharge.bind(this);
     this.onClickSelectItem = this.onClickSelectItem.bind(this);
@@ -104,11 +105,9 @@ class ChargeFeeFine extends React.Component {
     const owners = _.get(this.props.resources, ['owners', 'records'], []);
     const feefines = _.get(this.props.resources, ['feefines', 'records'], []);
     const selectedLoan = this.props.selectedLoan || {};
-
     const { intl: { formatMessage } } = this.props;
     const item = (selectedLoan.id) ? selectedLoan.item : this.item;
 
-    this.type = type;
     type.paymentStatus = {
       name: 'Outstanding',
     };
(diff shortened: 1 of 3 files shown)


src/components/Accounts/ChargeFeeFine/ChargeFeeFine.js
```diff
@@ -108,6 +108,7 @@ class ChargeFeeFine extends React.Component {
     const { intl: { formatMessage } } = this.props;
     const item = (selectedLoan.id) ? selectedLoan.item : this.item;
 
+    this.type = type;
```
maximdidenkoepam: why do we need it here?

src/components/Accounts/ChargeFeeFine/ChargeFeeFine.js
```diff
@@ -140,15 +141,17 @@ class ChargeFeeFine extends React.Component {
       commentInfo = `${commentInfo} \n ${tagPatron} : ${type.patronInfo}`;
     }
     delete type.comments;
-    delete type.notify;
     delete type.patronInfo;
-    this.type = type;
-    return this.props.mutator.accounts.POST(type)
-      .then(() => this.newAction({}, type.id, type.feeFineType, type.amount, commentInfo, type.remaining, 0, type.feeFineOwner));
+    const typeCopy = { ...type };
```
maximdidenkoepam: you can use lodash omit method

src/components/Accounts/ChargeFeeFine/ChargeFeeFine.js
```diff
@@ -140,15 +141,17 @@ class ChargeFeeFine extends React.Component {
...
+    delete typeCopy.notify;
+    return this.props.mutator.accounts.POST(typeCopy)
+      .then(() => this.newAction({}, typeCopy.id, typeCopy.feeFineType, typeCopy.amount, commentInfo, typeCopy.remaining, 0, typeCopy.feeFineOwner));
   }
 
   newAction = (action, id, typeAction, amount, comment, balance, transaction, createdAt) => {
     const path = `accounts/${this.type.id}`;
+    const notify = this.type.notify;
```
maximdidenkoepam: can we store notify value in state?

src/components/Accounts/ChargeFeeFine/ChargeFeeFine.js
annamelnyk: The code location does not have any special meaning

src/components/Accounts/ChargeFeeFine/ChargeFeeFine.js
annamelnyk: I guess we can

src/components/Accounts/ChargeFeeFine/ChargeFeeFine.js
annamelnyk: Agree

src/components/Accounts/ChargeFeeFine/ChargeFeeFine.js
maximdidenkoepam: I see several type assigning in this file, not sure that it is correct

annamelnyk: Yes, type assigning is implemented in two methods for different flows

maximdidenkoepam: We we can't just pass notify property to a local variable and to newAction method instead of store it in class property?

annamelnyk: Because we anyway should save the `notify` value from this method. I will store `notify` value in state then.

PR #1104, opened 2019-12-27
UIU-1413: Add check if object has status field to avoid page crash

annamelnyk:
### Purpose
Prevent page crash when multiple fee/fine items is paying 

---

### Issue
https://issues.folio.org/browse/UIU-1413

--- 

### Screenshot
image

### DIFF
--- a/src/components/Accounts/Actions/Actions.js
+++ b/src/components/Accounts/Actions/Actions.js
@@ -469,7 +469,10 @@ class Actions extends React.Component {
     const waives = _.get(resources, ['waives', 'records'], []);
     const transfers = _.get(resources, ['transfers', 'records'], []);
     const settings = _.get(resources, ['commentRequired', 'records', 0], {});
-    const warning = accounts.filter(a => a.status && a.status.name === 'Closed').length !== 0 && (actions.regular || actions.waiveMany || actions.transferMany) && params.accountstatus;
+    const closedAccounts = accounts.filter(a => a.status && a.status.name === 'Closed');
+    const isWarning = closedAccounts.length !== 0
+        && (actions.regular || actions.waiveMany || actions.transferMany)
+        && params.accountstatus;
     const warningModalLabelId = actions.regular
       ? 'ui-users.accounts.actions.payFeeFine'
       : actions.waiveMany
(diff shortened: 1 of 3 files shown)


src/components/Accounts/Actions/Actions.js
```diff
@@ -469,7 +469,10 @@ class Actions extends React.Component {
     const waives = _.get(resources, ['waives', 'records'], []);
     const transfers = _.get(resources, ['transfers', 'records'], []);
     const settings = _.get(resources, ['commentRequired', 'records', 0], {});
-    const warning = accounts.filter(a => a.status.name === 'Closed').length !== 0 && (actions.regular || actions.waiveMany || actions.transferMany) && params.accountstatus;
+    const closedAccounts = accounts.filter(a => a.status && a.status.name === 'Closed');
+    const isWarning = closedAccounts.length !== 0
+        && (actions.regular || actions.waiveMany || actions.transferMany)
```
SergiyVSergiyenko: What do you think about
``const hasClosedAccounts = accounts.some(a => a.status && a.status.name === 'Closed');``
``const isWarning = hasClosedAccounts``
``...`` ?

annamelnyk: It could make code more readable, but I guess here should not be used ternary operator, because value of `isWarning` boolean is used in several places inside the conditions

src/components/Accounts/Actions/WarningModal.js
```diff
@@ -173,7 +173,7 @@ class WarningModal extends React.Component {
     };
 
     const values = Object.values(checkedAccounts);
-    const checkedClosed = values.filter(a => a.status.name === 'Closed');
+    const checkedClosed = values.filter(a => a.status && a.status.name === 'Closed');
```
VictorSoroka1: By the way, I suggest using `.some` also here. As I see in the condition below, `checkedClosed.length > 0` the count is just needed to make some decision. So, in this case finding all closed loans is overhead. [Approach](https://github.com/folio-org/ui-users/pull/1104/files#r361674281) which @SergiyVSergiyenko has proposed, sounds more proper to me.

PR #1105, opened 2020-01-02
Refactor open loans list to use <Dropdown /> from stripes-components

VictorSoroka1:
## Purpose

Since `Dropdown` component from stripes-components has been refactored to use `Popdown` internally, the following things are possible now:
1) refactoring of open loan list to use `Dropdown` component so that `Popdown` component in `ui-users` can be removed;
2)  refactoring of closed loan list to use `Dropdown` component instead of `UncontrolledDropdown` which is now considered deprecated;


VictorSoroka1:
@JohnC-80, I found that default `modifiers` setting for the Popdown component

```
  placement: 'bottom',
  modifiers: {
    flip: { boundariesElement: 'scrollParent', padding: 10 },
    preventOverflow: { boundariesElement: 'scrollParent', padding: 10 }
  },
```
does not work well when there is a list of two items upon click on the second item with the `scrollParent` is set as a `boundariesElement` and the `placement` to 'bottom'. In this case, the menu is not placed right below the ellipsis button as it calculated the new position with the clue that there is not enough space to show the menu normally (based on the height of the scrollable container). In order to overcome it, I set `boundariesElement` to `viewport`, so in this case, the menu will be placed correctly. I wonder if it makes sense to set as default instead.

YevheniiMaltsev:
Could you please clarify what ticket is used for these changes?

VictorSoroka1:
There is not ticket for this. I would say this is PR is a result of the discussion in the scope of [PR](https://github.com/folio-org/ui-users/pull/1083) for `UIU-1359`.

YevheniiMaltsev:
Please, provide the information about this change to changelog

PR #1106, opened 2020-01-02
UIU-1408 correctly handle MCL checkboxes

zburke:
Checkboxes in the MCL in the `<WarningModal>` did not correctly reflect
their state because no `isSelected` function was passed to the MCL.

Refs [UIU-1408](https://issues.folio.org/browse/UIU-1408)


zburke:
Handled already in #1099

PR #1107, opened 2020-01-02
UIU-1407 correctly handle MCL checkboxes

zburke:
Checkboxes in the MCL in `<ViewFeesFines>`did not correctly reflect
their state when the select-all checkbox was used.

Refs [UIU-1407](https://issues.folio.org/browse/UIU-1407)


zburke:
Handled already in #1099

PR #1110, opened 2020-01-02
UIU-1412 return to user-profile when closing modals

zburke:
Go back to user's accounts when clicking on cancel or `x` from fee/fine form and accounts listing. 

Fixes [UIU-1412](https://issues.folio.org/browse/UIU-1412)


zburke:
Closed in favor of #1120.

PR #1111, opened 2020-01-03
UIU-1438: Omit 'notify' field upon creating fee/fine

VictorSoroka1:
## Purpose

Omit 'notify' field upon creating fee/fine in order to prevent backend error in the scope of [UIU-1438](https://issues.folio.org/browse/UIU-1438).

## Screenshot
Error


VictorSoroka1:
> What I don't understand is notify is part of the type object if we have to remove it from every request.

I think that it is needed (like [here](https://github.com/folio-org/ui-users/blob/8b85ca42b2b1b59a579030d402844e0176d2cf8c/src/components/Accounts/ChargeFeeFine/ChargeFeeFine.js#L170)) but not for all cases.

PR #1114, opened 2020-01-06
Get rid of 'modifiers' property from action dropdowns

VictorSoroka1:
## Purpose

Get rid of `modifiers` property from action dropdowns in open/closed loan lists since it now matches the defaults. Depends on the folio-org/stripes-components#1139 in stripes-components.


VictorSoroka1:
As a part of this, I tried to improve the situation with tests, seems like it improved a bit, but still a random factor is still present. Master became without fails after about five attempts. As I see the situations with broken tests occurs near to the end of the tests which makes me thinks that there is some kind of memory leak which affect some tests at some point.

PR #1117, opened 2020-01-06
UIU-1410 Clear item data in new fee/fine form

zburke:


zburke:
Do I feel bad merging a PR when some of its unit tests fail? TBH, not really, because those failures are random BS "succeeded but timed out" failures rather than actual errors, and the results will be different on every run. **sigh

Screen Shot 2020-01-06 at 10 32 45 AM

PR #1127, opened 2020-01-10
Migrate BigTest tests to miragejs. Refs STCOR-407

mkuklis:
This is an attempt to migrate our tests from `@bigtest/mirage` to `miragejs`. This depends on the additional work in `stripes-core`:

https://github.com/folio-org/stripes-core/pull/783

### Migration steps:

1.  `import { ... } from '@bigtest/mirage'` becomes `import { ...} from 'miragejs'`
2. The `faker` has to be installed as a separate dependency and imported directly:  `import faker from 'faker'` instead of `import { faker } from '@bigtest/mirage'`
3. Singular names should be used for factories and models and serializers otherwise miragejs will complain about not being able to find them.
4. Add `mirageOptions.serverType = 'miragejs'` in order to opt-in for miragejs implementation:

https://github.com/folio-org/ui-users/pull/1127/files#diff-afc65168f7c0334e511aba82e26e9977R4


mkuklis:
Thank you for pointing this out @zburke! I wouldn't know that we need to do that. I will change the `@folio/stripes-core` to `^3.12.0` after the other PR in stripes-core related to this is merged.

PR #1129, opened 2020-01-12
Improve tests

VictorSoroka1:
## Purpose
Improve the situation with BigTest tests which are unstable recently.

## Approach

- used React.lazy to not load all stuff (seems like now it is possible). I think it would be great to use it from now on in other modules;
- reduced the flow to get to the test (basically reducing the nested describes). For example, the need to test the edit user form. Instead of navigating to /users and then by clicking on the edit button to the edit form, -navigate straight away to edit route;
- grouped some unstable tests. For example, the tests which test for existing properties after the heavy beforeEach has done;

Additionally, the navigation was handled [here](https://github.com/folio-org/ui-users/pull/1129/commits/b4185107e43328f41852cfd2e0d7e86ee1a05906#diff-e74dacc8738c898345810bad610956a4R275) and [here](https://github.com/folio-org/ui-users/pull/1129/commits/b4185107e43328f41852cfd2e0d7e86ee1a05906#diff-e74dacc8738c898345810bad610956a4R238). Since it affected tests, these changed should be made. The motivation for it is that when the user navigates to the edit page straight away and user has not the previous history (imagine user navigates for the first time), clicking on the close button will throw user to the empty tab (like new tab view in the browser), instead of navigating to the users page with the preserved search query.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -13,22 +13,23 @@ import * as Routes from './routes';
 import pkg from '../package';
 import Settings from './settings';
 import commands from './commands';
-import PermissionSets from './settings/permissions/PermissionSets';
-import PatronGroupsSettings from './settings/PatronGroupsSettings';
-import AddressTypesSettings from './settings/AddressTypesSettings';
-import ProfilePictureSettings from './settings/ProfilePictureSettings';
-import OwnerSettings from './settings/OwnerSettings';
-import FeeFineSettings from './settings/FeeFineSettings';
-import WaiveSettings from './settings/WaiveSettings';
-import PaymentSettings from './settings/PaymentSettings';
-import CommentRequiredSettings from './settings/CommentRequiredSettings';
-import RefundReasonsSettings from './settings/RefundReasonsSettings';
-import TransferAccountsSettings from './settings/TransferAccountsSettings';
 import {
   NoteCreatePage,
   NoteViewPage,
   NoteEditPage
 } from './views';
+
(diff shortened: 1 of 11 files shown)


mkuklis:
@VictorSoroka1 let's merge this one in since it's passing. We will rebase master with @zburke's branch afterwards to continue testing the leak. Does that sounds good to you?

VictorSoroka1:
@mkuklis, sure. I just want to mention here:
- that tests failed two days ago and then they started to pass as it happened for #1130 which was failing for quite a long time. I just want to note that after the merge the situation is only going to improve but not resolve the issue with tests.
- I have not tested a thoroughly lazy loading feature, but it seems like testing is passing it should be okay. It would be great if key people of the module review the PR by switching to the branch.

VictorSoroka1:
@JohnC-80, please review stuff related to the routing which I handled [here](https://github.com/folio-org/ui-users/pull/1129/commits/b4185107e43328f41852cfd2e0d7e86ee1a05906#diff-e74dacc8738c898345810bad610956a4R275) and [here](https://github.com/folio-org/ui-users/pull/1129/commits/b4185107e43328f41852cfd2e0d7e86ee1a05906#diff-e74dacc8738c898345810bad610956a4R238). Since it affected tests, I had to make these changes.

mkuklis:
Yes @VictorSoroka1 I realize this won't be a 100% fix but it's moving us into the right direction.

PR #1132, opened 2020-01-15
UIU-1426: Prevent autofill in user's search box

annamelnyk:
# Approach
Add `autocomplete` attribute to prevent autofill in user's search box
# Issue
https://issues.folio.org/browse/UIU-1426


src/views/UserSearch/UserSearch.js
```diff
@@ -335,6 +335,7 @@ class UserSearch extends React.Component {
                               <div className={css.searchGroupWrap}>
                                 <SearchField
                                   aria-label="user search"
+                                  autoComplete="off"
```
zburke: Does `autocomplete` need to be [all lower case](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete) or is that irrelevant?

annamelnyk: HTML attributes are not case-sensitive, so that is irrelevant. But anyway I already changed it to lowercase to follow one common pattern. Thanks.

annamelnyk: It looks like I got error that React requires to write attributes in camelCase style, so I remove back `autoComplete`

PR #1134, opened 2020-01-16
UIU-1248: Handle accessibility issue with focus user's search field

annamelnyk:
# Description
Issue was in loosing focus after visiting Users tab (after setting filters then selecting certain user for example), then going through few another tabs, and then coming back to User's tab - then search field is loosing it's focus.
# Approach
Add prop `autoFocus` to `<SearchField>`
# Issue
https://issues.folio.org/browse/UIU-1248
# Screenshot
image


JohnC-80:
As part of this PR, could you please remove the error-prone portion of the code that did control this? in UserSearch.js at lines 96 - 105 the `cDM` method and the following method can be removed :

```
componentDidMount() {
    this.possiblyfocusSearchField();
  }

  possiblyfocusSearchField = () => {
    const { search, pathname } = window.location;
    if ((pathname + search) === pkg.stripes.home && this.searchField.current) {
      this.searchField.current.focus();
    }
  }
```

Also the variable/ ref to the search field and it's usage...

I think this behavior may be changed in the future, since autofocus will focus the component *every time that it mounts... As I mention in the comments on the JIRA issue, we may not always want to focus the search field every time it mounts... but we can wait for feedback on this from the accessibility SIG...

annamelnyk:
@JohnC-80, sure. Removed.

PR #1135, opened 2020-01-16
FOLIO-2097 eliminate setupApplication in before

zburke:
`before` blocks are utterly completely hoisted to the top of everything
and run before _every_ test. This means that in the change-due-date-test
script, we'd be running user-status-filter-test's `before` block and all
other tests' `before` blocks. That could lead to lots of unnecessary app
mounting and unmounting and general madness.

The findings of FOLIO-2360 notwithstanding and #1057 notwithstanding, it
seems that a better approach than begin blocks may be to nest `describe`
blocks. This allows us to make sure `setupApplication` is run once at
the top level of each test script and guarantees it finishes executing
before nested describe blocks, which seems to be the problem #1057
alludes too.

Refs [FOLIO-2097](https://issues.folio.org/browse/FOLIO-2097), [FOLIO-2360](https://issues.folio.org/browse/FOLIO-2360)


mkuklis:
@zburke thank you for merging this one! 🙏

PR #1136, opened 2020-01-16
UIU-1349 Navigate back to users preview/search from edit view on cancel

JohnC-80:
fixes [UIU-1349](https://issues.folio.org/browse/UIU-1349)

# Problem
UI-users stuck in a navigation loop if another app is visited while the edit/create screen is open. Clicking the X or Cancel buttons would use `history.back()` which could send the user back to an outside app.

## Approach
Navigate to the users' detail view by `history.push()` rather than depending on `history.back()`. The `search` portion of the location is retained and added to the pushed URL so that search results/filters/text will be kept.


src/views/UserEdit/UserEdit.js
```diff
@@ -185,7 +186,7 @@ class UserEdit extends React.Component {
           return mutator.perms.POST({ userId: user.id, permissions: [] });
         })
         .then(() => {
```
VictorSoroka1: I suggest to simplify logic and have just one mutator call with a check for credentials in the first then.

```suggestion
  create = ({ requestPreferences, creds, ...userFormData }) => {
    const { mutator, history, location: { search } } = this.props;
    const userData = cloneDeep(userFormData);
    const credentialsAreSet = userData.username;
    const user = { ...userData, id: uuid() };
    user.personal.addresses = toUserAddresses(user.personal.addresses);

    mutator.records.POST(user)
      .then((response) => {
        if (!credentialsAreSet) {
          return response;
        }

        const credentials = {
          password: '',
          ...creds,
          username: userData.username,
        };

        return mutator.creds.POST(Object.assign(credentials, { userId: user.id }));
      })
      .then(() => {
        this.createRequestPreferences(requestPreferences, user.id);
        return mutator.perms.POST({ userId: user.id, permissions: [] });
      })
      .then(() => {
        history.push(`/users/preview/${user.id}${search}`);
      });
    }
  }
```

PR #1138, opened 2020-01-17
UIU-1450 XHRs musn't leak memory after component unmounts

zburke:
Renew: Does not check its lifecycle
Sets the state, leaks out memory
Promise completes alone
When it resolves, component heaves, detached from the DOM
The console begins to moan

Memory, all alone without GC
It is leaking out two ways
From this component when:
It remembers loan policy and count-request too
Stop the memory leak again!

Refs [UIU-1450](https://issues.folio.org/browse/UIU-1450)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 * Refresh list of loans after anonymization. Fixes UIU-1046.
 * Add UI to mark items Declared lost. Refs UIU-1202.
 * Omit `comments`, `patronInfo` fields, when fee/fine is paying, to prevent backend error. Fixes UIU-1413.
+* Don't update state in `withRenew` when unmounted. Fixes UIU-1450.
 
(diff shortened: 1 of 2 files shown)


VictorSoroka1:
I think that maybe having full console logs is might be better to identify the warnings like this.

https://github.com/folio-org/ui-users/blob/master/karma.conf.js#L2

VictorSoroka1:
On the other hand I see that right now without that option the amount of lines for just logs from karma run would be 16422 instead of about 500. A lot of noise basically.

zburke:
Following up on your console logs comment: I totally agree that suppressing this stuff feels wrong. TBH, I'd sort of love to configure Jenkins to complain in a PR comment when something writes to the console. The console should be clean!

The problem we have is that stripes-connect makes heavy use of some deprecated React methods and those deprecation warnings, while important, pollute the console so much they make it useless. Refactoring off those methods will take some real effort and it's not clear that refactoring stripes-connect is worth the effort (see [STRIPES-663](https://issues.folio.org/browse/STRIPES-663)).

PR #1139, opened 2020-01-22
FOLIO-2097 avoid CommandList and HasCommand

zburke:
We have a theory that a memory leak in a transitive dependency of
`<CommandList>` and `<HasCommand>` is behind the failing unit tests. The
problem may be that event listeners are not properly detached and
recycled.

This is a test of that. Obviously, we don't actually want to get rid of
key bindings. But we would really REALLY like to have reliable unit
tests. 

Refs [FOLIO-2097](https://issues.folio.org/browse/FOLIO-2097)

### DIFF
--- a/src/components/PermissionsAccordion/components/PermissionsModal/PermissionsModal.js
+++ b/src/components/PermissionsAccordion/components/PermissionsModal/PermissionsModal.js
@@ -36,8 +36,8 @@ class PermissionsModal extends React.Component {
           excludePermissionSets,
         } = props;
         const query = [
-          ...(listInvisiblePerms || ['visible==true']),
-          ...(excludePermissionSets && ['mutable==false']),
+          ...(listInvisiblePerms ? [] : ['visible==true']),
+          ...(excludePermissionSets ? ['mutable==false'] : []),
         ];
         const queryString = query.length
           ? `query=(${query.join(' and ')})`
--- a/src/index.js
+++ b/src/index.js
@@ -6,13 +6,11 @@ import { hot } from 'react-hot-loader';
 import _ from 'lodash';
 
 import { Route, Switch, IfPermission } from '@folio/stripes/core';
-import { CommandList, HasCommand } from '@folio/stripes/components';
 
 import * as Routes from './routes';
 
 import pkg from '../package';
 import Settings from './settings';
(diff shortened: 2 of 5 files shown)


VictorSoroka1:
I think HotKeys may be also a good candidate to fit the theory. 

https://github.com/folio-org/stripes-core/blob/master/src/RootWithIntl.js#L70

JohnC-80:
This seems like it comes from `mousetrap` - if we can replace that dependency of Hotkeys, it may clean up this aspect of this.... but it's difficult here, because event listeners also leak in react dev build: https://github.com/facebook/react/issues/12141#issuecomment-385536273

If we go tearing out shortcut-keys as a temporary solution, my question about the future work will be who's going to put them back?

VictorSoroka1:
I wonder whether it sounds sane to try to run tests in production mode and see the difference.

zburke:
I don't intend to merge this; tearing out shortcut keys was just an experiment to see if it would allow the PR to pass, i.e. to see if we could identify a smoking gun.

zburke:
I agree, @VictorSoroka1, that running tests in production mode is another good option. We'd have to figure out how to get the production build of React _without_ stripping the `data-test-*` attributes the tests rely on. I don't know how easy/difficult that would be.

VictorSoroka1:
Good point about data-test attributes. It might be a good idea to add a check that if the env variable is `test` to preserve them in the prod build configuration.

src/components/PermissionsAccordion/components/PermissionsModal/PermissionsModal.js
```diff
@@ -36,8 +36,8 @@ class PermissionsModal extends React.Component {
           excludePermissionSets,
         } = props;
         const query = [
-          ...(listInvisiblePerms || ['visible==true']),
-          ...(excludePermissionSets && ['mutable==false']),
+          ...(listInvisiblePerms ? [] : ['visible==true']),
+          ...(excludePermissionSets ? ['mutable==false'] : []),
```
zburke: @OleksandrAntonenko1, you wrote the original code here; does this change look correct to you? The original intent was a bit hard to decipher but I believe this preserves the logic.

PR #1143, opened 2020-01-24
FOLIO-2097 comment CommandList and HasCommand

zburke:
We have a theory that a memory leak in a transitive dependency of
`<CommandList>` and `<HasCommand>` is behind the failing unit tests. The
problem may be that event listeners are not properly detached and
recycled.

This is an alternative to #1139, commenting out those components instead
of removing them, which may be easier to revert.

Refs [FOLIO-2097](https://issues.folio.org/browse/FOLIO-2097)

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -6,7 +6,7 @@ import { hot } from 'react-hot-loader';
 import _ from 'lodash';
 
 import { Route, Switch, IfPermission } from '@folio/stripes/core';
-import { CommandList, HasCommand } from '@folio/stripes/components';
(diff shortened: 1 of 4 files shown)


zburke:
I pushed this up as an alternative to #1139, but lint gets cranky about indentation and a variety of other things. TBH, I think the easiest path may be to merge #1139 and then revert it once we get a `mousetrap` alternative in place, assuming the API from `<CommandList>` and `<HasCommand>` to `<Hotkeys>` remains the same once we replace `mousetrap` under the hood.

PR #1147, opened 2020-01-29
UIU-897: Remove/add translation keys for routing refactor

julianladisch:


zburke:
OOC, @julianladisch, do you have a script that identifies inactive translation keys? I've done some simple things in the past, but nothing very robust. I know in some cases we assemble keys from substrings so I'm hesitant to remove keys as a rule, though lists like the countries (which have clearly moved to another repo) are less worrisome.

PR #1150, opened 2020-01-30
UIU-1350: Mark fields as required when filling in the username or pw

SergiyVSergiyenko:
## Purpose
Mark the username and password fields as required when filling in one of them. Story [UIU-1350](https://issues.folio.org/browse/UIU-1350).

### Screenshots
**Before**
Before

**After**
Filling_in_username
Filling_in_passwword


src/components/EditSections/EditExtendedInfo/EditExtendedInfo.js
```diff
@@ -30,6 +30,10 @@ class EditExtendedInfo extends Component {
     onToggle: PropTypes.func.isRequired,
   };
 
+  state = { isFieldRequired: false };
```
YevheniiMaltsev: It seems, there are a lot of different fields, but only 2 of them should be marked as required in the scope of the mentioned story. Please, use a more exact name to identify the fields in which the state is controlled.

src/components/EditSections/EditExtendedInfo/EditExtendedInfo.js
```diff
@@ -30,6 +30,10 @@ class EditExtendedInfo extends Component {
...
+
+  toggleFieldRequired = (_, value) => this.setState({ isFieldRequired: !!value });
```
YevheniiMaltsev: Please, clarify the fields that will be marked as required in the function name.

src/components/EditSections/EditExtendedInfo/PasswordControl/PasswordControl.js
```diff
@@ -21,6 +27,7 @@ class PasswordControl extends React.Component {
   }
 
   render() {
+    const { isFieldRequired, toggleFieldRequired } = this.props;
```
YevheniiMaltsev: Please, move every const to the own line

PR #1151, opened 2020-01-31
UIU-1448 - Filter Users by Tags

maximdidenkoepam:
# Purpose
Add ability to filter users by tags.

# Link
https://issues.folio.org/browse/UIU-1448

# Screenshot
Screen Shot 2020-01-31 at 11 18 33

### DIFF
--- a/src/views/UserSearch/Filters.js
+++ b/src/views/UserSearch/Filters.js
@@ -1,38 +1,139 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import { FormattedMessage } from 'react-intl';
+import {
+  get,
+  sortBy,
+} from 'lodash';
 
 import {
-  FilterGroups,
+  Accordion,
+  AccordionSet,
+  FilterAccordionHeader,
 } from '@folio/stripes/components';
+
+import {
+  CheckboxFilter,
+  MultiSelectionFilter,
+} from '@folio/stripes/smart-components';
+
+import { statusFilter } from '../../constants';
 
 export default class Filters extends React.Component {
   static propTypes = {
     activeFilters: PropTypes.object,
+    resources: PropTypes.object.isRequired,
     onChangeHandlers: PropTypes.object.isRequired,
-    config: PropTypes.arrayOf(PropTypes.object),
   };
 
   static defaultProps = {
     activeFilters: {},
   }
 
-  render() {
+  getStaticFilterValues = (items) => {
+    return items.map(item => ({ label: item, value: item }));
+  };
+
+  getValuesFromResources = (type, key, sortByKey = false) => {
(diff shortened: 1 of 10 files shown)


VictorSoroka1:
Please add entry to the changelog.

src/views/UserSearch/Filters.js
```diff
@@ -1,38 +1,116 @@
...
 
   static defaultProps = {
     activeFilters: {},
   }
 
-  render() {
+  getStaticFilterValues = (items) => {
+    return items.map(item => ({ label: item, value: item }));
+  };
+
```
zburke: We should convert `label` to a translated value, i.e. a `<FormattedMessage>` corresponding to something with a key like `filters.statusFilter.${item}`, since these are static values. This isn't criticism; the previous code didn't handle the situation either. Seeing `getStaticFilterValues()` nicely factored out just highlights the fact that the translation is missing. Filed [UIU-1467](https://issues.folio.org/browse/UIU-1467).

src/views/UserSearch/Filters.js
```diff
@@ -1,38 +1,116 @@
...
+          label={<FormattedMessage id="ui-users.information.patronGroup" />}
+          separator={false}
+          onClearFilter={() => clearGroup('pg')}
+        >
+          <CheckboxFilter
+            dataOptions={this.getValuesFromResources('patronGroups', 'group')}
+            name="pg"
+            selectedValues={pg}
+            onChange={this.handleFilterChange}
+          />
```
zburke: We sort the tags; why not the patron groups? It feels to me like we should choose between assuming all resources will (1) already be sorted or (2) need to be sorted here.

maximdidenkoepam: I was asked by @a-mbreaux  about it in similar ticket for ui-requests (sort tags) that is why I've added this ability and did not change existing filters.

src/views/UserSearch/Filters.js
maximdidenkoepam: Is it okay if I will do this in scope of this user story?

src/views/UserSearch/Filters.js
zburke: Either we should expect the list to be sorted (as we do for patron groups) or we should expect to sort it ourselves (as we do for tags). It does not matter which option we choose, but let's be consistent.

src/views/UserSearch/Filters.js
zburke: Up to you. I'd probably do it separately just to avoid having this PR languish if that part turns out to be more complicated than you expect. But if it turns out to be simple and you can sneak it in, I'm not gonna argue!

PR #1170, opened 2020-02-18
UIU-1483: Send notify patron value in the Confirm fee/fine cancellation modal

annamelnyk:
# Description
Send `Notify Patron` checkbox value to backend in the Confirm fee/fine cancellation modal
# Issue
https://issues.folio.org/browse/UIU-1483
# Screenshots
Annotation 2020-02-18 184107
Annotation 2020-02-18 184145


zburke:
I restate my case that omitting un-ticked checkboxes from form submissions rather than including them with empty values is one of the single greatest design failures in the history of computing.

annamelnyk:
> I restate my case that omitting un-ticked checkboxes from form submissions rather than including them with empty values is one of the single greatest design failures in the history of computing.

Oh,  yes, I agree

PR #1171, opened 2020-02-18
UIU-1299: Execute loan renew in sequence

mkuklis:
https://issues.folio.org/browse/UIU-1299


src/components/Loans/OpenLoans/OpenLoansControl.js
```diff
@@ -288,6 +292,7 @@ class OpenLoansControl extends React.Component {
       changeDueDateDialogOpen,
       allChecked,
       patronBlockedModal,
+      renewing
```
SergiyVSergiyenko:
```suggestion
      renewing,
```

PR #1172, opened 2020-02-19
UIU-1213: Add UI to mark items Claim returned

SergiyVSergiyenko:
## Purpose
Add the ability to mark items "Claimed returned" on UI. [Story](https://issues.folio.org/browse/UIU-1213).

**Screenshot**
Screen Shot 2020-02-19 at 15 41 35

### DIFF
--- a/src/components/ClaimReturnedDialog/ClaimReturnedInfo.js
+++ b/src/components/ClaimReturnedDialog/ClaimReturnedInfo.js
@@ -1,0 +1,123 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { FormattedMessage } from 'react-intl';
+
+import {
+  Button,
+  Layout,
+  TextArea,
+  Row,
+  Col,
+} from '@folio/stripes/components';
+import { stripesConnect } from '@folio/stripes/core';
+import SafeHTMLMessage from '@folio/react-intl-safe-html';
+
+import css from './ClaimReturnedDialog.css';
+
+class ClaimReturnedInfo extends React.Component {
+  static manifest = Object.freeze({
+    claimReturned: {
+      type: 'okapi',
+      fetch: false,
+      POST: {
+        path: 'circulation/loans/!{loan.id}/claim-item-returned',
+      },
+    },
+  });
+
+  static propTypes = {
+    mutator: PropTypes.shape({
+      claimReturned: PropTypes.shape({
+        POST: PropTypes.func.isRequired,
+      }).isRequired,
+    }).isRequired,
+    loan: PropTypes.object.isRequired,
+    onClose: PropTypes.func.isRequired,
+  };
+
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      additionalInfo: '',
+    };
+  }
+
+  handleAdditionalInfoChange = event => {
+    this.setState({ additionalInfo: event.target.value });
+  };
+
+  submitClaimReturned = async () => {
+    const { additionalInfo } = this.state;
+
+    const {
+      mutator: {
+        claimReturned: {
+          POST,
+        }
+      },
+      onClose,
+    } = this.props;
+
+    await POST({
+      itemClaimedReturnedDateTime: new Date().toISOString(),
+      comment: additionalInfo,
+    });
+
+    onClose();
+  };
+
+  render() {
+    const {
+      loan,
+      onClose,
+    } = this.props;
+
+    const { additionalInfo } = this.state;
+
+    return (
+      <div>
+        <Layout className="flex">
+          <Layout className="flex">
+            <SafeHTMLMessage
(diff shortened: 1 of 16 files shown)


src/components/Loans/OpenLoans/OpenLoansControl.js
```diff
@@ -227,15 +229,17 @@ class OpenLoansControl extends React.Component {
...
+  claimReturned = loan => this.props.claimReturned(loan);
 
   feefineDetails = (loan, e) => {
     const {
       resources,
       history,
       match: { params }
     } = this.props;
-    const accounts = get(resources, ['loanAccount', 'records'], []);
+    const accounts = resources?.loanAccount?.records || [];
```
mkuklis: This looks so much cleaner 👏

src/components/ClaimReturnedDialog/ClaimReturnedInfo.js
```diff
@@ -0,0 +1,124 @@
...
+
+import {
+  Button,
+  Layout,
+  TextArea,
+  Row,
+  Col,
+} from '@folio/stripes/components';
+import { stripesConnect } from '@folio/stripes/core';
+
```
YevheniiMaltsev: Please, remove this blank line

src/components/ClaimReturnedDialog/ClaimReturnedInfo.js
```diff
@@ -0,0 +1,124 @@
...
+import SafeHTMLMessage from '@folio/react-intl-safe-html';
+
+import css from './ClaimReturnedDialog.css';
+
+class ClaimReturnedInfo extends React.Component {
```
YevheniiMaltsev: It seems, this class is very similar to DeclareLostInfo.js. Could you please clarify if there are opportunities/reasons to create a shared component?

src/components/Loans/OpenLoans/OpenLoansControl.js
```diff
@@ -253,9 +257,9 @@ class OpenLoansControl extends React.Component {
     }
   };
 
-  feeFineCount = (loan) => {
+  feeFineCount = loan => {
     const { resources } = this.props;
-    const accounts = get(resources, ['loanAccount', 'records'], []);
+    const accounts = resources?.loanAccount?.records || [];
```
YevheniiMaltsev: Thanks for syntax updates. It looks great.
Please, provide a refactoring in the separated commit to separate functional and styling changes. It might be useful for reviewers.

YevheniiMaltsev: Thanks for syntax updates. It looks great.
Please, provide a refactoring in the separated commit to separate functional and styling changes. It might be useful for reviewers.

src/components/Loans/OpenLoans/OpenLoansControl.js
mkuklis: @SergiyVSergiyenko I wonder if we should use `nullish coalescing` in this case so:

 `resources?.loanAccount?.records ?? [];` 

I don't think it makes any difference in this example but perhaps keep it in consistent with other cases and having just one pattern will be better? What do you think?

src/views/LoanDetails/LoanDetails.js
```diff
@@ -473,10 +485,13 @@ class LoanDetails extends React.Component {
...
               </Col>
-              <Col xs={2}>
+              <Col
+                data-test-loan-claimed-returned
+                xs={2}
+              >
                 <KeyValue
                   label={<FormattedMessage id="ui-users.loans.details.claimedReturned" />}
-                  value="TODO"
+                  value={claimedReturnedDate ? (<FormattedTime value={claimedReturnedDate} day="numeric" month="numeric" year="numeric" />) : (<NoValue />)}
```
YevheniiMaltsev: It seems, this line is too long. Please, split it

src/components/Loans/OpenLoans/OpenLoansControl.js
SergiyVSergiyenko: @mkuklis, thanks for the question. Since resources.loanAccount.records is an array (or undefined), and we do not expect to get an empty string or zero there, the `nullish coalescing` operator is not required. Although, in our case, it works as well.

src/components/ClaimReturnedDialog/ClaimReturnedInfo.js
SergiyVSergiyenko: Thanks for raising this question. Since the method `submit ...` is fundamentally different by the parameters passed in the POST request, refactoring will require significant efforts not provided for by scope of this issue.

YevheniiMaltsev: It seems, the submit and the connection logic can be moved to the parent component.

PR #1174, opened 2020-02-21
UIU-1446: Security update eslint to >= 6.2.1 or eslint-util >= 1.4.1

SergiyVSergiyenko:
# Purpose
Update eslint to v6.2.1

# Link
https://issues.folio.org/browse/UIU-1446


src/components/AddServicePointModal/AddServicePointModal.js
```diff
@@ -23,8 +23,12 @@ class AddServicePointModal extends React.Component {
     intl: intlShape.isRequired,
   }
 
-  state = {
-    selection: {},
+  constructor(props) {
+    super(props);
```
ryandberger: `react/state-in-constructor` was turned off in `eslint-config-stripes` 5.2.0. If you upgrade to that version, you can revert these changes.

test/bigtest/interactors/users/header-dropdown-menu.js
```diff
@@ -0,0 +1,12 @@
...
+  clickable,
+  interactor,
+  isVisible,
+} from '@bigtest/interactor';
+
+@interactor class HeaderDropdownMenu {
+  clickExportToCSV = clickable('#export-overdue-loan-report');
+  exportBtnIsVisible = isVisible('#export-overdue-loan-report');
+}
+
```
annamelnyk: @SergiyVSergiyenko, plenty of changes in current module were done! And in tests occurs very simular interactors with same names (like HeaderDropdown, CellInteractor, RowInteractor), I guess that will confuse other developers, when they try to search if current interactor already exists and will find few identical interactors. I think you could rename interactors with more informative way (with clarifying to which page/view current interactor relates), or if that is possible in some cases - to make one reusable interactor without duplications. What do you think?

src/settings/permissions/PermissionSetForm.js
```diff
@@ -150,10 +149,9 @@ class PermissionSetForm extends React.Component {
 
   saveSet(data) {
     const filtered = omit(data, ['childOf', 'grantedTo', 'dummy']);
-    const permSet = Object.assign({}, filtered, {
+    const permSet = { ...filtered,
```
YevheniiMaltsev: It seems, every object should be placed to its own line for consistency.

src/components/HelperApp/HelperApp.js
```diff
@@ -8,7 +8,7 @@ class HelperApp extends React.Component {
   static propTypes = {
     match: PropTypes.object,
     stripes: PropTypes.object,
-    close: PropTypes.func,
+    onClose: PropTypes.func,
```
YevheniiMaltsev: As far as I understand, It was changed from close to onClose in one place. Could you please clarify if it was broken before or this change will affect the behavior?

CHANGELOG.md
```diff
@@ -37,6 +37,7 @@
 * Display `effective call number prefix`, `call number`, `call number suffix`, `enumeration`, `chronology`, `volume` in loans contexts. Refs UIU-1347, UIU-1391.
 * Prevent change due date for declared lost items. Refs UIU-1207.
 * Tweak text for declared lost modal. Refs UIU-1444.
+* Update eslint to >= 6.2.1 or eslint-util >= 1.4.1. Refs UIU-1446.
```
YevheniiMaltsev: There are a lot of provided changes before the release date, and there is a poor test code coverage level. Even stylish refactoring can cause some issues. So how can we be sure that there are not any broken changes?

PR #1179, opened 2020-02-26
UIU-1446: Security update eslint to >= 6.2.1 with eslint-config-strip…

SergiyVSergiyenko:
# Purpose
Update eslint to v6.2.1 in accordance with eslint-config-stripes 5.2.0

# Link
https://issues.folio.org/browse/UIU-1446


SergiyVSergiyenko:
After upgrading eslint-config-stripes to v5.2.0, it looks much better and, most importantly, much more readable. @ryandberger, thanks for pointing this out.

PR #1181, opened 2020-02-27
UIU-1247:  Add policy links on Fee/Fine Details page (plus other changes)

annamelnyk:
# Description

On Fee/Fine Details page:
- add the display of "Overdue Policy" and "Lost Item Policy" as links to the actual policy

- display  "Contributors"

- combine the display of existing fields "Instance" and "Material type"

- rename the "Instance (Item type)" column on the Fees/Fines History page

# Story
https://issues.folio.org/browse/UIU-1247

# Screenshots
image
image


src/routes/AccountDetailsContainer.js
```diff
@@ -87,13 +94,49 @@ class AccountDetailsContainer extends React.Component {
     return groups.filter(g => g.id === user.patronGroup)[0] || {};
   }
 
+  getItemDetails = () => {
+    const account = this.getAccount();
+    const { resources } = this.props;
+    const loanRecord = _.get(resources, ['loans', 'records'], []);
```
YevheniiMaltsev: Could you please clarify if the syntax from the line 103 'const contributorsRecord = item[0]?.item?.contributors ?? [];' can be applied here for the consistency?

And could you please check changes in other files? For example, AccountDetails.js.

test/bigtest/network/scenarios/fee-fine-details.js
```diff
@@ -0,0 +1,66 @@
...
+
+  server.post('/loans', (schema, request) => {
+    const body = JSON.parse(request.requestBody);
+    return schema.feefineactions.create(body);
+  });
+
+  server.get('/loans', (schema, request) => {
+    const url = new URL(request.url);
+    const cqlQuery = url.searchParams.get('query');
+    if (cqlQuery != null) {
```
YevheniiMaltsev: Please, could you add a blank line before every if statement to increase the code readability?

The same comment can be applied for line 54

src/components/Accounts/ViewFeesFines/ViewFeesFines.js
```diff
@@ -190,6 +190,12 @@ class ViewFeesFines extends React.Component {
     return res;
   }
 
+  formatTitle(f) {
```
YevheniiMaltsev: Could you please clarify what does it mean 'f' in this case? 

This approach is used at this file, but it seems we can try to add more clear names in scope of new changes. @ViktorSoroka What do you think about it?

src/routes/AccountDetailsContainer.js
```diff
@@ -39,6 +41,11 @@ class AccountDetailsContainer extends React.Component {
     activeRecord: {
       accountId: '0',
     },
+    loans: {
+      type: 'okapi',
+      records: 'loans',
+      path: 'circulation/loans?query=(userId=:{id})&limit=100',
```
YevheniiMaltsev: As far as I understand the idea, this solution requests the information about the first 100 loans and filter them to find the needed value. If there are 101 loans and user selects the item 101, it seems the application will work incorrectly. As far as I remember, the similar issue was reproducible a lot of times in scope of the ui-users. Maybe, it should be fixed in scope of the separated stories.

src/routes/AccountDetailsContainer.js
YevheniiMaltsev:
```suggestion
    const loanRecords = _.get(resources, ['loans', 'records'], []);
```
It seems there are information about several records. Or do I misunderstand the idea?

src/routes/AccountDetailsContainer.js
annamelnyk: Hmm, I do not know if there are any limitations (how many loans can have single user), but in current request we can just remove the part of query with limit, right? And if we cannot remove limit then should this problem be solved on BE side?

YevheniiMaltsev: If the limit=100 is removed, the server will use the value 10 by default.
In the best case, we should endpoint (or find other mechanisms) to get information about the target loan by id instead of requesting a lot of items. It seems, there are no accurate numbers about the amount of the possible loans per user. But according to the created Jira tickets (For example, UIU-1299), we should support more than 100-500 loans. Usually, we just increase the limit value keeping in mind, that we have to request hundreds of loans just to select one (It seems, this is not about a good performance). I think we can try to increase the limit value for now (temp solution) and discuss with PO new endpoint creation (if there are not any other ways to request the information about 1 loan) to handle this problem in the future.

My comment is not a change request. I just wanted to mention the possible issue.

src/routes/AccountDetailsContainer.js
```diff
@@ -87,13 +94,49 @@ class AccountDetailsContainer extends React.Component {
     return groups.filter(g => g.id === user.patronGroup)[0] || {};
   }
 
+  getItemDetails = () => {
+    const account = this.getAccount();
+    const { resources } = this.props;
+    const loanRecords = resources?.loans?.records ?? [];
+    const itemId = account?.itemId;
+    const item = loanRecords.filter((loan) => loan.itemId === itemId);
+    const contributorsRecord = item[0]?.item?.contributors ?? [];
```
YevheniiMaltsev: Sorry, I missed it yesterday.  It seems, contributorsRecord can contain information about many contributors. Please, rename it.

src/components/Accounts/ViewFeesFines/ViewFeesFines.js
```diff
@@ -190,6 +190,12 @@ class ViewFeesFines extends React.Component {
     return res;
   }
 
+  formatTitle(item) {
+    const instanceTypeString = item.materialType ? `(${item.materialType})` : '';
+
+    return `${item.title} ${instanceTypeString}` || '-';
```
mkuklis: @annamelnyk I think `|| '-'` would never happen here right? Perhaps we should remove it or refactor it a bit if we do expect the title to be empty.

src/views/AccountDetails/AccountDetails.js
```diff
@@ -179,6 +184,16 @@ class AccountDetails extends React.Component {
...
   }
 
+  getInstanceInfo = () => {
+    const { resources } = this.props;
+    const account = resources?.accountHistory?.records[0] ?? {};
+    const instanceTitle = account?.title;
+    const instanceType = account?.materialType;
+    const instanceTypeString = instanceType ? `(${instanceType})` : '';
+
+    return `${instanceTitle} ${instanceTypeString}` || '-';
```
mkuklis: @annamelnyk I think it is similar to the other pattern above.

test/bigtest/network/scenarios/fee-fine-details.js
```diff
@@ -0,0 +1,68 @@
...
+      name: 'Open'
+    },
+    amount: 100,
+    remaining: 100,
+    loanId: loan.id,
+  });
+  server.create('account', {
+    userId: user.id,
+    status: {
+      name: 'Closed'
```
mkuklis: missing comma

test/bigtest/network/scenarios/fee-fine-details.js
```diff
@@ -0,0 +1,68 @@
...
+    },
+    amount: 0,
+    remaining: 0
```
mkuklis: missing comma

PR #1182, opened 2020-02-28
UIU-1265: Implement declare items lost permission

SergiyVSergiyenko:
## Purpose

Implement declare items lost permission as part of [UIU-1265](https://issues.folio.org/browse/UIU-1265). This permission works in conjunction with permissions for viewing the loans.

## Screenshot

Screen Shot 2020-02-27 at 13 24 58

### DIFF
--- a/src/views/LoanDetails/LoanDetails.js
+++ b/src/views/LoanDetails/LoanDetails.js
@@ -323,7 +323,7 @@ class LoanDetails extends React.Component {
       actionDate: la => <FormattedTime value={get(la, ['metadata', 'updatedDate'], '-')} day="numeric" month="numeric" year="numeric" />,
       dueDate: la => <FormattedTime value={la.dueDate} day="numeric" month="numeric" year="numeric" />,
       itemStatus: la => la.itemStatus,
-      source: la => <Link to={`/users/view/${la.user.id}`}>{getFullName(la.user)}</Link>,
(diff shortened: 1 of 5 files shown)


src/views/LoanDetails/LoanDetails.js
```diff
@@ -406,14 +406,16 @@ class LoanDetails extends React.Component {
...
+                    data-test-declare-lost-button
+                    disabled={buttonDisabled || isDeclaredLostItem}
+                    buttonStyle="primary"
+                    onClick={() => declareLost(loan)}
+                  >
+                    <FormattedMessage id="ui-users.loans.declareLost" />
+                  </Button>
+                </IfPermission>
               </span>
             </Row>
```
annamelnyk: @SergiyVSergiyenko, I am wondering if user does not have any of those permissions, will current row display as empty one? Should we check  if user has any permission then render `Row` etc ?

zburke: Good point, though in this case I'm not worried about it. The `<Row>` in this case is just a wrapper for a bunch of `<Button>` elements at the top of the pane. If it is present but empty it doesn't really affect the layout. If we had, e.g. an empty dropdown menu, then it would definitely be worth wrapping the other Row/Menu/Whatever element to suppress the display when all permission-tests return false. 

It's great that you are thinking about this kind of display issue though. Thanks for bringing it up!

SergiyVSergiyenko: Thank you, @zburke, for clarification.

PR #1187, opened 2020-03-03
UIU-1492 Create UI permissions for Settings > Users > Custom Fields CRUD

BogdanDenis:
## Purpose

- Add ability to control who at the library has the ability to CRUD custom fields

### DIFF
--- a/package.json
+++ b/package.json
@@ -257,10 +257,11 @@
         "permissionName": "ui-users.settings.customfields.view",
         "displayName": "Settings (Users): Can view custom fields",
         "subPermissions": [
-          "customfields.collection.get",
-          "customfields.item.get",
+          "custom.fields.collection.get",
+          "custom.fields.item.get",
           "custom.fields.item.stats.get",
-          "settings.users.enabled"
+          "settings.users.enabled",
+          "configuration.entries.collection.get"
         ],
         "visible": true
       },
@@ -270,8 +271,10 @@
         "subPermissions": [
           "ui-users.settings.customfields.view",
           "custom.fields.collection.put",
-          "customfields.item.post",
-          "customfields.item.put"
+          "custom.fields.item.post",
+          "custom.fields.item.put",
+          "configuration.entries.item.post",
+          "configuration.entries.item.put"
         ],
         "visible": true
       },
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -253,6 +253,39 @@
         ],
         "visible": true
       },
+      {
+        "permissionName": "ui-users.settings.customfields.view",
+        "displayName": "Settings (Users): Can view custom fields",
+        "subPermissions": [
+          "customfields.collection.get",
+          "customfields.item.get",
```
MaksymDryha: It should be `custom.fields.collection.get` and `custom.fields.item.get`
Please, correct the names of other backend permissions too
Also, `configuration.entries.collection.get` should be added to the subpermissions too

package.json
```diff
@@ -253,6 +253,39 @@
...
+        "visible": true
+      },
+      {
+        "permissionName": "ui-users.settings.customfields.all",
+        "displayName": "Settings (Users): Can create, edit, view and delete custom fields",
+        "subPermissions": [
+          "ui-users.settings.customfields.view",
+          "ui-users.settings.customfields.edit",
+          "custom.fields.item.delete",
+          "settings.users.enabled"
```
MaksymDryha: We already have `settings.users.enabled` as a subPermission of `ui-users.settings.customfields.view`, so it can be removed from here

package.json
```diff
@@ -253,6 +253,39 @@
...
+      },
+      {
+        "permissionName": "ui-users.settings.customfields.edit",
+        "displayName": "Settings (Users): Can create, edit, and view custom fields",
+        "subPermissions": [
+          "ui-users.settings.customfields.view",
+          "custom.fields.collection.put",
+          "customfields.item.post",
+          "customfields.item.put"
+        ],
```
MaksymDryha: We also need to add `configuration.entries.item.post` and `configuration.entries.item.put` to the `customfields.edit` subPermissions since editing custom fields also implies the ability to view and edit the custom fields accordion label.

BogdanDenis: Do we also need to add `configuration.etries.item.delete` to `customfields.delete` subPermissions?

PR #1190, opened 2020-03-05
UIU-1518: Fix permissions

annamelnyk:
# Description
Bug fix: Can not assign and unassign permissions to users unless you have "Settings (Users): Can create, edit and remove permission sets"
# Issue
https://issues.folio.org/browse/UIU-1518
# Screenshots


src/views/UserEdit/UserForm.js
```diff
@@ -521,8 +521,8 @@ class UserForm extends React.Component {
                       onToggle={this.handleSectionToggle}
                       headlineContent={<FormattedMessage id="ui-users.permissions.userPermissions" />}
                       permToRead="perms.permissions.get"
```
oleksandr-dekin: Please change `perms.permissions.get` to `perms.users.get`

PR #1196, opened 2020-03-11
UIU-1486 - Default charge and action notices not saved on manual fee/fine charge settings

maximdidenkoepam:
# Purpose
Fix form save where url depends on local resource.

# Link
https://issues.folio.org/browse/UIU-1486

# Screenshot
Screenshot 2020-03-11 at 20 44 53


maximdidenkoepam:
Huge thanks to @mkuklis for help with investigation.

PR #1208, opened 2020-03-18
UIU-1457 restore CommandList, HasCommand

zburke:
`<CommandList>` and `<HasCommand>`, due to a transitive dependency on
`mousetrap`, leaked memory which caused problems in CI when running unit
tests.

This branch restores those wrappers which were removed in #1139.

Refs [UIU-1457](https://issues.folio.org/browse/UIU-1457), [STCOM-636](https://issues.folio.org/browse/STCOM-636)

Replaces #1161. There's no material difference between that PR and this one, but that branch was based on a much earlier version of things and it was easier to just start from scratch instead of resolving a bunch of gnarly diffs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change history for ui-users
 
+## 3.1.0 (IN PROGRESS)
+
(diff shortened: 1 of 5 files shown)


zburke:
Replaced by #1389

PR #1210, opened 2020-03-23
UIU-1511: Remove password field

annamelnyk:
# Description
- On Create New user record should no longer show a password field
- Entering a username should no longer require entering a password
- If username has already been saved to the record then display the Send a create password email link
# Story
https://issues.folio.org/browse/UIU-1511
# Screenshots
image


VictorSoroka1:
Anna, please attach the screenshot with the password link for the completeness of the description.

PR #1211, opened 2020-03-24
Added hasMargin-prop for MCL in users results list

rasmuswoelk:
We made a change in the `<MultiColumnList>`-component that removes the horizontal margin by default, making the component span from edge to edge.

This works for most implementations but MCL's used in the result panes must have the `hasMargin`-prop applied to enable proper spacing and alignment.

This PR adds that prop for the results view.

## Before
image

## After
image

### DIFF
--- a/src/views/UserSearch/UserSearch.js
+++ b/src/views/UserSearch/UserSearch.js
@@ -443,6 +443,7 @@ class UserSearch extends React.Component {
                           isSelected={this.isSelected}
                           autosize
                           virtualize
+                          hasMargin
                         />
 
                       </Pane>


rasmuswoelk:
@zburke The tests are failing but it's not related to this simple change. Can I go ahead and merge?

zburke:
Ugh, tests again. Let me try a late afternoon build when Jenkins tends to be quieter and see if we get lucky. If that doesn't work today, go ahead and merge it tomorrow (Thursday). I get it that this is a low-risk change and the failures are unrelated, but it's really nice to be a good citizen when possible.

zburke:
Midnight build FTW.

PR #1214, opened 2020-03-26
UIU-1540: Add query param

annamelnyk:
# Description
**Bug:** Settings --> Users --> Fee/Fine Owners only displays ten service points in the Associated Service Points drop-down.

**To do:** increase limit of service points in query
# Issue
https://issues.folio.org/browse/UIU-1540
# Screenshot
**Before**

image

**After**

image

### DIFF
--- a/src/settings/OwnerSettings.js
+++ b/src/settings/OwnerSettings.js
@@ -18,7 +18,7 @@ class OwnerSettings extends React.Component {
     ownerServicePoints: {
       type: 'okapi',
       resource: 'service-points',
-      path: 'service-points',
+      path: 'service-points?limit=50',
     },
     owners: {
       type: 'okapi',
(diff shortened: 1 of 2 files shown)


src/settings/OwnerSettings.js
```diff
@@ -18,7 +18,7 @@ class OwnerSettings extends React.Component {
     ownerServicePoints: {
       type: 'okapi',
       resource: 'service-points',
-      path: 'service-points',
+      path: 'service-points?limit=50',
```
zburke: The `owners` resource accepts a limit of `500`. I gotta believe we want at least `100`, maybe `1000`.

annamelnyk: I have asked Holly, and current limit should be increased to 200. Thanks for review, Zak

VictorSoroka1: It is usually a good idea to ask about such things POs.

PR #1218, opened 2020-03-27
UIU-1546 Patron notice not always sent

maximdidenkoepam:
# Purpose
Patron notice not being sent when "Default charge notice" and "Default action notice" in effect.

# Approach
Exclude blank fields from the model before save.

# Link
https://issues.folio.org/browse/UIU-1546


maximdidenkoepam:
@zburke we have a case when user tries to unselect previously chosen charge or/and action notices. In this case ControlledVocab component returns model with empty string as value for unchecked field and BE tries to use this value to get template. In case of POST request model does not contain empty fields even if we check and uncheck them before save.

PR #1219, opened 2020-03-27
UIU-1547 - Incorrectly unable to save Fee/Fine Types already saved by other Fee/Fine Owners

maximdidenkoepam:
# Purpose
Updated check to identify shared user.

# Link
https://issues.folio.org/browse/UIU-1547

### DIFF
--- a/src/settings/FeeFineSettings.js
+++ b/src/settings/FeeFineSettings.js
@@ -89,7 +89,7 @@ class FeeFineSettings extends React.Component {
 
   componentDidMount() {
     this.props.mutator.owners.GET().then(records => {
-      const shared = records.find(o => o.owner === 'Shared');
+      const shared = records.find(o => o?.owner?.toLowerCase() === 'shared');
       this.shared = shared;
       const ownerId = (shared) ? shared.id : ((records.length > 0) ? records[0].id : '');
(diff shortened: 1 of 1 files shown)


zburke:
FYI, I remember having concerns about this implementation when it first showed up [back in September](https://folio-project.slack.com/archives/CHLD3AKU7/p1568472039009300):

> Specifically, tying business logic to a user-editable field (owner.name) seems dangerous; somebody could change the name to “Common Owner” or even “shared” (the current logic is case-sensitive) and unwittingly break things. Likewise, this will fail in any non-English environment when the tenant translates this owner-name to its own locale.

maximdidenkoepam:
@zburke I covered only case described in the ticket. I am not closely familiar with "shared" logic, but I agree that this is a pretty strange idea. If it's needed I can update this check for other cases in this module.

zburke:
The specific circumstance in the ticket description was actually a [backend issue, MODFEE-42](https://issues.folio.org/browse/MODFEE-42) and has been resolved. 

If we are to implement case-insensitive handling of the special user `Shared` we need to do so throughout the entire codebase. That said, this design is brittle and I would rather revisit it wholesale and implement something more robust.

PR #1224, opened 2020-03-31
UIU-1411: prevent UI from crashing when loading closed loans list with a deleted item

Baroquem:
https://issues.folio.org/browse/UIU-1411

This guards the closed loans list against the possibility of an item linked to a loan having been deleted, meaning that item-related information typically displayed in the list is unavailable.


src/components/Loans/ClosedLoans/ClosedLoans.js
```diff
@@ -340,19 +338,21 @@ class ClosedLoans extends React.Component {
 
     const accounts = _.get(this.props.resources, ['loanAccount', 'records'], []);
     const accountsLoan = accounts.filter(a => a.loanId === loan.id) || [];
-    const itemDetailsLink = `/inventory/view/${loan.item.instanceId}/${loan.item.holdingsRecordId}/${loan.itemId}`;
+    const itemDetailsLink = loan.item ? `/inventory/view/${loan.item.instanceId}/${loan.item.holdingsRecordId}/${loan.itemId}` : undefined;
```
SergiyVSergiyenko, line 341: Since we expect the `loan.item` type to be either an `object` or `undefined` (if the `item` property does not exist in the `loan`), and not a `string` or `number`, perhaps it makes sense not to return the `undefined` value explicitly? @Baroquem , what do you think? Can we change this to:
```suggestion
    const itemDetailsLink = loan.item && `/inventory/view/${loan.item.instanceId}/${loan.item.holdingsRecordId}/${loan.itemId}`;
```

Baroquem: @SergiyVSergiyenko I like it; thanks for the suggestion.

PR #1228, opened 2020-04-02
UIU-1553 - Item record not found when clicking on link from Fee/Fine Details

maximdidenkoepam:
# Purpose
Update link to inventory item page.

# Link
https://issues.folio.org/browse/UIU-1553


zburke:
This is out of scope for this PR, but note it is preferable for a11y reasons to use [`<NoValue>`](https://github.com/folio-org/stripes-components/tree/master/lib/NoValue) rather than just a `-` to indicate "no value".

PR #1230, opened 2020-04-02
UIU-1540: Add query param (#1214)

zburke:
Settings > Users > Fee/Fine Owners only displays ten service points
in the Associated Service Points drop-down. Bump the limit to 200.

Fixes [UIU-1540](https://issues.folio.org/browse/UIU-1540)


zburke:
I don't know why SonarCloud suddenly has its knickers in a twist over `checkScope`. It _is_ defined twice (what the ...?) but it's also never called so I'm not gonna worry about it now.

PR #1235, opened 2020-04-05
UIU-1544 - updated recordId entity in manifest (#1217)

zburke:
It's very confusing how to use a local resource that should evaluate to a scalar.
A resource is an object. It may have scalar values, but the resource itself is an 
object. 

Here, trying to embed the resource `%{recordId}` in the path caused an error
because `recordId`, i.e. the resource itself, is an object, and thus would be 
serialized to `[object Object]`. The documentation isn't great on how this is
supposed to work. 

Fixes [UIU-1544](https://issues.folio.org/browse/UIU-1544)


zburke:
Three different runs with three different failures. IOW, they are not likely failures at all but problems with the way the tests run in a CI environment. Damn the torpedoes; full speed ahead!

PR #1236, opened 2020-04-05
UIU-1546 remove empty fields from request before submission

zburke:
Remove empty fields from the request before submitting it. When a field is present, 
its value is expected to contain a foreign key. Thus, empty-string values present a 
problem and should instead be omitted.

Refs [UIU-1546](https://issues.folio.org/browse/UIU-1546)


zburke:
Three different runs with three different failures. IOW, they are not likely failures at all but problems with the way the tests run in a CI environment. Damn the torpedoes; full speed ahead!

PR #1240, opened 2020-04-06
UIU-488 allow for localizable permission names

zburke:
At present, we pull the display names for permission sets out of the
structure defined in `package.json` which doesn't allow for them to be
localized. Instead, the `permissionName` should be used to construct a
translation key and permission names should be displayed just as are any
other embedded value.

NB: this PR _does not handle_ the fact the other modules likely have not created `permission.${permissionName}` values in their respective translation files. 

Additionally, correct sorting of the values in the permission-assignment dialog is left as an exercise to ... somebody else 😉, maybe after we convert to react-intl v3 and have easier access to stringified values? 

Refs [UIU-488](https://issues.folio.org/browse/UIU-488)


src/constants.js
```diff
@@ -28,3 +31,8 @@ export const statusFilter = [
   { label: 'ui-users.filters.status.active', value: 'active' },
   { label: 'ui-users.filters.status.inactive', value: 'inactive' }
 ];
+
+export const renderPermission = (name) => {
+  const [pPrefix, ...pName] = name.split('.');
+  return <FormattedMessage id={`${pPrefix}.permission.${pName.join('.')}`} />;
+};
```
MikeTaylor, line 38: At least in the short term, when almost all modules will not have these translations, I think we need to fall back to the `displayName` in the absence of a translation. That's may be hard to do with `<FormattedMessage>` but is probably relatively easy with `intl.formatMessage`.

MikeTaylor: Also, I don't understand why this function is in a file called `constants.js`. Surely something like `src/util/renderPermission.js` would make more sense?

zburke: You are correct on all fronts. My answer on all fronts is, "I am lazy." I totally recognize that I cherry-picked the easy parts and left the thorny bits un-handled. 

`constants` was there so I used it; sorting looked painful so I skipped it. Maybe there's enough here to get started with for somebody else whose team is actually allocated to this task to finish it off?

MikeTaylor: Oh, it's easy to do this with `<FormattedMessage>` after all, using the `defaultMessage` prop:
```
<FormattedMessage id={`${pPrefix}.permission.${pName.join('.')}`} defaultMessage={...}/>;
};
```

MikeTaylor: I think you're better than that :-)

MikeTaylor:
Although @mkuklis has approved this PR, and I hesitate to go against his judgement, I think we need to be very cautious about it in its present state. It currently introduces two actual regressions with respect to the present behaviour:
* In non-English domains, permissions for nearly all modules will change their apparent names, with the change itself introducing some confusion, and with the changed names being in at least some cases too technical for end-users to make much sense of.
* The permission list will not be sorted.

I think both of these are relatively easy to fix, and we should do so before merging.

src/constants.js
```diff
@@ -28,3 +31,8 @@ export const statusFilter = [
   { label: 'ui-users.filters.status.active', value: 'active' },
   { label: 'ui-users.filters.status.inactive', value: 'inactive' }
 ];
+
+export const renderPermission = (name) => {
```
mkuklis, line 35: @zburke do you think we can move this somewhere under: https://github.com/folio-org/ui-users/tree/master/src/components/util

The `constants` file seems like a good place for dictionary/list/constants.

zburke: oh fer sure. 

TBH, this was more of a "How close can I get with less than an hour's work" POC than an "I'm all done, here it is." PR. I probably should've just pushed a branch rather than actually issued the PR.

zburke:
You're right about the regressions, @MikeTaylor. See Ann-Marie's comments on [the ticket](https://issues.folio.org/browse/UIU-488) re: the grunt work of getting the display-names into translation files. 

It's relatively easy to judge the completeness of that work for platform-core and platform-complete UI modules. For other modules we are yet unaware of, there's really no way to know when they've been tweaked. Feels like maybe doing this all in ui-users is kinda wrong-headed from that point of view and this should be done in stripes-core where we could increment the major version number.

MikeTaylor:
I think doing it in ui-users is fine — let's just do it without regressions.

doytch:
I rejigged this a bit. I originally wanted to just do this where we do package translations but then the cross-referencing of that vs the permissions record inside ui-users would be disgusting so I kept @zburke's main thrust.

But I: 
- pulled the string generation out into a data converter called `getPermissionLabelString`
- used that inside a `PermissionLabel` component for ease of use when you don't _need_ a JS string
- used `getPermissionLabelString` for sorting purposes
  - bonus: localeCompare them instead of a naive `>` check

**Here's the wrinkle right now though.** Even when you pass in a `defaultMessage`, react-intl throws an error on the console if the translation message doesn't exist. This results in _thousands_ of errors on the console when you view diku_admin rn.

[We _can_ pass in an `onError` to `IntlProvider`](https://github.com/formatjs/react-intl/blob/master/docs/Components.md#intlprovider) and filter out those errors, or just live with them for now. Or we can do my personal preference: filter them out for now, create a JIRA to remove that filter in 2-3 months, and berate anybody who didn't translate their permission names by then.

doytch:
https://github.com/folio-org/stripes-core/pull/831 has been created to suppress the React Intl errors. STCOR-424 has been filed to _remove_ that suppression at a later date.

src/components/data/converters/permission.js
```diff
@@ -0,0 +1,12 @@
+/* eslint-disable import/prefer-default-export */
+
+export function getPermissionLabelString(permission, formatMessage, showRaw) {
```
MikeTaylor, line 3: In an otherwise elegant solution, it's an unsightly wart that this function needs to have `formatMessage` passed into it — apart from anything else, it requires us to `injectIntl` into the components that use it just so they can obtain the `formatMessage` function. Is there really no way for this function to find that for itself?

doytch: Unfortunately...no. The issue here is that we eventually need to return a plain JS string. The only way to get that out of react-intl is via `formatMessage`. react-intl only makes _that_ available via React-y methods because all of its translations, locale-data, etc is stored that way and needs to be...Reactive.

Because everything is only available via React-y methods, there's literally no way to get a translated string from a basic execution context like that function. We'd need to get at react-intl first via something like its provided components or injectIntl or useIntl. But then we're in a React component, and we can't return a plain string from that, we can only return a React element. Sigh!

I could've avoided passing formatMessage, but then what happens is that you move some parts of the formatting (eg, the `showRaw` ternary) out of that function. So that's not great either. :\

doytch: I really do hate that I have to pass in formatMessage so would _definitely_ appreciate a better solution. I just have never found one. I actually ran into this same issue a while ago in ERM code and poked around for a while without finding anything better. :(

doytch: Also, it should be noted that unless a component _absolutely, positively_ wants a `String` copy of the permission label, it doesn't need to injectIntl/useIntl. If it just wants to render the permission label, it should just render the `PermissionLabel` component.

MikeTaylor: Yes, all of that sounds right. It's mildly disappointing that the way the underlying APIs are set up doesn't allow us to do this more elegantly, but given the lack of alternatives I don't think we need to agonise too much.

PR #1244, opened 2020-04-08
do not use hard-coded expiration dates

zburke:
The user expiration-date was originally hard-coded to `2020-04-07`. What
do you know, today, tests started failing! Who knew that would happen!?!

Thanks to @mkuklis for fixing this on master in #1212 so I could easily
steal the fix and apply it here.


zburke:
After three different failures, or "succeeded but timed out", I gave up and merged anyway.

PR #1249, opened 2020-04-10
UIU-1532: Add 'Load more' button at the end of the result list

SergiyVSergiyenko:
## Purpose
Add `Load more` button at the end of the result list in order to scroll records more easily.
Story [UIU-1532](https://issues.folio.org/browse/UIU-1532).

### Approach
To achieve the needed functionality, in addition to the steps described in the [Migration Paths](https://github.com/folio-org/stripes-connect/blob/master/MIGRATIONPATHS.md), it was necessary to override the `onNeedMoreData` method in the `UserSearchContainer` component.
Thanks, @ryandberger, for pointing this out.

### Screenshot
Screen Shot 2020-04-10 at 21 50 19

### DIFF
--- a/src/views/UserSearch/Filters.js
+++ b/src/views/UserSearch/Filters.js
@@ -24,6 +24,9 @@ export default class Filters extends React.Component {
     activeFilters: PropTypes.object,
     resources: PropTypes.object.isRequired,
     onChangeHandlers: PropTypes.object.isRequired,
+    resultOffset: PropTypes.shape({
+      replace: PropTypes.func.isRequired,
+    }),
   };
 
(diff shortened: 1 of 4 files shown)


src/views/UserSearch/Filters.js
```diff
@@ -48,8 +51,13 @@ export default class Filters extends React.Component {
     const {
       activeFilters,
       onChangeHandlers,
+      resultOffset,
     } = this.props;
 
+    if (resultOffset) {
+      resultOffset.replace(0);
```
ryandberger, line 58: Excellent, I see you were able to reset the offset when changing filters.

PR #1252, opened 2020-04-14
UIU-1167: Create Patron Blocks Limits

annamelnyk:
# Description
Add Patron Blocks section with Limits option to Settings -> Users -> Patron Blocks -> Limits
# Story
https://issues.folio.org/browse/UIU-1167
# Screenshots
L1
L2
L3


src/settings/patronBlocks/Limits/Limits.js
```diff
@@ -0,0 +1,203 @@
...
+  }
+
+  render() {
+    const {
+      patronGroup,
+      patronGroupId,
+      patronBlockConditions,
+    } = this.props;
+
+    if (!this._isMounted) return null;
```
VictorSoroka1, line 183: I am wondering why it is needed. As I see it can be added to every component in that way. Did you have some error messages/warning that made you do it?

src/constants.js
```diff
@@ -28,3 +28,5 @@ export const statusFilter = [
   { label: 'ui-users.filters.status.active', value: 'active' },
   { label: 'ui-users.filters.status.inactive', value: 'inactive' }
 ];
+
+export const feeFineBalanceId = 'cf7a0d5f-a327-4ca1-aa9e-dc55ec006b8a';
```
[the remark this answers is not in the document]
annamelnyk: With current id I determine that this is fee/fine condition, because this condition is validating with different message, and condition fields are rendering dinamically. All conditions (thare are 6 of them) are always present on BE with hardcoded ids for now.

src/settings/patronBlocks/Limits/Limits.js
annamelnyk: I have not any errors. I just wanted to prevent possibility of memory leak, it is recommended for components with state, isn't it. Or do you think I should remove `_isMounted`?

VictorSoroka1: I think so, usually such things are added on purpose. As I said, in that way we can introduce such things nearly to every component. But I leave it up to you)

package.json
```diff
@@ -565,6 +565,19 @@
         ],
         "visible": true
       },
+      {
+        "permissionName": "ui-users.settings.limits",
+        "displayName": "Settings (Users): Can create, edit and remove patron blocks limits",
```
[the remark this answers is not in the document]
annamelnyk: Thanks for review. I added permission name into translations

PR #1267, opened 2020-04-20
UIU-1594: Settings > Users > Custom Fields: Display Edit form in 3rd pane in settings

BogdanDenis:
## Purpose
- Move Custom Fields Edit form route to Settings to fix routing loop in Users app

## Issue
[UIU-1594](https://issues.folio.org/browse/UIU-1594)

## Screenshots

CustomFieldsSettings
image


src/settings/CustomFieldsSettings.js
```diff
@@ -26,17 +33,33 @@ const CustomFieldsSettings = ({
...
+        />
+      </Route>
+      <Route exact path={`${base}/edit`}>
+        <EditCustomFieldsSettings
+          backendModuleName={backendModuleName}
+          entityType={entityType}
+          redirectToView={redirectToView}
+          permissions={permissions}
+        />
+      </Route>
```
zburke, line 61: Is there no way to get these moved up to `src/settings/sections.js` where all other settings routes are defined?

src/settings/CustomFieldsSettings.js
```diff
@@ -16,8 +17,14 @@ const CustomFieldsSettings = ({
...
+  const base = '/settings/users/custom-fields';
+
   const redirectToEdit = () => {
-    history.push('/users/custom-fields/edit');
+    history.push(`${base}/edit`);
+  };
+
+  const redirectToView = () => {
+    history.push(`${base}`);
   };
```
zburke, line 28: These get rendered as `<Buttons>` in SSC; can't we just use the `to` prop and let `react-router` take care of history manipulation?

src/settings/CustomFieldsSettings.js
BogdanDenis: In this case there would be two links for Custom Fields in the 2nd pane, like so:
image
I was thinking of adding these routes to `sections`, but also defining a new `hidden` property for them so the edit route is not shown in the list.

src/settings/CustomFieldsSettings.js
BogdanDenis: Additional PR for this change: https://github.com/folio-org/stripes-smart-components/pull/778

src/settings/CustomFieldsSettings.js
zburke: Hmm; I see why we don't want that. I guess we handle this elsewhere by appending a different query-string to a route, instead of using separate route paths, is that right? If so, is that possible here?

zburke:
The four test failures all appear related specifically to custom fields. Can you take a look?

src/settings/CustomFieldsSettings.js
BogdanDenis: Some apps do use a query string for separate routes, but in combination with the `<Layer>` component. So if we're not using `<Layer>` there's not much difference between route paths and query parameters

zburke:
Do you know what changed that caused those tests to pass?

PR #1279, opened 2020-04-27
UIU-1609: Refactoring of Patron Block Conditions

annamelnyk:
# Description
Refactoring of Patron Blocks:
- remove hardcoded ids and some code
- remove duplicated css code and move it in single file

# Story
https://issues.folio.org/browse/UIU-1609


annamelnyk:
@zburke regarding to our discussion about `patron block conditions`, from BE side currently will be nothing changed, because patron blocks will be divide into separate module, and then will be discussed what methods can be used in future, and now API is implemented with REST. So we deсided to remove hardcoded ids from UI and to do some refactoring

PR #1282, opened 2020-04-28
UIU-1604: Preserve filters after user is updated

mkuklis:
Preserve filters after user record is successfully updated:

https://issues.folio.org/browse/UIU-1604

### DIFF
--- a/src/views/UserEdit/UserEdit.js
+++ b/src/views/UserEdit/UserEdit.js
@@ -206,7 +206,10 @@ class UserEdit extends React.Component {
       history,
       resources,
       match: { params },
-      location: { state },
+      location: {
+        state,
+        search,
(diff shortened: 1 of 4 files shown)


src/views/UserEdit/UserEdit.js
```diff
@@ -249,9 +252,11 @@ class UserEdit extends React.Component {
       data.active = (moment(user.expirationDate).endOf('day').isSameOrAfter(today));
     }
 
+    console.log('user edit search', search);
```
Baroquem, line 255: Don't forget to remove this!

mkuklis: Thank you for catching this @Baroquem!

PR #1302, opened 2020-05-14
UIU-1608: Display create/reset password link

annamelnyk:
# Description
When Edit user ---> show correct link `create` or `reset` password
# Approach
We need to differ the user WITH setted password from user WITHOUT password. When user was created by default setted password is `''` (empty string). When we `edit` user, we receive the object with `credentials`. Then we encrypt the `''`(empty string) with received `salt` from `credentials`, and if `hash` is simular --> then we kow that password was not set for the current user yet --> display `create password link`  
# Task
https://issues.folio.org/browse/UIU-1608
# Screenshot
image


test/bigtest/network/config.js
```diff
@@ -613,4 +613,29 @@ export default function config() {
...
+
+  this.put('/credentials/:id', ({ users }, request) => {
+    return users.find(request.params.id).attrs;
+  });
+
+  this.get('/credentials/:id', ({ users }, request) => {
+    return users.find(request.params.id).attrs;
+  });
+
+  // this.get('credentials/:userId', {
```
VictorSoroka1, line 630: Should this be removed?

annamelnyk: Sure, definitely. I still continue to work on tests.

annamelnyk:
@zburke, thanks for review. I will add comment to code. 
Password is hashing on BE with `PBKDF2WithHmacSHA1` algoryithm

zburke:
@annamelnyk, you've done great work here, but there I have some broad concerns about how the UI is handling password management. This isn't related to this PR specifically, but I think we need to put it on hold for a little bit before we figure out the right path forward. 

Thanks for your patience.

mkuklis:
@annamelnyk This all looks good but it's very concerning to see the password encryption happening on the client:

https://github.com/folio-org/ui-users/pull/1302/files#diff-2f6c8eeed7317177c134543e798c2ecdR96

I hope this can be managed by the server.

annamelnyk:
@mkuklis thanks for review. But basically there is no any threat for security, nothing is encrypted on the client. Just checking the hash to compare if the password matches to empty string.

annamelnyk:
As we discussed with team again, ui is the appropriate place to check the hash, otherwise why do we receive the hash on client side, - to make such checks. The other question is if algorythm would be changed, then changes shoud be on UI and BE as well, then 2 tickets might be created. @zburke, @mkuklis what do you think, guys? This is necessary functionality, so Il wil be waiting for your conclusion.

mkuklis:
@annamelnyk why can't this be done on the server?

mkuklis:
@annamelnyk @zburke maybe the best way here is to move forward with this and review it again later? It still feels uneasy that we are sending salt and hash back to the client. I wish this could be done on the server but I also understand that we want to move forward.

src/components/EditSections/EditExtendedInfo/CreateResetPasswordControl/CreateResetPasswordControl.js
```diff
@@ -1,7 +1,10 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { FormattedMessage } from 'react-intl';
-import { get } from 'lodash';
+import { isEmpty } from 'lodash';
+// eslint-disable-next-line import/no-extraneous-dependencies
+import crypto from 'crypto';
```
VictorSoroka1, line 6: Should it be included in dependency then?

src/components/EditSections/EditExtendedInfo/CreateResetPasswordControl/CreateResetPasswordControl.js
```diff
@@ -37,17 +40,73 @@ class CreateResetPasswordControl extends React.Component {
...
+    } = this.props;
+    const credentialsRecord = credentials?.records[0]?.credentials[0] ?? {};
+
+    if (isEmpty(credentialsRecord)) {
+      return;
+    }
+
+    const normalizeSalt = Buffer.from(credentialsRecord.salt, 'hex');
+
+    await crypto.pbkdf2('', normalizeSalt, 1000, 32, 'sha1', (err, derivedKey) => {
```
VictorSoroka1, line 96: I think the comment is needed here to describe what is logic beyond this.

```
check that the user has no password set by comparing user hashed password with the hashed empty string which

src/components/EditSections/EditExtendedInfo/CreateResetPasswordControl/CreateResetPasswordControl.js
annamelnyk: `crypto` is a built-in Node module, it should not be include as dependency

VictorSoroka1: Ok, but the code will be running in browser environment and not in node. Please check that the production build would contain the package.

ViktorSoroka: It seems like webpack handles it - https://webpack.js.org/configuration/node/. I have not tested though.

src/components/EditSections/EditExtendedInfo/CreateResetPasswordControl/CreateResetPasswordControl.js
VictorSoroka1, line 96: I suggest calling then pbkdf2Sync so there won't be the need for the await stuff.

NikitaSedyx:
@mkuklis I agree with you, it shouldn't be done on client side - now we know hash, salt and how password is hashed, it's super easy to get password

as it was mentioned, it should be done on server side

annamelnyk:
@NikitaSedyx, I guess it is never can be super easy, with hash and salt impossible crack the password, but password can be just picked up with selection method, but then it can be done with any encryption algoryth, further is the question of complexity. 
I am wondering, for what purposes method `GET` for `/authn/credentials/{id}` does exist then? [(link)](https://s3.amazonaws.com/foliodocs/api/mod-login/p/login.html#authn_credentials__id__get)

ViktorSoroka:
@annamelnyk, could you please clarify why the `path: 'authn/credentials-existence?userId=!{userId}',` was not suitable for that case?

annamelnyk:
@ViktorSoroka, sure, I already get to used that path in this PR. My question was about for what purposes the current endpoint exist if I can not use `hash` and `salt` on ui

annamelnyk:
@ViktorSoroka, `path: 'authn/credentials-existence?userId=!{userId}'` unfortunately is not suitable, because it returns `true`, if `username` was set, not the `password`

zburke:
@annamelnyk, regarding your question: 

> for what purposes the current endpoint exist if I can not use `hash` and `salt` on ui

In fact, other folks [have concluded](https://issues.folio.org/browse/MODLOGIN-128) that endpoint should not exist, hence #1344.

PR #1303, opened 2020-05-14
MODUSERBL-89 remove custom fields routes to prevent carnage

zburke:
Users with custom fields attached cannot login at present. This creates
problems in the production environments if folks happen to add
custom-fields values to common users like `diku_admin`. Until we resolve
this larger issue, we are disabling the custom-fields routes to prevent
this common problem.

Refs [MODUSERBL-89](https://issues.folio.org/browse/MODUSERBL-89)


zburke:
Oh, right, I probably should have disabled the custom fields tests since I commented out custom fields routes! 🤦

PR #1304, opened 2020-05-15
Revert "MODUSERBL-89 remove custom fields routes to prevent carnage"

zburke:
Reverts folio-org/ui-users#1303

Restore custom fields just as soon as we figure out how to git mod-user-bl to cope with them so accounts with custom fields can still authenticate.


zburke:
Replaced by #1308.

PR #1306, opened 2020-05-15
don't leak memory after unmount

zburke:
If a promise returns after a component has unmounted, you can no longer
call `setState()`, so now we don't.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@
 * Correctly configure fee/fines and profile-pictures permissions. Refs UIU-1574.
 * Increment `stripes` to `v4.0`, `react-intl` to `v4.5`, `react-intl-safe-html` to `v2.0`. Refs STRIPES-672.
 * Fix `Reset all` button for filters and query. Fixes UIU-1628.
+* Fix memory leak in `UserAccounts.js` discovered by @mkuklis.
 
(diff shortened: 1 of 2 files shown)


NikitaSedyx:
@zburke usually signals (in case of http requests) are used to abort, is it possible to do this with stripes-connect?

zburke:
@NikitaSedyx, I like the idea! 

The problem here is simply that a fetch can return after a component unmounts, which causes calls to `setState` to fail. I looked a bit at [signals](https://developer.mozilla.org/en-US/docs/Web/API/AbortController/signal) and while I see how they could be used to cancel a request, or discover that one has been canceled, my initial take is that we would still have to implement the same sort of case-by-case handler as we have here. The fetch isn't being canceled; it's just returning after unmount. 

Do you see a way it could be handled universally?

NikitaSedyx:
@zburke basically we should handle only `catch` section with request cancel, I don't think there is a good way to have generic handler as it depends how promise is processed, but with signals we can have less `if` blocks

VictorSoroka1:
Is `signals` used somewhere on folio already? I think this might be a good topic for stripes-architecture as I am sure fetch is used in so many places already. Maybe it makes sense to try out makeCancelable which is proposed in the react docs at the end of the article.

https://reactjs.org/blog/2015/12/16/ismounted-antipattern.html

NikitaSedyx:
`makeCancelable` sounds good to me

PR #1307, opened 2020-05-19
UIU-1648 correct state init in withServicePoints

zburke:
The original implementation did not correctly initialize state within a
constructor, leading to a memory leak that was notably apparent in unit
tests. Heap memory would balloon > 700MB, leading to failed tests due
to timeouts. With state correctly initialized, heap memory is reduced by
roughly 90% to ~55MB.

Post-fix, there are 1365 fewer copies of `Wrapper` components (`Wrapper` is the name of the class returned from `stripes-connect` when a component is connected) in the heap when unit tests finish running: 

Screen Shot 2020-05-18 at 9 37 37 PM
Screen Shot 2020-05-18 at 9 39 13 PM

Refs [FOLIO-2097](https://issues.folio.org/browse/FOLIO-2097), [UIU-1648](https://issues.folio.org/browse/UIU-1648)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@
 * Correctly configure fee/fines and profile-pictures permissions. Refs UIU-1574.
 * Increment `stripes` to `v4.0`, `react-intl` to `v4.5`, `react-intl-safe-html` to `v2.0`. Refs STRIPES-672.
 * Fix `Reset all` button for filters and query. Fixes UIU-1628.
+* Don't leak memory in `withServicePoints` wrapper. Refs FOLIO-2097.
 
 ## [3.0.0](https://github.com/folio-org/ui-users/tree/v3.0.0) (2020-03-17)
 [Full Changelog](https://github.com/folio-org/ui-users/compare/v2.26.0...v3.0.0)
--- a/src/components/Wrappers/withServicePoints.js
+++ b/src/components/Wrappers/withServicePoints.js
@@ -69,10 +69,14 @@ const withServicePoints = WrappedComponent => class WithServicePointsComponent e
       stripes: stripesShape.isRequired,
     };
 
-    state = {
-      userServicePoints: [],
-      userPreferredServicePoint: undefined,
-    };
+    constructor(props) {
+      super(props);
+
+      this.state = {
+        userServicePoints: [],
+        userPreferredServicePoint: undefined,
+      };
+    }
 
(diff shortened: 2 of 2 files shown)


mkuklis:
We don't understand yet why this is happening and the underlining issue can be actually something different but since we do see the difference in memory usage I would say let's merge this in.

doytch:
@zburke do you have a way of reproducing/methodology for collecting your memory statistics? I couldn't come up with anything great because of the performance implications of instrumented collection. 

I did watch the heap allocation counter along with the tab's memory during a runthrough of `yarn test` in ui-users, like so. (This is a run off of this branch)
Screen Shot 2020-05-19 at 1 46 30 PM

I found no discernible difference between this branch and just swapping the arrow function to a traditional function off of `master`. In fact, despite the fact that I reran it a few times, my arrow-to-function swap seemed to run at a lower memory footprint. But given the vagaries of sampling intervals, I'm hesitant to buy-in too much.

zburke:
The method: Run the suite of unit tests, then college garbage, then take a heap snapshot.

zburke:
As [noted on the ticket](https://issues.folio.org/browse/UIU-1648): 
> This was not the memory leak [I thought it was](https://folio-project.slack.com/archives/C210UCHQ9/p1589896813080000). At the end of that thread, note: 
> 
> So it turns out React keeps a reference to components that call cWRP so it can provide precisely one warning per component when they call that method. IOW, STCON-70, via [PR #147](folio-org/stripes-connect/pull/147), renaming cWRP to UNSAFE_cWRP solved this problem.

VictorSoroka1:
@zburke, it would be great if you notify folks in the community about this finding with mentioning that changes in this PR are not relevant for tuning performance.

zburke:
Good thinking, @VictorSoroka1. I had already updated the [Slack#stripes thread](https://folio-project.slack.com/archives/C210UCHQ9/p1589896813080000) where most of the discussion took place, and added the following in the main channel: 
> I made a bunch of noise in this channel yesterday about potential performance problems related to how you initialize state. Long story short, state init patterns were a red herring. There was a performance issue, but it was related to something else entirely: if your component uses `componentWillReceiveProps` instead of `UNSAFE_componentWillReceiveProps`, react will issue a console warning about it, and, crucially, _create a reference to that component so you don’t get warned more than once_. This reference is the source of the leak.

PR #1308, opened 2020-05-19
MODUSERBL-89: Enable Custom Fields routes and tests

BogdanDenis:
Enable Custom Fields routes and tests after resolving issue with User schema deserialization

## Issue
[MODUSERBL-89](https://issues.folio.org/browse/MODUSERBL-89)


zburke:
Reverts #1303, Replaces #1304.

PR #1309, opened 2020-05-19
UIU-1353: Sort permissions

annamelnyk:
# Description
Sort user permissions in alphabetical order on create/edit screen.
# Task
https://issues.folio.org/browse/UIU-1353
# Screenshots
**Before**
image

**After**
image

### DIFF
--- a/src/components/PermissionsAccordion/PermissionsAccordion.js
+++ b/src/components/PermissionsAccordion/PermissionsAccordion.js
@@ -6,6 +6,7 @@ import {
 } from 'redux-form';
 import { connect as reduxConnect } from 'react-redux';
 import { FormattedMessage } from 'react-intl';
+import { sortBy } from 'lodash';
 import {
   Icon,
   Button,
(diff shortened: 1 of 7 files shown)


src/components/PermissionsAccordion/PermissionsAccordion.js
```diff
@@ -115,8 +116,9 @@ class PermissionsAccordion extends React.Component {
   }
 
   renderList = ({ fields }) => {
+    const sortedFields = sortBy(fields.getAll(), 'displayName');
     const listFormatter = (_fieldName, index) => (
-      this.renderItem(fields.get(index), index, this.props.stripes?.config?.showPerms)
+      this.renderItem(sortedFields[index], index, this.props.stripes?.config?.showPerms)
```
zburke, line 121: Sadly, it's not so simple; this will always sort by the (deprecated) US-English value from `package.json`. If you take a look at `this.renderItem()`, it's implemented in terms of `<PermissionLabel>` which in turn is implemented in terms of [getPermissionLableString](https://github.com/folio-org/ui-users/blob/132c7a0542f8366738bb55cf8a50f0af6d6db8a7/src/components/data/converters/permission.js) which extracts the localized value. 

I don't know if there's an easy way to avoid iterating through the list an extra time. We're not talking about 10k permissions here; iterating twice may will be the easiest approach.

doytch: Yeah, I think the `.sort()` needs to be [something like this](https://github.com/folio-org/ui-users/blob/fb703df781181db92abb0f1c4701ceec6914ca66/src/components/RenderPermissions/RenderPermissions.js#L51-L56).

doytch: That function could be sped up a bit by running through the list once and memoizing the return value from `getPermissionLabelString` before sorting, but like Zak said...it's a relatively small list and I was lazy.

PR #1314, opened 2020-05-20
Make moment a peerDependency

doytch:


doytch:
Dear Jenkins, thank you for always brightening my morning with a letter in my notifications. It's always fun to read what's going on with you. Sorry I haven't been as good a pen pal lately but it's been kind of busy out here in meatspace. I hope to grow as a person and be as steadfast and consistent as you.

PR #1322, opened 2020-05-26
UIU-1642: When using screenreader, Patron Group, Status, and Preferred contact fields, are not read a required 

BogdanDenis:
## Description
When using a screen reader while creating or editing a user, required fields Patron Group, Status and Preferred contact are not read as required

## Approach
This is simular to what @annamelnyk has done in her PR https://github.com/folio-org/ui-users/pull/1315. Only difference is that `required` prop is now passed to these fields, which automatically adds a red asterisk and more importantly notifies screen readers that this is a required field

## Issue
[UIU-1642](https://issues.folio.org/browse/UIU-1642)

## Screenshots
image


src/components/EditSections/EditContactInfo/EditContactInfo.js
```diff
@@ -84,16 +83,12 @@ const EditContactInfo = ({
         </Col>
         <Col xs={12} md={3}>
           <Field
-            label={(
-              <>
-                <FormattedMessage id="ui-users.contact.preferredContact" />
-                <Asterisk />
-              </>
-            )}
+            label={<FormattedMessage id="ui-users.contact.preferredContact" />}
```
annamelnyk, line 86: @BogdanDenis, I see that you just removed my changes. I did not put `required` attribute because those fields (`preferredContact`, `patronGroup`, `status`) are not `required` on BE side, so I guess it is not a good idea to make it `requred`. Please clarify this changes with BE team.

annamelnyk: Maybe you should try `aria-required="true"`, then screen reader will read this fields, but it can stay without `required` attribute. But I did not test it.

BogdanDenis: @annamelnyk `preferredContact` and `patronGroup` are not marked as `required`, but in [`validate` function in `UserForm` component](https://github.com/folio-org/ui-users/blob/master/src/views/UserEdit/UserForm.js#L60) they are still being checked for empty value, so in theory the behavior hasn't changed.
If we need to make them nor required in UI, then we'll have to remove these checks in validation as well.
Anyways, I'll raise this with our BE team and PO (I think Khalilah is PO in your team also, right?)

annamelnyk: No, Khalilah is not our PO any more. 
But in my [ticket](https://issues.folio.org/browse/UIU-1381) was left the comment with suggestion not to make fields `required`, like they shoud stay "conditionally" required.
I guess the right way discuss it with BE team.

BogdanDenis:
@annamelnyk @zburke a decision has been made by User Management SIG to keep these field required on the front-end, but not required on back-end ([comment by Khalilah Gambrell](https://issues.folio.org/browse/UIU-1642?focusedCommentId=78622&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-78622))
For now it seems that it's too much of a breaking change to make them required on the back-end, so I was asked to proceed with the work as described in the ticket

annamelnyk:
@BogdanDenis, great that this issue is resolved now. Thanks.

PR #1332, opened 2020-05-28
UIU-496 use local placeholder image

zburke:
Use a local placeholder image rather than a remote one to circumvent
some peculiarities of GDPR related to logging of IP addresses by
third-parties (of which placeholder.it is one).

Refs [UIU-496](https://issues.folio.org/browse/UIU-496)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [4.0.0] (IN PROGRESS)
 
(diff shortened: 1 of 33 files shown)


zburke:
True confessions, @JohnC-80, this is quick and dirty. Would be less dirty to use the app-icon's SVG if there is any easy way to do that? Or should we create some other placeholder image?

PR #1340, opened 2020-06-03
UIU-1652: Protect loan action history page from CQL null query errors

Baroquem:
https://issues.folio.org/browse/UIU-1652

This is a difficult bug to reproduce as reported, since it apparently stems from an abnormal condition in which a loan is missing either an itemId or a loanPolicyId. But in artificial tests (i.e., deliberately breaking things), filtering out the null values before mapping the arrays seems to protect against the CQL errors.


SergiyVSergiyenko:
Good catch! Cool! 👍

PR #1341, opened 2020-06-04
UIU-1600: Add interface checks for notes, feesfines and circulation

mkuklis:
This PR is trying to address issue described in: 

https://issues.folio.org/browse/UIU-1600

The user's details screen and edit form should now work without throwing bunch of errors when the `feefines` and `notes` interfaces are not present.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,14 +34,17 @@
     "okapiInterfaces": {
       "users": "15.0",
       "configuration": "2.0",
-      "circulation": "9.0",
       "permissions": "5.0",
       "loan-policy-storage": "1.0 2.0",
       "loan-storage": "4.0 5.0 6.0 7.0",
       "login": "6.0",
-      "feesfines": "15.0",
       "request-storage": "2.5 3.0",
       "users-bl": "5.0"
+    },
+    "optionalOkapiInterfaces": {
+      "circulation": "9.0",
+      "feesfines": "15.0",
+      "notes": "1.0"
     },
     "permissionSets": [
       {
--- a/src/views/UserDetail/UserDetail.js
+++ b/src/views/UserDetail/UserDetail.js
@@ -444,19 +444,21 @@ class UserDetail extends React.Component {
                 expanded={sections.userInformationSection}
                 onToggle={this.handleSectionToggle}
               />
-              <IfPermission perm="manualblocks.collection.get">
-                <PatronBlock
-                  accordionId="patronBlocksSection"
-                  user={user}
-                  hasPatronBlocks={(hasPatronBlocks === 1 && totalPatronBlocks > 0)}
-                  patronBlocks={patronBlocks}
-                  expanded={sections.patronBlocksSection}
-                  onToggle={this.handleSectionToggle}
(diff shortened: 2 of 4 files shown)


src/views/UserDetail/UserDetail.js
```diff
@@ -444,19 +444,21 @@ class UserDetail extends React.Component {
...
-                  patronBlocks={patronBlocks}
-                  expanded={sections.patronBlocksSection}
-                  onToggle={this.handleSectionToggle}
-                  onClickViewPatronBlock={this.onClickViewPatronBlock}
-                  addRecord={this.state.addRecord}
-                  {...this.props}
-                />
-              </IfPermission>
+              <IfInterface perm="feefines">
+                <IfPermission perm="manualblocks.collection.get">
```
MikeTaylor, line 448: It's a real shame that everything inside the `<IfInterface>` has to be re-intented, so it's hard to see what has actually changed.

Oh well  ¯\\\_(ツ)_/¯

src/views/UserDetail/UserDetail.js
```diff
@@ -444,19 +444,21 @@ class UserDetail extends React.Component {
...
-                  hasPatronBlocks={(hasPatronBlocks === 1 && totalPatronBlocks > 0)}
-                  patronBlocks={patronBlocks}
-                  expanded={sections.patronBlocksSection}
-                  onToggle={this.handleSectionToggle}
-                  onClickViewPatronBlock={this.onClickViewPatronBlock}
-                  addRecord={this.state.addRecord}
-                  {...this.props}
-                />
-              </IfPermission>
+              <IfInterface perm="feefines">
```
MikeTaylor, line 447: Surely `<IfInterface>` doesn't take a `perm` prop?!

mkuklis: good catch  @@MikeTaylor!

src/views/UserDetail/UserDetail.js
mkuklis: @MikeTaylor I fixed the `perm` => `name` issue.

I'm not sure if we currently have anything in place to mark interfaces as optional do we? Does something like `okapiOptionalInterfaces` exist?

mkuklis: Ah we do have `optionalOkapiInterfaces` :)

package.json
```diff
@@ -34,15 +34,18 @@
...
       "loan-storage": "4.0 5.0 6.0 7.0",
       "login": "6.0",
-      "feesfines": "15.0",
       "request-storage": "2.5 3.0",
       "users-bl": "5.0"
     },
+    "optionalOkapiInterfaces": {
+      "circulation": "9.0",
+      "feesfines": "15.0",
+      "notes": "1.0"
```
MikeTaylor, line 47: Oh, good catch on `notes`!

src/views/UserDetail/UserDetail.js
MikeTaylor: Yep, I added `optionalOkapiInterfaces` myself! :-)

PR #1345, opened 2020-06-04
UIU-1653: Check for an empty loan to protect from generating incorrect CQL

mkuklis:
This is a continuation of work @Baroquem did under https://github.com/folio-org/ui-users/pull/1340.

It looks like in some circumstances an empty loan object is being created here:

https://github.com/folio-org/ui-users/blob/3a3a2cf25082cb8dc8260b64a41ffec94a3f2071/src/routes/LoanDetailContainer.js#L129

Which was causing bunch of issues in `withRenew` (because the `loans` array contained one empty loan object):

https://github.com/folio-org/ui-users/blob/2f10ab037d7fb1704e6ec09441127b502cefdcf9/src/components/Wrappers/withRenew.js#L83

https://github.com/folio-org/ui-users/blob/2f10ab037d7fb1704e6ec09441127b502cefdcf9/src/components/Wrappers/withRenew.js#L63

This PR should address it.

### Link

https://issues.folio.org/browse/UIU-1653

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,7 @@
 * Protect loan action history page from CQL null query errors. Fixes UIU-1652.
 * Fix the arrangement of elements inside the `SafeHTMLMessage` component. Fixes UIU-1660.
 * Fix Patron Group, Status, and Preferred contact fields, are not read a required by screen reader. Refs UIU-1642.
+* Check for an empty loan to protect from generating incorrect CQL. Fixes UIU-1653.
 * Add checks for multiple okapi interfaces on user's details screen. Fixes UIU-1600.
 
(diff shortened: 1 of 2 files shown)


SergiyVSergiyenko:
On the one hand, it’s a little disappointing when the work done is expressed in one line of code. On the other hand, we can immediately see the hand of the master! Thank you @mkuklis!

zburke:
That is smart debugging you two! I see how this fix resolves the problem by avoiding pushing that empty object onto the array. But, moving one step up, what is the value of assigning that empty object in the first place? If there is no loan, shouldn't we just set it to `null` instead of `{}`?

mkuklis:
That's a very good question which perhaps deserves another JIRA :)

PR #1367, opened 2020-06-16
prefer aria-label to ariaLabel per documentation

zburke:
[The React WAI-ARIA docs say so](https://reactjs.org/docs/accessibility.html#wai-aria).


mkuklis:
Thanks for sharing the official accessibility doc from react @zburke. I was not aware of it.

PR #1368, opened 2020-06-16
fix custom fields payload formatting

MaksymDryha:
When a custom field has no value, the backend expects us to not include that field in the payload.

### DIFF
--- a/src/views/UserEdit/UserEdit.js
+++ b/src/views/UserEdit/UserEdit.js
@@ -178,6 +178,18 @@ class UserEdit extends React.Component {
       });
   }
 
+  formatCustomFieldsPayload(customFields) {
+    const copiedCustomFields = { ...customFields };
+
+    Object.keys(copiedCustomFields).forEach(customFieldId => {
+      if (!copiedCustomFields[customFieldId]) {
+        delete copiedCustomFields[customFieldId];
+      }
+    });
+
+    return copiedCustomFields;
+  }
+
   update({ requestPreferences, ...userFormData }) {
     const {
       updateProxies,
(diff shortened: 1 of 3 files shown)


MaksymDryha:
> Please, add to PR description link of story/bug which it's relates to

This PR is not directly related to any ticket, but its changes are needed to support selection of empty custom field values.
https://github.com/folio-org/stripes-smart-components/pull/837

src/views/UserEdit/UserEdit.js
```diff
@@ -178,6 +178,18 @@ class UserEdit extends React.Component {
...
   }
 
+  formatCustomFieldsPayload(customFields) {
+    const copiedCustomFields = { ...customFields };
+
+    Object.keys(copiedCustomFields).forEach(customFieldId => {
+      if (!copiedCustomFields[customFieldId]) {
+        delete copiedCustomFields[customFieldId];
+      }
+    });
```
Vladyslav-Velytskyi, line 188: I don't really like this solution because you mutate the object, it's not good point to use it with `const`. I would suggest writing more code but make it more correct. For example:

```suggestion
    const filteredCopiedCustomFields = Object.keys(copiedCustomFields).reduce((acc, customFieldId) => {
      if (!copiedCustomFields[customFieldId]) {
        return acc;
      }
      return { ..acc, copiedCustomFields[customFieldId] };
    }, {});
    
    return filteredCopiedCustomFields;
```

MaksymDryha: @Vladyslav-Velytskyi , I made a copy of `customFields` exactly to avoid making changes to the original object, so I don't see any problems with mutating `copiedCustomFields`. 
Also, I tried using `reduce`:
```js
formatCustomFieldsPayload(customFields) {
    return Object.keys(customFields).reduce((acc, cfId) => (
      customFields[cfId]
        ? { ...acc, [cfId]: customFields[cfId] }
        : acc
    ), {});
  }
``` 
To me, it seems less straightforward than the current solution which simply says "check every custom field, remove the ones that are empty"

PR #1373, opened 2020-06-18
UIU-1675: Change validation messages

annamelnyk:
# Description
According to the new requirments validation for `Settings --> Users --> Patron blocks --> Limits` was changed. 
# In scope of ticket
https://issues.folio.org/browse/UIU-1675
# Screenshot
image

### DIFF
--- a/src/settings/patronBlocks/Limits/LimitsForm.js
+++ b/src/settings/patronBlocks/Limits/LimitsForm.js
@@ -7,6 +7,7 @@ import {
   map,
 } from 'lodash';
 
+import SafeHTMLMessage from '@folio/react-intl-safe-html';
 import {
   Row,
   Col,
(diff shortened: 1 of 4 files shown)


src/settings/patronBlocks/Limits/LimitsForm.js
```diff
@@ -18,28 +19,55 @@ import {
...
 import css from '../patronBlocks.css';
 
 function validation(value, type) {
   const numberValue = toNumber(value);
-  const min = 0.01;
-  const max = 9999.99;
+  let min;
+  let max;
+
+  if (type === 'limits') {
```
mkuklis, line 34: @annamelnyk I wonder if `validation` could take `min` and `max` as params instead of `type`. That way you could avoid having these `if` statements. What do you think?

annamelnyk: Totally agree. Thanks @mkuklis for review and fresh look.

src/settings/patronBlocks/Limits/LimitsForm.js
```diff
@@ -18,28 +19,38 @@ import {
...
-  const min = 0.01;
-  const max = 9999.99;
 
   if (numberValue < min || numberValue > max) {
-    return <FormattedMessage id={`ui-users.settings.limits.${type}.error.message`} />;
+    return (
+      <SafeHTMLMessage
+        id="ui-users.settings.limits.validation.message"
+        values={{ min, max }}
+      />
```
zburke, line 37: Is there a reason we need `SafeHTMLMessage` here? The translation value shows no HTML: 
```
  "settings.limits.validation.message": "Must be blank or a number from {min} to {max}",
```
NB: technically, there should be a [`number` type](https://formatjs.io/docs/icu-syntax#number-type) given to the message formatter: 
```
  "settings.limits.validation.message": "Must be blank or a number from {min, number} to {max, number}",

```

PR #1379, opened 2020-06-20
UIU-1214 Claim returned-1: effect on SET COST fees for item status Declared Lost

ymacedor:
https://issues.folio.org/browse/UIU-1214

Patrons may think they have returned items that are still showing as checked out on their account. When this happens, the item is still considered loaned to the patron, but the library will treat the item differently from other items on the patron's account (the item will not be considered overdue, and it will be searched for extensively). A patron should not be charged for an item that they claim they've returned until the library takes further action (described in separate stories).


zburke:
Can you please rebase this to tip of master? There are conflicts with many source files here that need to be resolved before I can sort through the changes.

ymacedor:
We are working on that code, because we realize that it is an incorrect branch, but we will correct it

src/components/Wrappers/withClaimReturned.js
```diff
@@ -15,6 +66,180 @@ const withClaimReturned = WrappedComponent => class withClaimReturnedComponent e
...
+      const lostStatus = 'Lost item fee';
+      const processingStatus = 'Lost item processing fee';
+      const pathParts = [
+        'accounts?query=',
+        `barcode=="${itemBarcode}"`,
+        ` and userId=="${userId}"`,
+        ` and (feeFineType=="${lostStatus}"`,
+        ` or feeFineType=="${processingStatus}")`
+      ];
+      const path = pathParts.reduce((acc, val) => acc + val, '');
```
zburke, line 87: nothing wrong with `reduce`, but a simpler approach is ```const path = pathParts.join('');```

src/components/Wrappers/withClaimReturned.js
```diff
@@ -15,6 +66,180 @@ const withClaimReturned = WrappedComponent => class withClaimReturnedComponent e
...
+    const getAccounts = (itemBarcode, userId) => {
+      const {
+        mutator: {
+          accounts: {
+            GET,
+          },
+        }
+      } = this.props;
+      const lostStatus = 'Lost item fee';
+      const processingStatus = 'Lost item processing fee';
```
zburke, line 79: Instead of hard-coding strings like this, maybe make a list of possible values in `src/constants.js`? Then you can put all the potential fee statuses together, all the processing statuses together, all the payment statuses (see below on line 93) etc.

src/components/Wrappers/withClaimReturned.js
```diff
@@ -15,6 +66,180 @@ const withClaimReturned = WrappedComponent => class withClaimReturnedComponent e
...
+        ? 'Refund'
+        : 'Refunded';
+      return {
+        dateAction: now,
+        typeAction: type,
+        comments: '',
+        notify: false,
+        amountAction: amount,
+        balance: 0,
+        transactionInformation: `${transactionVerb} to ${orderedActions[0].paymentMethod}`,
```
zburke, line 160: I'm concerned `transactionInformation` won't localize correctly because of the way it is structured. Could/should the verb and method be handled separately?

src/components/Wrappers/withClaimReturned.js
```diff
@@ -15,6 +66,180 @@ const withClaimReturned = WrappedComponent => class withClaimReturnedComponent e
...
+        userId: currentUserId,
+        createdAt: servicePointId,
+      };
+    };
+
+    const createRefunds = (account, actions) => {
+      return actions.length > 0
+        ? [
+          createRefundActionTemplate(account, actions, 'Credited fully'),
+          createRefundActionTemplate(account, actions, 'Refunded fully')
```
zburke, line 173: Can refund-actions be listed in `src/constants.js`?

src/components/Wrappers/withClaimReturned.js
ymacedor: These strings were changed to the file constants.js

src/components/Wrappers/withClaimReturned.js
ymacedor: Yes, these strings were moved to constants.js

src/components/Wrappers/withClaimReturned.js
ymacedor: This is how the transaction information is handled:
imagen 
but we can ask Holly Mistlebauer

src/components/Wrappers/withClaimReturned.js
ymacedor: If necessary we can change it

PR #1384, opened 2020-06-23
Pin react-intl to v4.6.10

BogdanDenis:
Pinning `react-intl` to `v4.6.10` to fix CI failures when installing `@formatjs/intl-datetimeformat`

### DIFF
--- a/package.json
+++ b/package.json
@@ -709,7 +709,7 @@
   "peerDependencies": {
     "@folio/stripes": "^4.0.0",
     "react": "*",
-    "react-intl": "^4.5.3",
+    "react-intl": "~4.6.10",
     "react-router": "*",
     "react-router-dom": "*"
   },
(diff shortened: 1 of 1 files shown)


zburke:
`react-intl` `4.7.2` resolves the build problem. Our existing caret dep _should_ find this version automatically, but it's not a guarantee of course. What do you think, @BogdanDenis, should we close this in favor of a PR to set the minimum to `4.7.2`, or keep this one with the more conservative tilde dep since third party deps keep killing us, or just close this and move on?

BogdanDenis:
@zburke I can see that those who locked the older version have already reverted their PRs and new builds seem to be ok, so there's no reason to lock `4.6.10`.
And I think that we should leave it as it is and not lock it. As I see it - we're continuously "updating" our dependencies (by updating I mean we have the latest versions installed on CI and built with). Yes, there are sometimes issues like these, but not updating deps this often could lead to a lot of work when we manually update them and there's an avalanche of errors somewhere in the pipeline

PR #1385, opened 2020-06-23
UIU-1715 retrieve up to 200 patron groups

zburke:
There was no `limit` clause on the patron group query for fee/fine
limits, causing only the default limit (10) to be retrieved. This was
inadequate.

Fixes [UIU-1715](https://issues.folio.org/browse/UIU-1715)


src/settings/LimitsSettings.js
```diff
@@ -36,14 +42,19 @@ class LimitsSettings extends Component {
...
       type: 'okapi',
       records: 'patronBlockLimits',
       GET: {
-        path: 'patron-block-limits?limit=1500',
-      }
+        path: 'patron-block-limits',
+      },
+      params: {
+        query: 'cql.allRecords=1',
+        limit: MAX_RECORDS,
```
annamelnyk, line 56: Oh, looks like I just missed the limit for patron groups, and when I did work with limits the maximum patron groups could be 200,  as I discussed with Holly, and here is 2000, but I don't mind. But then we know that there 6 conditions (are always exist on BE and can not be created from UI), and for now it is 2000 patron groups, then patronBlockLimits should be increased to 2000 * 6 = 12000, right? I guess before I was going to set it to 200*6=1200, and set it to 1500 just in case.

zburke: I pulled 2000 from the [ControlledVocab](https://github.com/folio-org/stripes-smart-components/blob/6be4b917d22149f1945b16cf90bd96deec369727/lib/ControlledVocab/ControlledVocab.js#L29) to kinda sorta aim for consistency. 

If you have real intel about different numbers to use here, by all means let's use them! Feel free to take over this branch if you want, or I can make the changes. Just LMK your preference.

annamelnyk: Sure, I will add changes to this branch.

PR #1391, opened 2020-06-25
UIU-1675: Add validation check for not decimal values

annamelnyk:
# Description
For patron block limits are allowed only integers values, but only for `Maximum outstanding fee/fine balance` condition allowed both decimal and integers.
# Issue
https://issues.folio.org/browse/UIU-1675
# Screenshot
image

### DIFF
--- a/test/bigtest/tests/patron-blocks/settings-limits-form-test.js
+++ b/test/bigtest/tests/patron-blocks/settings-limits-form-test.js
@@ -104,14 +104,15 @@ describe('Patron blocks limits form', () => {
       });
     });
 
-    describe('set float limit value for fee fine limit', () => {
+    describe('save float limit value for fee fine limit', () => {
       beforeEach(async function () {
         await SettingsLimitsForm.limitField(5)
           .fillAndBlur(12.5);
+        await SettingsLimitsForm.saveButton.click();
       });
 
-      it('should show error message', () => {
-        expect(SettingsLimitsForm.errorMessage.isPresent).to.be.false;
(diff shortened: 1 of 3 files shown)


test/bigtest/tests/patron-blocks/settings-limits-form-test.js
```diff
@@ -91,6 +103,17 @@ describe('Patron blocks limits form', () => {
...
+
+    describe('set float limit value for fee fine limit', () => {
+      beforeEach(async function () {
+        await SettingsLimitsForm.limitField(5)
+          .fillAndBlur(12.5);
+      });
+
+      it('should show error message', () => {
+        expect(SettingsLimitsForm.errorMessage.isPresent).to.be.false;
+      });
```
zburke, line 115: "should show error message ... expect errorMessage is present ... to be false"

This doesn't add up to me.

annamelnyk: @zburke, do you mean will be appropriate to rewrite it to "should not show error message"?

zburke: The test title is effectively "error message is present" but the test passes when `errorMessage.isPresent` is `false`. One of those conditions needs to be flipped.

annamelnyk: I rewrote test, looks like make more sence to show that we can save float value for current limit, not just show that there are not error message

PR #1400, opened 2020-07-02
UIU-1730: Add subpermissions

annamelnyk:
# Description
Add required subpermissions for renew through override.
# Issue
https://issues.folio.org/browse/UIU-1730
# Screenshots
image

image


annamelnyk:
Permission issue was resolved on BE side (https://issues.folio.org/browse/CIRC-820).

PR #1404, opened 2020-07-07
UIU-1712: Add servicePointId property when overriding a loan

SergiyVSergiyenko:
https://issues.folio.org/browse/UIU-1712

### Purpose
Add `servicePointId` property to the request when the user overrides a loan.

### DIFF
--- a/src/components/BulkOverrideDialog/BulkOverrideInfo.js
+++ b/src/components/BulkOverrideDialog/BulkOverrideInfo.js
@@ -134,6 +134,13 @@ class BulkOverrideInfo extends React.Component {
       user: {
         barcode: userBarcode
       },
+      stripes: {
+        user: {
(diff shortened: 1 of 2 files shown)


src/components/BulkOverrideDialog/BulkOverrideInfo.js
```diff
@@ -143,13 +143,15 @@ class BulkOverrideInfo extends React.Component {
         item: {
           barcode,
         },
+        checkoutServicePointId,
```
bohdan-suprun, line 146: Is it loan's check-out service point or currently chosen service point?

SergiyVSergiyenko: Good catch. Fixed.

PR #1406, opened 2020-07-07
sentence case

zburke:
Instead of 10,000 little PRs to change the casing wherever we notice it,
here's one big one to fix 'em all.


zburke:
This was a nice idea but it got buffeted by some other PRs that landed between when it was opened and when it was completed. It should just be closed and re-opened from scratch at this point as too much time has elapsed.

PR #1411, opened 2020-07-08
UIU-1755: Change '-' to 'Default' if default notice exists at Fee/Fine: Manual Charges

annamelnyk:
# Description
We need to differ when there was not setted `charge/action notices` and display `-`, from when there was not setted `charge/action notices` but there are exist `charge/action notices` by default and display `Default`. 
 # Task
https://issues.folio.org/browse/UIU-1755
# Screenshots
**Before**
image

**After**
image

### DIFF
--- a/src/settings/FeeFineSettings.js
+++ b/src/settings/FeeFineSettings.js
@@ -93,7 +93,7 @@ class FeeFineSettings extends React.Component {
     this.state = {
       ownerId: '',
       owners: [],
-      templates: []
+      templates: [],
     };
 
     this.connectedControlledVocab = props.stripes.connect(ControlledVocab);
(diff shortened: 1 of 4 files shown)


src/settings/FeeFineSettings.js
```diff
@@ -230,6 +240,24 @@ class FeeFineSettings extends React.Component {
...
+    const defaultNoticeId = defaultNotices[`default${noticeType}NoticeId`];
+    const defaultMessage = <FormattedMessage id="ui-users.settings.default" />;
+    let templateName = templates.find(t => t.id === noticeTypeId) || {};
+
+    if (noticeTypeId) {
+      templateName = templateName?.name;
+    } else if (!noticeTypeId && defaultNoticeId) {
+      templateName = defaultMessage;
+    } else {
+      templateName = '-';
```
zburke, line 255: Please use the [NoValue](https://github.com/folio-org/stripes-components/tree/master/lib/NoValue) component here to show `-` in an a11y-friendly way.

`getNotice()` cleans up the duplicate template-finding logic neatly. Nice work!

PR #1414, opened 2020-07-09
Stripes 5 react router 5.2

zburke:
maaaaaaaybe this resolves UIU-1759? I don't think it should, as the platform should _already_ be supplying all the deps that are changed here, but I can't reproduce UIU-1759 locally where I've had this change in place for a while.


zburke:
OK, this is _nothing_ to do with UIU-1719 (which is really STCON-111). But we still need to migrate to stripe 5 and react-router v5.2.

PR #1419, opened 2020-07-15
Update stripes to v5.0.0

mkuklis:


mkuklis:
Oh @zburke already had that in https://github.com/folio-org/ui-users/pull/1414 so I'm going to close this one

PR #1422, opened 2020-07-21
UIU-1773: Make fee/fine not allowed to be created without required information (itemId)

annamelnyk:
# Description
Buttons `Charge & pay now` and `Charge only` are disabled until will not inut valid item to the `Item information` field. 
# Issue
https://issues.folio.org/browse/UIU-1773
# Screenshots
**Before**
image

**After**
image
image


annamelnyk:
Requirments for https://issues.folio.org/browse/UIU-1773 were changed, and it is not a bug. Later after BE changes, need to be send `empty value` instead of `0` for `itemId`.

PR #1425, opened 2020-07-22
UIU-1211: Settings > Users > Departments CRUD

Vladyslav-Velytskyi:
## Description
Add Settings -> Users -> Departments page

## Issue
[UIU-1211](https://issues.folio.org/browse/UIU-1211)

## Screenshot
image


src/settings/DepartmentsSettings.js
```diff
@@ -0,0 +1,59 @@
...
+      stripes={stripes}
+      baseUrl="departments"
+      records="departments"
+      label={intl.formatMessage({ id: 'ui-users.settings.departments' })}
+      labelSingular={intl.formatMessage({ id: 'ui-users.settings.department' })}
+      objectLabel={<FormattedMessage id="ui-users.settings.departments.users" />}
+      visibleFields={['name', 'code']}
+      columnMapping={{
+        name: intl.formatMessage({ id: 'ui-users.settings.departments.name' }),
+        code: intl.formatMessage({ id: 'ui-users.settings.departments.code' }),
```
annamelnyk, line 42: Nice work!
As recommendation, you can extract `formatMessage` above to make code cleaner:
`const { formatMessage } = useIntl();`

test/bigtest/network/config.js
```diff
@@ -658,4 +658,31 @@ export default function config() {
...
+      code: 'test2',
+      usageNumber: 1,
+      metadata: {
+        createdDate: () => '2019-02-05T18:49:20.839+0000',
+        createdByUserId: () => 'ce0e0d5b-b5f3-4ad5-bccb-49c0784298fd',
+        updatedDate: () => '2019-02-05T18:49:20.839+0000',
+        updatedByUserId: () => 'ce0e0d5b-b5f3-4ad5-bccb-49c0784298fd'
+      },
+    }]
+  });
```
annamelnyk, line 687: I guess next time probably will be more appropriate to create separate scenarious for such responses, if this data can be created via client and not always exists on server side. It would convey the real picture of app, and also `config` file will be cleaner without  mocked data.

src/settings/DepartmentsSettings.js
```diff
@@ -0,0 +1,59 @@
...
+} from 'react-intl';
+
+import { NoValue } from '@folio/stripes/components';
+import { ControlledVocab } from '@folio/stripes/smart-components';
+import { useStripes } from '@folio/stripes/core';
+
+const DepartmentsSettings = () => {
+  const intl = useIntl();
+  const stripes = useStripes();
+  const ConnectedControlledVocab = stripes.connect(ControlledVocab);
```
MaksymDryha, line 14: `ConnectedControlledVocab` will be mounted from scratch every time when `DepartmentsSettings` renders. To avoid this, a would suggest wrapping `stripes.connect(ControlledVocab);` with `useMemo`:
```js
const ConnectedControlledVocab = useMemo(() => stripes.connect(ControlledVocab), [stripes]);
```

test/bigtest/interactors/settings-departments.js
```diff
@@ -0,0 +1,47 @@
...
+  interactor,
+  clickable,
+  count,
+  text,
+  scoped,
+  collection,
+  isPresent,
+} from '@bigtest/interactor';
+
+import TextFieldInteractor from '@folio/stripes-components/lib/TextField/tests/interactor'; // eslint-disable-line
```
BogdanDenis, line 11: Please add name of the rule that you're disabling here

PR #1439, opened 2020-08-04
UIU-1791 handle known/unknown loan action values

zburke:
Add handling for the loan action `itemAgedToLost`, and add handling if
new, as of yet unknown statuses, are added so the UI can display
`unknown action` instead of exploding with a stacktrace when they do.

Refs [UIU-1791](https://issues.folio.org/browse/UIU-1791)


bohdan-suprun:
Hi @zburke, can it be merged? just wanted to unblock issues before my vac?

PR #1441, opened 2020-08-04
UIU 1750: Filter users by tags does not work

K-Felk:
The source of this problem was that the tags filter in the user search
was set to use tag ID when filtering by tag. This does not work because
the 'tags' smart component does not store tag IDs in entity records. It
stores the tag names as a list. So the only way to filter for any entity,
including users, is to filter by tag name. Changing this would require
reworking a fairly substantial portion of the tags smart component, so I
simply changed the search in ui-users back to filtering by tag name.

This PR also includes fixes for broken tests in UI-users.  These changes
are related to UIU-1801, but they don't fix that issue-I just adjusted the 
tests to work around it so people can merge PRs while it gets worked 
on, since the fix for that issue seems likely not to be simple.

refs: https://issues.folio.org/browse/UIU-1750
https://issues.folio.org/browse/UIU-1801

### DIFF
--- a/src/components/Loans/OpenLoans/OpenLoansControl.js
+++ b/src/components/Loans/OpenLoans/OpenLoansControl.js
@@ -21,6 +21,8 @@ import {
 } from '../../Wrappers';
 import TableModel from './components/OpenLoansWithStaticData';
 
+/* eslint react/prop-types: "off" */
+
 class OpenLoansControl extends React.Component {
   static propTypes = {
     stripes: stripesShape.isRequired,
--- a/test/bigtest/tests/declare-lost-test.js
+++ b/test/bigtest/tests/declare-lost-test.js
@@ -231,10 +231,9 @@ describe('Declare Lost', () => {
         await OpenLoansInteractor.declareLostDialog.additionalInfoTextArea.fill('item lost');
         await OpenLoansInteractor.declareLostDialog.confirmButton.click();
       });
-    });
-
-    it('should update fine incurred amount', () => {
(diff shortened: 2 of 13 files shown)


src/components/Loans/OpenLoans/OpenLoansControl.js
```diff
@@ -21,6 +21,8 @@ import {
 } from '../../Wrappers';
 import TableModel from './components/OpenLoansWithStaticData';
 
+/* eslint react/prop-types: "off" */
+
```
zburke, line 25: Why are we turning off prop-types linting here? Do you have reason to believe this is a bogus warning?

test/bigtest/tests/declare-lost-test.js
```diff
@@ -231,10 +231,9 @@ describe('Declare Lost', () => {
         await OpenLoansInteractor.declareLostDialog.additionalInfoTextArea.fill('item lost');
         await OpenLoansInteractor.declareLostDialog.confirmButton.click();
       });
-    });
-
-    it('should update fine incurred amount', () => {
-      expect(LoanActionsHistory.feeFines.text).to.equal('250.00');
+      it('should update fine incurred amount', () => {
+        expect(LoanActionsHistory.feeFines.text).to.equal('250.00');
+      });
```
zburke, line 236: Nice catch!

zburke:
This does indeed resolve [UIU-1750](https://issues.folio.org/browse/UIU-1750) but on further review it also reintroduces [UIU-1596](https://issues.folio.org/browse/UIU-1596).

PR #1443, opened 2020-08-06
UIU-1224: Add departments to User record

Vladyslav-Velytskyi:
## Description
- Add Departments name to User Create/Edit/View Pages
- Add validation to Settings > Departments for empty code value

## Issue
[UIU-1224](https://issues.folio.org/browse/UIU-1224)

## Screenshot
image

### DIFF
--- a/src/routes/UserRecordContainer.js
+++ b/src/routes/UserRecordContainer.js
@@ -10,6 +10,8 @@ import {
   withProxy,
   withServicePoints
 } from '../components/Wrappers';
+import { departmentsShape } from '../shapes';
+import { MAX_RECORDS } from '../constants';
 
 class UserRecordContainer extends React.Component {
   static manifest = Object.freeze({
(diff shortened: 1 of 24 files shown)


src/routes/UserRecordContainer.js
```diff
@@ -56,6 +57,11 @@ class UserRecordContainer extends React.Component {
       path: 'addresstypes?query=cql.allRecords=1 sortby desc',
       records: 'addressTypes',
     },
+    departments: {
+      type: 'okapi',
+      path: 'departments?query=cql.allRecords=1',
+      records: 'departments',
+    },
```
zburke, line 64: FYI, even with `cql.allRecords=1`, this query will only retrieve 10 records unless you add a `limit=N` clause, as for the `patronGroups` resource a few lines above. Consider importing `MAX_RECORDS` from `../constants`. 

Additionally, consider a `sortby` clause.

PR #1447, opened 2020-08-07
UIU-1801: crashing when accessing the "edit user details" screen 

K-Felk:
With some help, worked out that this was caused by the form rendering 
before the data fetch operation was complete, the form would then
crash because it had no user data to display.  Added a check in render
that looks to see if the edit is on an existing user, and wether the 
data fetch is done.  Also rewrote all test changes so that tests 
can now visit the user edit screen directly.

refs: https://issues.folio.org/browse/UIU-1801


src/views/UserEdit/UserEdit.js
```diff
@@ -292,6 +292,11 @@ class UserEdit extends React.Component {
     }
 
     // data is information that the form needs, mostly to populate options lists
+
+    // check to see if we are editing an existing user, and if the data on that user has finished loading.
+    if (!this.getUser() && this.props.match.params.id) {
+      return null;
+    }
```
zburke, line 299: Would it make more sense, or perhaps be a better UX, to incorporate this check into the `resourcesLoaded` check above and return a `<LoadingView>` instead of `null`?

K-Felk: I tried it out-it makes that "if" a bit long, but it does present a loading indicator while it's loading the data.  Pushed the change up just now.

zburke: `params.id` is already destructured, which would help with the line-length a bit. 

You could leave it as-is, or put the separate condition-groups on separate lines if it bugs you.

PR #1467, opened 2020-09-02
UIU-1803: Include Aged to lost in loan details action history

Baroquem:
https://issues.folio.org/browse/UIU-1803

This is to ensure that Aged to lost actions are properly represented in the loan details action history table. Most of the work for this is done automatically on the backend. The one UI change needed is to insert "System" into the source column whenever it's blank in the original.

Screen Shot 2020-09-02 at 1 04 05 PM


SergiyVSergiyenko (approved):

mkuklis (approved):
@Baroquem I wonder if `System` should be moved to translations. What do you think?

Baroquem:
> @Baroquem I wonder if `System` should be moved to translations. What do you think?

I think you're probably right! Thanks for catching that.

PR #1469, opened 2020-09-03
UIU-1804: Update Lost date in loan details

K-Felk:
I added some code to the loan details view that checks
to see if the item has been declared lost/aged to lost/
lost and paid, and calculates the lost date field accordingly.

refs: https://issues.folio.org/browse/UIU-1804


src/views/LoanDetails/LoanDetails.js
```diff
@@ -351,11 +350,17 @@ class LoanDetails extends React.Component {
     const isClaimedReturnedItem = itemStatus === itemStatuses.CLAIMED_RETURNED;
     const isDeclaredLostItem = itemStatus === itemStatuses.DECLARED_LOST;
     const isAgedToLostItem = itemStatus === itemStatuses.AGED_TO_LOST;
+    const isLostAndPaid = itemStatus === itemStatuses.LOST_AND_PAID;
     let lostDate;
     const declaredLostActions = loanActionsWithUser.filter(currentAction => get(currentAction, ['action'], '') === 'declaredLost');
-
+    const agedTolostLostActions = loanActionsWithUser.filter(currentAction => get(currentAction, ['action'], '') === 'itemAgedToLost');
```
mkuklis, line 356: @K-Felk could we introduce more constants for loan actions here: https://github.com/folio-org/ui-users/blob/master/src/constants.js#L22
and reference them in `LoadDetails` instead of using a string? 

For example: `loanActions.DECLARED_LOST` or `loanActions.ITEM_AGED_TO_LOST`

This will help if we ever need to change them in the future.

mkuklis (approved):

src/views/LoanDetails/LoanDetails.js
K-Felk: Done!

PR #1476, opened 2020-09-09
UIU-1369: Assign permissions more efficiently

Baroquem:
https://issues.folio.org/browse/UIU-1369

This PR replaces the one-by-one permissions assignments with an all-in-one assignment, thanks to the API that Julian pointed out. For the case of assigning all available permissions to a user, that means a reduction from 200+ API requests down to one, or at most two. Hooray!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@
 * Add missing permission `departments.collection.get` permission. Fixes UIU-1812.
 * Prevent UI crashing when loading loan with deleted item. Fixes UIU-1819.
 * Remove user count from patron groups. Fixes UIU-1562.
+* Assign user permissions more efficiently. Fixes UIU-1369.
 
 ## [4.0.0](https://github.com/folio-org/ui-users/tree/v4.0.0) (2020-06-17)
 [Full Changelog](https://github.com/folio-org/ui-users/compare/v3.0.0...v4.0.0)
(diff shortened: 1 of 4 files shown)


zburke (approved):
Nicely done! Thanks for the comments; they are helpful in general, and specifically in terms of helping to understand the not-really-obvious way that user permissions are managed. Thank you!

zburke:
Aside, do we want to copy `eachPromise` to `stripes-connect` and export it there? It's a really useful piece of code, and I wonder if it could be the core of a generic client-side message-queue implementation that UI apps could use to hand of a batch of requests. Could we do something evil/clever with `eachPromise`, `dataKey`, and a callout-callback (e.g. [ClipCopy](https://github.com/folio-org/stripes-smart-components/blob/d0a0bef3dcd3115934ef25405d15c63dc7d46266/lib/ClipCopy/ClipCopy.js#L20-L29)) to create uniquely-named resources that would allow these requests to be processed asynchronously, in series, without worry that they could be stomped on by loading another page that accesses the same resource?

Thoughts @mkuklis, @skomorokh?

mkuklis (approved):
Nice work @Baroquem!

I agree with @zburke that it would be nice to be able to reuse it and I also agree with @skomorokh about `stripes-util`

mkuklis:
I'm also glad that new api works!

Baroquem:
Well, I can move that function to stripes-util as-is, if that's the consensus. Thanks for the feedback!

PR #1483, opened 2020-09-14
UIU-1849: Only set servicePointUserId when it's present

mkuklis:
This took a while to find.

the `nextProps.mutator.servicePointUserId.replace` was called multiple times when navigating away and back to users module causing:

````
 Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops.
````

https://issues.folio.org/browse/UIU-1849


Baroquem (approved):

zburke (approved):
👏 👏 👏 

Nice work!

PR #1484, opened 2020-09-14
cleanup invalid popover code

zburke:
There were a collection of invalid things going on here:

* invalid `` link to a third party server
* hard-coded/invalid `<Link>` pointing to `/users/123`
* the way the popover here was configured, I think its click-handler
fought with the row's click-handler and lost, so the `<Popover>` was
never triggered.

Given all this stuff that didn't really add up, the column handler was
simplified to just display the fine-type, which is the one thing that
_did_ work.


K-Felk (approved):

mkuklis (approved):
Thank you for cleaning this up @zburke!

annamelnyk (approved):

Baroquem (approved):

PR #1487, opened 2020-09-15
UIU-1836: Move Charge and pay form to react-final-form

annamelnyk:
# Description
Refactor charge and pay action and move form to react-final-form
# Task
https://issues.folio.org/browse/UIU-1836


src/components/Accounts/ChargeFeeFine/ChargeForm.js
```diff
@@ -199,16 +208,24 @@ class ChargeForm extends React.Component {
         </Button>
         <Button
           id="chargeAndPay"
-          disabled={this.props.pristine || this.props.submitting || this.props.invalid}
-          onClick={this.props.handleSubmit(data => this.props.onSubmit({ ...data, pay: true, notify }))}
+          type="submit"
```
maximdidenkoepam, line 211: Why do we need submit here if we have onClick handler? Does this handler will be called when user click enter button?

maximdidenkoepam (approved):

src/components/Accounts/ChargeFeeFine/ChargeForm.js
annamelnyk: Looks like `type=submit` does not need here.
Yes handler will be called if click enter, if focus button with tab first.

mkuklis (approved):

zburke (approved):

PR #1489, opened 2020-09-16
Additional adjustments accompanying #1477

JohnC-80:
In #1477, the columnWidth API was applied to the Loans listing....

This PR consists of more aesthetic adjustments passed along from the PO level - 
- increasing the width of the "Title" column
- Adjusting the 'dueDate' formatter so that the time wraps to the next line.
- hiding the same columns in the Closed loans list... 

One item for future development is to have these column UI's behave the same way with both views using show/hide column functionality...

### DIFF
--- a/src/components/Accounts/ChargeFeeFine/UserInfo.js
+++ b/src/components/Accounts/ChargeFeeFine/UserInfo.js
@@ -20,7 +20,7 @@ const UserInfo = (props) => {
         </Col>
         <Col className={css.userInfoLink}>
           <FormattedMessage id="ui-users.charge.barcode" />
-          {': '}
+          :
           <Link to={`/users/view/${user.id}`}>
             {user.barcode}
           </Link>
--- a/src/components/Loans/ClosedLoans/ClosedLoans.js
+++ b/src/components/Loans/ClosedLoans/ClosedLoans.js
@@ -3,6 +3,7 @@ import React from 'react';
 import {
   FormattedMessage,
   FormattedTime,
+  FormattedDate,
   injectIntl,
 } from 'react-intl';
 import PropTypes from 'prop-types';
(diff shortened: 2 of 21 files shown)


src/components/Accounts/ChargeFeeFine/UserInfo.js
```diff
@@ -20,7 +20,7 @@ const UserInfo = (props) => {
         </Col>
         <Col className={css.userInfoLink}>
           <FormattedMessage id="ui-users.charge.barcode" />
-          {': '}
+          :
```
zburke, line 23: Heh, freaking ESLint, man. Are you confident this will show up as `Barcode: 12345` with a space after the colon? Previously, this looked like
```
<FormattedMessage id="ui-users.charge.barcode" />
<Link to={`/users/view/${user.id}`}>
  {user.barcode}
</Link>
```
and the translation string was variously coded with and without the trailing space, in fact with and without the colon at all 🤦 , and I just couldn't handle seeing `Barcode:1234` any longer. The horror. The HORROR!

src/components/Loans/ClosedLoans/ClosedLoans.js
```diff
@@ -183,14 +184,12 @@ class ClosedLoans extends React.Component {
...
-          value={loan.dueDate}
-          day="numeric"
-          month="numeric"
-          year="numeric"
-        />;
-      },
+      'dueDate': loan => (
+        <div>
+          <div><FormattedDate tagName="div" value={loan.dueDate} /></div>
+          <div><FormattedTime tagName="div" value={loan.dueDate} /></div>
```
zburke, line 190: Do we need the outer `<div>` given we are passing `tagName="div"`?

translations/ui-users/ca.json
```diff
@@ -609,7 +609,7 @@
     "reports.overdue.item.callNumberComponents.suffix": "Call number suffix",
     "reports.overdue.item.enumeration": "Enumeration",
     "reports.overdue.item.volume": "Volume",
-    "loans.details.effectiveCallNumber": "Effective call number string",
+    "loans.details.effectiveCallNumber": "Effective call number",
```
zburke, line 612: Please update only `en.json` and let the change there feed the other locale files. Any changes made here will just be overwritten the next time the import-from-lokalise script runs. 

(That said, I am guilty of doing this too when a PO reports an error in a translation string because it's the only way I can cause the fix to be visible to a PO, who is the only one with the power to close a ticket. @dltj, is there/could there be a way for devs to have some control of this process, or do you think the Right Fix is simply to educate all parties about how long these changes take to filter through the system -- is there documentation about how it works (is it periodic, automatic based on the presence of changes, manual, etc)?)

dltj: I'd like to totally automate the handling of translations ([FOLIO-1820|https://issues.folio.org/browse/FOLIO-1820]) to be a nighly thing, but the lack of stability in the test means that I have to go in and restart the Jenkins jobs for at least a couple repositories every time I run the lokalise batch job.  I'm hoping the new testing framework brings that stability!

dltj: I should add that it is okay to update translation files other than `en.json`—just be aware that they will be overwritten the next time that the translations from Lokalise are imported.

zburke:
Bump. Where does this stand, @JohnC-80?

zburke:
Bump. Where does this stand, @JohnC-80? It's been open since September.

zburke:
This has likely drifted so far from master that it would probably be less work to just start over than to deal with the merge conflicts.

PR #1498, opened 2020-09-30
UIU-1226 Claimed returned in-app report: Existing fields

K-Felk:
The existing loans report had a lot of code that I would have duplicated,
so rather than write new code that runs a separate claimed returned report,
I altered the loan report code to make it more generic.  When you call the
"generate" method, it now takes a string that tells it what kind of report
you want to run.  The report generator sets the columns and query string
appropriately, then retrieves the data.  I moved the column names into a
separate file and separated them out-right now, the columns for the two
reports are identical, but I can see from the story that later we'll want
to add specific columns to the claim returned report-that should now be easy.
I shuffled around a lot of labels that both reports use so their names aren't
as tied to the overdue report.

refs: https://issues.folio.org/browse/UIU-1226


K-Felk:
I'm not sure how much I should care about the code quality failure here-first off, my understanding is that we are still unclear on wether we are moving to a new testing framework, and second...how exactly would I test this?  Can the testing framework actually open and parse a CSV file?

translations/ui-users/en.json
```diff
@@ -697,33 +697,34 @@
   "transfers.singular": "Transfer account",
   "transfers.nodata": "There are no Fee/fine: Transfer accounts",
 
+  "reports.claimsReturned.label": "Claims returned report (CSV)",
```
SergiyVSergiyenko, line 700: Please pay attention that the menu item should be named "Claim**ed** returned report (CSV)".

src/components/data/reports/CsvReport.js
```diff
@@ -1,49 +1,39 @@
...
   constructor(options) {
     const { formatMessage } = options;
+    this.formatMessage = formatMessage;
+  }
+
+  setUp(type) {
+    const overDueDate = moment().tz('UTC').format();
+    let columns;
+    this.queryString = '';
+    if (type === 'overdue') {
```
mkuklis, line 18: @K-Felk could `type` be used as a `key` in `reportColumns`? That way you could skip the if/else statements.

For example:

````js
const columns = reportColumns[type];
````

You could also extract `queryString` in the same fashion maybe as a function since you have to use `overDueDate` as a dynamic param but something similar to this should work:

````js
this.queryString = queryStrings[type]();
```

mkuklis (approved):
Nice refactoring @K-Felk!

translations/ui-users/en.json
K-Felk: corrected.

src/components/data/reports/CsvReport.js
K-Felk: that is simpler.  I'll refactor.

src/components/data/reports/reportSettings.js
```diff
@@ -0,0 +1,71 @@
...
+      'item.contributors',
+      'item.location.name',
+      'item.instanceId',
+      'item.holdingsRecordId',
+      'itemId',
+    ],
+  }
+};
+
+export default settings;
```
SergiyVSergiyenko, line 71: Sorry for the late comment. I noticed that the content of the `columns` arrays in both `overdue` and `claimedReturned` objects are the same. If so, wouldn't it be better to move their content into a variable somewhere in constants to avoid duplication?

And also, seems that it is possible not to use quotes in object keys.

K-Felk: I address the duplication in the original PR comment.  I can remove the quotes.

SergiyVSergiyenko: I'm just trying to be guided in these cases by the principle `YAGNI`. ))

SergiyVSergiyenko (approved):

PR #1503, opened 2020-10-05
UIU-1859: Search permission name translations

Baroquem:
https://issues.folio.org/browse/UIU-1859

This PR adds searching of translations of permission displayNames (where such translations exist). The search depends on proper labeling of translation keys for permissions (i.e., including '.permission.' as part of the key, with the remainder corresponding to the official permission name); otherwise, identifying matched translations with the permissions they represent isn't possible (assuming the primary use case, where the permission's `displayName` is missing.


src/components/PermissionsAccordion/components/PermissionsModal/PermissionsModal.js
```diff
@@ -128,18 +134,52 @@ class PermissionsModal extends React.Component {
...
+    });
+
+    // Matched permissions may not have a displayName value. We have to compare a permission's
+    // permissionName value to a matched translation string key, which is not quite the same:
+    // permissionName: ui-users.settings.departments.all
+    // translation key: ui-users.permission.settings.departments.all
+    // This transforms the matched permission translations into an array of permissionNames.
+    // BUT -- this assumes that all permission translation keys are identifiable by the inclusion of
+    // the string '.permission.', which seems rather optimistic.
+    const matchedPermissionKeys = Object.keys(matchedPermTranslations).map(k => k.replace('permission.', ''));
```
mkuklis, line 177: @Baroquem I'm not sure if I'm following this correctly so I apologize if this doesn't make any sense but I was wondering if we could replace this with: https://lodash.com/docs/4.17.15#mapKeys

````js
const matchedPermissionKeys = mapKeys(matchedPermTranslations, (v, k) => k.replace('permission.', ''));
````

With that I think we could avoid `n^2` (with `filter` and `include`) below and just do:

````js
permissions.filter(p => matchedPermissionKeys[p.permissionName]);
````

Would that work? Sorry if this doesn't make any sense.

mkuklis (approved):
@Baroquem this is great work. Thank you for the comments too. It helps with following what is going on. I left one small comment.

K-Felk (approved):

src/components/PermissionsAccordion/components/PermissionsModal/PermissionsModal.js
Baroquem: Thanks, @mkuklis! A wise suggestion for improvement, as usual. 🥇

PR #1508, opened 2020-10-08
UIU-1560: Make assigned service points checked in Add service points …

SergiyVSergiyenko:
https://issues.folio.org/browse/UIU-1560

### Purpose
Make the assigned service points for the user have been checked in the Add service points modal similar to how it works when assigning permissions.

### Approach
1. An array of assigned service points was passed to the modal and used as initial values.
2. When adding new service points, the concatenation of the initial and added service points was used to update the state.
    When the service points were deleted, the updated array of assigned service points was used.
3. Created some tests for the `Add service points` modal.
4. For testing, it was set `interfaces = {}` as an argument to the `setupApplication` function to be able to add interfaces to existing ones.

A bug about doubling the quantity of the service points was also identified and filed [UIU-1892](https://issues.folio.org/browse/UIU-1892).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,7 @@
 * Search translations of permission names. Refs UIU-1859.
 * Hide the Overdue loans report option when user doesn't have view loans permissions. Refs UIU-1858.
 * Allow renewal override if due date isn't changed. Refs UIU-1853.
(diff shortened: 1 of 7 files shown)


Baroquem (approved):

mkuklis (approved):
Nice work @SergiyVSergiyenko!

PR #1529, opened 2020-10-21
UIU-775 add create buttons to actions menu

elsenhans:
https://issues.folio.org/browse/UIU-775

Add create request, create fee/fine and create block options to the actions menu

During the resources of the users details are loading, the focus of both actions menus (in the list of results and in the detail view) is jumping. Also seen this problem in other apps like the orders app. 
Are you aware of this issue already?


src/views/UserDetail/UserDetail.js
```diff
@@ -314,6 +316,69 @@ class UserDetail extends React.Component {
...
+    || this.props.stripes.hasPerm('ui-requests.all');
+
+    const linkToCreateRequest = barcode ?
+      `/requests/?${queryString.stringify({ layer: 'create', userBarcode: barcode })}` :
+      `/requests/?${queryString.stringify({ layer: 'create' })}`;
+
+    if (showActionMenu) {
+      return (
+        <>
+          <IfPermission perm="ui-requests.all">
```
rchr, line 332: This is duplicate to code in _UserForm.js_, as reported by sonarcloud.
Can you please reduce the code duplicity so that your code will pass the quality gate?

rchr (changes requested):
Code looks good.

Even if the ticket specifies the action menu should be displayed in edit mode, I think we don't need the action menu in the create user form. There is no user yet, and hence these actions will throw errors in this case.

And can you please increase the test coverage and decrease code duplicity so that your code passes the quality gate?

Regarding the jumping of focus in the action menu: I noticed this behavior in other apps as well (e.g. inventory, finance). I am not sure how to deal with this right now.
@zburke whats your opinion about this?

zburke:
> Regarding the jumping of focus in the action menu: I noticed this behavior in other apps as well (e.g. inventory, finance). I am not sure how to deal with this right now.

@rchr, I wasn't aware of that behavior but let's not let it sidetrack this PR. Please file an issue for it in ui-users and assign it to Core Functional. After we triage it we can move it to another project if the problems looks to be in a shared component used by other apps like inventory and finance, or clone it into those other apps if the component is innocent but our pattern of using it needs to change. If you are not comfortable with this, LMK and I can file it.

src/views/UserDetail/UserDetail.js
elsenhans: Well, I de-duplicate the code, do not show the action menu in the create user form anymore and finally I get the necessary code coverage.
Thanks for your feedback @rchr!

rchr (changes requested):
For me, the code looks good!

But I have two remarks:
* I think the name of `ActionMenuCreateButtons.js` could be more meaningful. What do you think of `RequestFeeFineBlockActionDialog.js` or `RequestFeeFineBlockActionMenu.js`?

* In addition, `ActionMenuCreateButtons.js` is a component and no utility function. Hence, it is better stored in a different folder. Can you please move it to its own subfolder under `/components/`?

rchr (changes requested):
This looks good.

However, I have one comment to keep the convention of the project. Here, every folder has its own `index.js`.
A consequence of having an `index.js` is that you can shorten your imports, as described in the following example:

If you add an `index.js` to your folder 'ActionDialog' (similar to other folders in this project) you can rename the folder 'ActionDialog' to e.g. 'RequestFeeFineBlockActionDialog'. Then you can shorten your imports to something like:
`import RequestFeeFineBlockActionDialog from '../../components/RequestFeeFineBlockActionDialog';`

rchr (approved):

mkuklis:
@zburke, @rchr and @elsenhans this change completely removed the ability to tag users. It happened after this commit: 

https://github.com/folio-org/ui-users/commit/73aa2f732a78784adfa446f0b5325b5d4a76b6c2#diff-2836ed3693bb2870a04526fa54d42daaf9c3a4c757b852f4519ce38789fa9ddfR478

@elsenhans could you please take a look and see if you can restore it?

You should be able to tag users after the tags are enable from:

https://folio-testing.dev.folio.org/settings/tags/general

It was previously happening in `renderDetailMenu` here:

https://github.com/folio-org/ui-users/blob/73aa2f732a78784adfa446f0b5325b5d4a76b6c2/src/views/UserDetail/UserDetail.js#L299-L312

rchr:
@mkuklis thanks or reporting this!

We will take a look at this and fix it.
I have also created a bug ticket for this: https://issues.folio.org/browse/UIU-1949

mkuklis:
Thank you @rchr!

PR #1533, opened 2020-10-26
UIU-1918: Testing blocker - add user's fees/fines permissions

annamelnyk:
# Description
Add user's fees/fines permissions, without it testing of `fees/fines` are blocked.
# Task
https://issues.folio.org/browse/UIU-1918


maximdidenkoepam (approved):

mkuklis (approved):

zburke (approved):
LGTM. Do these permissions need to be added under `ui-users.accounts` (L#347) as well, or are these effectively "action-permissions" that may be applied to an accounts item?

annamelnyk:
@zburke those permission only related to `actions` and allow to do them

PR #1535, opened 2020-10-26
UIU-1708: Search index barcode

rchr:
This PR adds support for searching on the `barcode` field of a user explicitly.
It adds `barcode` to the rawSearchableIndexes of UserSearch.js.

Refs https://issues.folio.org/browse/UIU-1708


zburke (approved):
Nice work!

PR #1536, opened 2020-10-26
UIU-1890: Show number of open requests in LoanActionDialog

SergiyVSergiyenko:
https://issues.folio.org/browse/UIU-1890

### Purpose
To show the number of requests when the item status changed, we had to pass the value of the `itemRequestCount` property through several components to `LoanActionDialog`.

**Screenshot**
Screen Shot 2020-10-26 at 22 38 54


mkuklis (approved):
@SergiyVSergiyenko this looks good! I wish there was a better way to get access to the `itemRequestCount` so we don't have to pass it down but given current setup I understand your approach. 👏 .

Baroquem (approved):

zburke (approved):
I had exactly the same thought as @mkuklis. Is this the kind of thing we could just stuff into context? (Or is that the kind of ephemeral thing we want to keep out of context to prevent it from getting bloated?)

PR #1539, opened 2020-10-27
UIU-1897: Work with multiple and single refund

annamelnyk:
# Description
Refund count should not include `waived` amount, because no money were actually paid and can not be refunded
# Approach
Previously we receive information from `account.paymentStatus.name` and it is equivalent to `Latest payment status`, so if `pay/transfer` actions were applied and then `waive` - we could not `refund` money, because `account.paymentStatus.name: Waive`.
This means that we have to work with `feefineactions` data instead of `account`.  
# Issue
https://issues.folio.org/browse/UIU-1897
# Screenshots
image


maximdidenkoepam (approved):

zburke (approved):
Wow, this work is complicated! I am so glad you have a firm handle on it, but trying to follow things really drives home the need (for me at least) to have documentation that describes how all these pieces work together and what each endpoint represents (i.e. what is an `account`, what is an `action`, etc.) 

Do you know if such documentation exists, @annamelnyk?

mkuklis (approved):

annamelnyk:
@zburke, as I know, we do not have any detailed documentation regarding `fees/fines`, only [short one](https://wiki.folio.org/pages/viewpage.action?pageId=36576978) about new endpoints

PR #1549, opened 2020-11-03
UIU-1933: Add back confirmation modal

annamelnyk:
# Description
Not asked to confirm leaving New Fee/Fine page anymore
# Issue
https://issues.folio.org/browse/UIU-1933
# Screenshot
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 * Remove optional `action` fields if no value was setted. Refs UIU-1773.
 * Show the number of open requests in the Claim returned bulk action modal. Refs UIU-1891.
 * Add column 'Expiration date offset (days)' to patron group table. Refs UIU-1908.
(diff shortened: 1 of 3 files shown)


maximdidenkoepam (approved):

mkuklis (approved):

zburke (approved):
Nice work! I'm impressed you figured this out. That said, needing to do 
```
this.state.pay && <SomeModal open>
``` 
instead of 
```
<SomeModal open={this.state.pay}>
```
feels a bit wrong to me (not wrong enough to block the PR, but contrary to how I've learned to think in React). And as you noted on Slack, this is necessary to avoid console warnings from react-router like

> A history supports only one prompt at a time at StripesFinalFormWrapper

@JohnC-80, is this something we should investigate either in `stripes-components/lib/Modal` or `stripes-final-form`?

PR #1554, opened 2020-11-04
UITAG-34: Fix filtering by tags

mkuklis:
https://issues.folio.org/browse/UITAG-34

It looks like due to the change in stripes-components:

folio-org/stripes-components#1368

The searching by tags stopped working because tags.tagList only works with `=` operator.


zburke (approved):
LGTM. Thank you for linking to the stripes-components PR that is the root cause of this change; it is really helpful for understanding the context and how this regression came to be. 

Also, sorry for causing this regression 😬 .

PR #1568, opened 2020-11-12
UIU-1909: Create patron block templates in settings

rchr:
## Ticket
https://issues.folio.org/browse/UIU-1909

## What?
Added support to manage manual patron block templates in settings. 

## Why?
When creating a manual patron block the staff member should be able to select from a list of templates. For instance, this is needed if the same block is created over and over. This PR creates the basis for this by creating the possibility of creating the templates. See [UXPROD-2269](https://issues.folio.org/browse/UXPROD-2269) for more info.

## How?
This PR adds a new section block templates to the app's settings. Implementation uses the `<EntryManager>` component from stripes-smart-components.
As patron block templates are managed by _mod-feesfines_ and the interface `feesfines` is listed as an `optionalOkapiInterface` of this app, I added the optional property `interface` to the section objects. Each section is displayed only if the interface exists. For that, I added an additional optional check for the interface in `settings/index.js`.


elsenhans (approved):
Looks good!

zburke (approved):
Nice work! Please choose "Squash and merge" when you merge this. With so many commits on a single PR, it is especially helpful to consolidate them all in a single commit. 

Thank you!

PR #1569, opened 2020-11-12
UIU-1364 Patron block modal erroneously including non-renewal block in list of renewal blocks

ccruz17:
### Ticket
https://issues.folio.org/browse/UIU-1364

### What happened?

A patron has three manual patron blocks, as shown. Only two of the manual patron blocks actually block renewing.
Renewing-Block-Bug1

When we attempt to renew an item, all three blocks show up in the block modal as reasons why the patron is blocked (as shown). Only the two that block renewing should be appearing.

Renewing-Block-Bug2

### Changes
Add renewals filter: Fix Patron block modal that erroneously including non-renewal block in list of renewal blocks


zburke (approved):

ccruz17:
Hi Zack, I made some commits to add the big test for this issue. I request your review. Thank you.

zburke (approved):
Nice work! Thanks for adding the test and cleaning up those lint complaints!

PR #1570, opened 2020-11-16
UIU-1951: Validate value for exiration date offset (settings)

rchr:
## Description
This PR adds validation for the field `expirationOffsetInDays` in patron group settings. 
`expirationOffsetInDays` must be a number > 0 or empty. Otherwise, an error is displayed and the save button is disabled.

## Ticket
https://issues.folio.org/browse/UIU-1951

## Screenshot
validate

### DIFF
--- a/translations/ui-users/en.json
+++ b/translations/ui-users/en.json
@@ -265,7 +265,7 @@
   "information.selectPatronGroup": "Select patron group",
   "information.patronGroup.users": "Users",
   "information.patronGroup.expirationOffset": "Expiration date offset (days)",
-  "information.patronGroup.expirationOffset.error": "Must be a number > 0",
+  "information.patronGroup.expirationOffset.error": "Must be a integer > 0",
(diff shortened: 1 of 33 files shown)


elsenhans (approved):
Looks good!

translations/ui-users/en.json
```diff
@@ -265,6 +265,7 @@
   "information.selectPatronGroup": "Select patron group",
   "information.patronGroup.users": "Users",
   "information.patronGroup.expirationOffset": "Expiration date offset (days)",
+  "information.patronGroup.expirationOffset.error": "Must be a integer > 0",
```
zburke, line 268: Let's include an "allow" empty phrase here too, e.g.  

> Must be empty or an integer > 0

zburke (approved):

translations/ui-users/en.json
rchr: That makes sense! Changes that.

PR #1574, opened 2020-11-23
UIU-1975: Fix empty payments select

annamelnyk:
# Description
'Payment methods' not always appearing for 'Charge & pay now' option of New fee/fine for predefined Owner
# Issue
https://issues.folio.org/browse/UIU-1975
# Screenshot
image

### DIFF
--- a/src/components/Accounts/ChargeFeeFine/ChargeFeeFine.js
+++ b/src/components/Accounts/ChargeFeeFine/ChargeFeeFine.js
@@ -438,7 +438,6 @@ class ChargeFeeFine extends React.Component {
       }
     });
     const feefines = _.get(resources, ['allfeefines', 'records'], []);
-    const payments = _.get(resources, ['payments', 'records'], []).filter(p => p.ownerId === this.state.ownerId);
     const accounts = _.get(resources, ['accounts', 'records'], []);
     const settings = _.get(resources, ['commentRequired', 'records', 0], {});
     const barcode = _.get(resources, 'activeRecord.barcode');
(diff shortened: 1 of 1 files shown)


src/components/Accounts/ChargeFeeFine/ChargeFeeFine.js
```diff
@@ -475,6 +474,10 @@ class ChargeFeeFine extends React.Component {
     const selectedFeeFine = feefines.find(f => f.id === feeFineTypeId);
     const currentOwnerFeeFineTypes = feefines.filter(f => f.ownerId === resources.activeRecord.ownerId);
     const selectedOwner = owners.find(o => o.id === initialOwnerId);
+    const payments = _.get(resources, ['payments', 'records'], []).filter(
+      p => p.ownerId === this.state.ownerId || resources.activeRecord.ownerId
+    );
+
```
zburke, line 480: Am I misreading or will this include all payments when `resources.activeRecord.ownerId` evaluates as truthy? Is that the intent?

annamelnyk: Yes, that is right. 
I see this piece is confusing, I guess would be better just to extract all payments here. Since `payments` are passing to `ActionModal`, where it will filter again (because in multiple payment we have opportunity to change owner there as well).
Will do that

maximdidenkoepam (approved):

src/components/Accounts/ChargeFeeFine/ChargeFeeFine.js
zburke: Ah, OK. I'm fine with either implementation. I see you pushed up a new version already, but if you think this code is better, I'm fine with it. Just include a comment like the one above about how `ActionModal` will filter this. 

Thanks for the explanation!

zburke (approved):

src/components/Accounts/ChargeFeeFine/ChargeFeeFine.js
annamelnyk: I think better is not to filter `payments` here indeed for several reasons: the clearer code, and  avoid unnecessary process

PR #1576, opened 2020-11-24
UIU-1973 New/Edit Record | Change focus to first field

axelhunn:
Refs [UIU-1973](https://issues.folio.org/browse/UIU-1973)

In the User Edit form, when the form opens, there were no focused field by default.
Now focus goes to Last Name field.

I refactored the form code, moving the section components into the same directory as the form itself.
This components are not used anywhere else but this form, and so they are this form's concern.

Also I introduced `Form` and `Field` components:
- `Form` takes the `autoFocusField` prop with the name of the field we want to auto-focus. It's then propagated using context. This way the field to be auto-focused can be defined on top-most level of complex multi-component form (like this one is) 
- `Field` wraps the redux-form field component, it consumes that context.
- If the field name matches, the field is auto-focused by adding internal component prop `autoFocus: true`
- matching can be overriden if we explicitly put the internal component prop `autoFocus` in the field


aliaksei-chumakou:
I'm not sure that introducing another wrapper (Yours) around wrapper (stripesForm) around wrapper (reduxForm) is the best approach. Maybe it's enough to use autoFocus prop at the required field?

axelhunn:
> I'm not sure that introducing another wrapper (Yours) around wrapper (stripesForm) around wrapper (reduxForm) is the best approach. Maybe it's enough to use autoFocus prop at the required field?

I do use `autoFocus` prop. But if I use it directly on the field, it is drown under the layers of sub-components. I believe it's much more convenient to define it on form root level.

zburke:
I share your concern, @aliaksei-chumakou, though I understand how you arrived at that approach @axelhunn. I'm not sure what the right outcome is here. TBH, if we are going to bother refactoring at all, I would love to spend that effort swapping out redux-form in favor of final-form. 

Are you confident, @axelhunn, that there is no simpler approach to setting focus on a specific field in the current codebase?

axelhunn:
> Are you confident, @axelhunn, that there is no simpler approach to setting focus on a specific field in the current codebase?

The simplest way was to make something like
```js
<Field
  //...
  name="personal.lastName"
  component={TextField}
  props={{ autoFocus: true }}
/>
```

But that information, what field is auto-focused, would be buried in sub-component of the form. So it would be hard to find, easy to miss.

I've been told that "this implementation will serve as a pattern for other FOLIO apps", so decided to make the code a little better, while trying to stay in the scope of the ticket.

axelhunn:
I'm aware about moving towards final-form.
I'm also not very happy how this form was decomposed at all, and my refac doesn't solve it fully.
But, again, tried to stay at scope.

rasmuswoelk:
I agree with @aliaksei-chumakou and @zburke on this one.

I am not sure that it's necessary to have this extra layer of abstraction to achieve this. I acknowledge your concern, but in most cases, it should be the first field of the form that will be autofocused, so I don't think it will be too difficult to find the field. A scoped search or simply rendering the page should reveal where the `autoFocus`-prop is set.

axelhunn:
OK, here is alternative PR https://github.com/folio-org/ui-users/pull/1582

PR #1577, opened 2020-11-24
UIU-1910: Use patron block templates to populate fields in create block screen

rchr:
## Ticket
https://issues.folio.org/browse/UIU-1910

## What?
Added support to select patron block templates in order to populate fields in 'create block' screen.

## Why?
When creating a manual patron block the staff member should be able to select from a list of templates. For instance, this is needed if the same block is created over and over. This PR adds the possibility to select a block template from a list of templates. After selection, the fields that are defined by the template are automatically populated in the block's form.

## How?
This PR adds a new `<Selection>` component to `<PatronBlockForm>`. The selection contains the available block templates as options. When a user selects one option, the fields of the form in `<PatronBlockForm>` are changed according to the template.

## Screenshot
use_block_template


elsenhans (commented):
Hello Richard,
in the template I am able to set Block actions. If I am adding a template to a user probably the  selected Block actions should be the same like in the template, shouldn't they? Because at the moment are always all three Block actions are selected.

rchr:
@elsenhans, you're right. At the moment all three block actions are select. I didn't notice that before. Thanks for reporting that! I will fix it.

rchr:
@elsenhans the issue that alle three block actions are selected should be fixed now.

elsenhans (approved):
Looking good!

Baroquem (approved):

src/components/PatronBlock/PatronBlockForm.js
```diff
@@ -142,17 +157,60 @@ class PatronBlockForm extends React.Component {
...
+  }
+
+  onChangeTemplate = (id) => {
+    const { form: { batch, change, getRegisteredFields }, blockTemplates } = this.props;
+    const template = blockTemplates.find(t => t.id === id) || {};
+    const code = get(template, 'code', null);
+
+    // if no template is selected, all blockActions are set to true (default behavior when creating a block)
+    const blockActions = isEmpty(template);
+    batch(() => {
```
mkuklis, line 179: This is very nice @Baroquem!

mkuklis (approved):

src/routes/PatronBlockContainer.js
```diff
@@ -26,6 +26,11 @@ class PatronBlockContainer extends React.Component {
         path: 'manualblocks/%{activeRecord.blockid}',
       }
     },
+    blockTemplates: {
+      type: 'okapi',
+      records: 'manualBlockTemplates',
+      path:'manual-block-templates',
```
zburke, line 32: Without an explicit `limit` clause, you get the default limit of 10. Even if that likely meets your needs, it may be worth bumping to 100 or 1000 to be more certain it meets all tenants' needs.

zburke (approved):
LGTM!

PR #1582, opened 2020-11-25
UIU-1973 auto-focus Last Name field in user edit form

axelhunn:
Refs [UIU-1973](https://issues.folio.org/browse/UIU-1973)

In the User Edit form, when the form opens, there were no focused field by default.
Now focus goes to Last Name field.

### DIFF
--- a/src/components/EditSections/EditUserInfo/EditUserInfo.js
+++ b/src/components/EditSections/EditUserInfo/EditUserInfo.js
@@ -16,6 +16,8 @@ import {
 
 import css from './EditUserInfo.css';
 
+const autoFocusThisField = { autoFocus: true };
+
 class EditUserInfo extends React.Component {
   static propTypes = {
     accordionId: PropTypes.string.isRequired,
@@ -88,6 +90,7 @@ class EditUserInfo extends React.Component {
               name="personal.lastName"
               id="adduser_lastname"
               component={TextField}
+              props={autoFocusThisField}
               required
               fullWidth
             />
--- a/src/components/EditSections/EditUserInfo/EditUserInfo.js
+++ b/src/components/EditSections/EditUserInfo/EditUserInfo.js
@@ -16,8 +16,6 @@ import {
 
 import css from './EditUserInfo.css';
 
-const autoFocusThisField = { autoFocus: true };
-
 class EditUserInfo extends React.Component {
   static propTypes = {
     accordionId: PropTypes.string.isRequired,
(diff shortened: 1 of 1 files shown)


src/components/EditSections/EditUserInfo/EditUserInfo.js
```diff
@@ -88,6 +90,7 @@ class EditUserInfo extends React.Component {
               name="personal.lastName"
               id="adduser_lastname"
               component={TextField}
+              props={autoFocusThisField}
```
aliaksei-chumakou, line 93: is it better in something than just `autoFocus`?

axelhunn: less vague

zburke: Sorry, I'm with @aliaksei-chumakou on this one again.

aliaksei-chumakou: maybe "vague" - is too subjective thing, `props` is a reserved React word, so to me it's not obvious why we pass `id`, `required`, `fullWidth` separately, but `autoFocus` in other way. It's not an error to do like this, but maybe a question of project code style and consistency

axelhunn: About the prop `pops`, exactly!
But it's an API of [redux-form Field](https://redux-form.com/6.4.3/docs/api/field.md/)

And for me, add upon this this
```js
const autoFocus = { autoFocus: true };
```
would be too much for my taste ))

Can we maybe agree on `AUTO_FOCUS` or `AUTO_FOCUS_PROPS`, in all-caps, like the constants usually done?

aliaksei-chumakou: @axelhunn  right, but it will not work with react-final-form, IIUC https://final-form.org/docs/react-final-form/types/FieldProps

axelhunn: Should I make a universal thing like
```js
const TexFieldAutofocused = props => <TextField {...props} autoFocus />;
//...
<Field
  //...
  name="personal.lastName"
  component={TexFieldAutofocused}
/>
```

In final-form this would work just like this, in more elegant way
```js
<Field
  //...
  component={TextField}
  autoFocus
/>
```

aliaksei-chumakou: I think 
```
<Field
  //...
  component={TextField}
  autoFocus
/>
```
works with both frameworks, no?

axelhunn: Seems like it should work.
My mistake, have read the docs poorly.

axelhunn: fixed

aliaksei-chumakou (approved):

mkuklis (approved):

PR #1583, opened 2020-11-26
UIU-1907: Set expiration date on offset

elsenhans:
https://issues.folio.org/browse/UIU-1907

### Background:
In Settings, a user can define a default offset in days for the expiration date of each patron group. this is optional for each patron group.
If the patron group does have a defined offset, the expiration date field can be filled with the current date plus the given offset.
This automatically entered value can be edited or deleted by the user: it is only a help for choosing the right date, no restriction.

### Scenarios:
Users App, Create/Edit Screen.

(1)
Given: No patron group is selected.
Then: No button is displayed.

(2)
Given: A patron group is already selected (In edit screen)
(2a)
Given: The patron group has no offset defined in settings.
Then: No button is displayed.
(2b)
Given: The patron group does have an offset defined in settings.
Then: The button "Recalculate" is displayed underneath the field "expiration date".
When: The user clicks the button
Then: The expiration date is changed to the current date (server time) plus the number of days defined in the settings as offset.

(3)
Given: The user selects a patron group in the dropdown.
(3a)
Given: The patron group has no offset defined in settings.
Then: No button is displayed.
(3b)
Given: The patron group does have an offset defined in settings.
Then: A modal appears:
"Users with patron group group usually expire after offset days.
Do you want to change the expiration date accordingly?"
Options: "Cancel" - "Recalculate"
When: The user clicks the button "Cancel", Then: The field Expiration date remains unchanged.
When: The user clicks the button "Recalculate". Then: The expiration date is changed to the current date (server time) plus the number of days defined in the settings as offset.
After closing the modal, see (2b): The button still displays, because the user could manually change the filled in date and might want to recalculate again.

### In addition:

Given: An expiration date that is already entered an lies in the past, the status is "inactive", the information text "User has expired" is displayed in red underneath the field status.
When: The expiration date field is changed (manually or via button "Recalculate") to a value >= the current date
Then: An information text apperas instead of or underneath the "User has expired" info, reading: "User will reactivate after saving"


src/components/EditSections/EditUserInfo/EditUserInfo.js
```diff
@@ -71,112 +142,166 @@ class EditUserInfo extends React.Component {
       }
     ];
 
+    const offset = this.getPatronGroupOffset();
+    const group = _.get(this.props.patronGroups.find(i => i.id === this.props.selectedPatronGroup), 'group', '');
+    const footer = (
```
rchr, line 147: Please rename this constant to e.g. `modalFooter`. Then it is clearer that this is the footer of the modal.

src/views/UserEdit/UserForm.js
```diff
@@ -503,6 +503,7 @@ class UserForm extends React.Component {
                   onToggle={this.handleSectionToggle}
                   initialValues={initialValues}
                   patronGroups={formData.patronGroups}
+                  stripes={this.props.stripes}
```
rchr, line 506: Just a matter of style. I would prefer to add 'stripes' to the object desctructuring in L445 - L452. Then you can write `stripes={stripes}` here.

rchr (changes requested):
Good work! Thank you!

However, I do have a few small remarks. Mostly questions of style.

elsenhans:
@rchr, I fixed your style remarks. Thanks a lot for your feedback!

rchr (approved):
LGTM!

zburke (approved):
Nicely done!

PR #1585, opened 2020-12-01
UIU-1987 allow fetch of up to 1000 loans

zburke:
All the links we click with our cursors
All the fines we owe to the bursar
Will never be enough
Never be enough
One hundred loans was far too little
Patrons have more; a thousand? it'll
Have to be enough
Have to be enough

Refs [UIU-1987](https://issues.folio.org/browse/UIU-1987)


annamelnyk:
I believe even 1000 is temporary solution for the time being for loans or patron blocks since all history is saved. I am wondering should we probably increase it to 10000, I mean use `MAX_RECORDS` constant so it could be increased in single place when the time comes?

annamelnyk (approved):

SergiyVSergiyenko (approved):

mkuklis (approved):

Baroquem (approved):

zburke:
You may be right about `MAX_RECORDS`, @annamelnyk, though displaying 10k rows isn't going to be a great UX. I thought 1k would be enough to handle the vast majority of cases, though Cate recently suggested 2k.

PR #1590, opened 2020-12-08
UIU-1912: Fix selection of service points to display in the user record

SergiyVSergiyenko:
https://issues.folio.org/browse/UIU-1912

### Purpose
To be similar to the `Add permissions` functionality, the mechanism for selecting service points to be displayed in the user record has been changed. Now only the marked service points will be displayed, and not added to the existing ones, as it was before.

### Demo
chrome-capture


Baroquem (commented):
Thanks, Sergiy! This behavior makes a lot more sense.

mkuklis (approved):

PR #1596, opened 2020-12-28
UIU-1991: Add error message to action modal

annamelnyk:
# Description
Wrong error message appears when required field not selected on Pay/Waive/Transfer modal
# Issue
https://issues.folio.org/browse/UIU-1991
# Screeshots
**If action methods does not exist**
image

**If action methods exists but not selected**
image


maximdidenkoepam (approved):

src/components/Accounts/Actions/ActionModal.js
```diff
@@ -275,12 +275,24 @@ class ActionModal extends React.Component {
...
-    const { action } = this.props;
+    const {
+      action,
+      data,
+      accounts,
+    } = this.props;
+
+    const actionOptions = action === 'payment'
+      ? data.filter((d) => d.ownerId === accounts[0].ownerId)
+      : data;
```
zburke, line 286: On the one hand, creating and validating `actionOptions` feels wrong to me for two reasons:

1. we should be validating user input (`value`) only, not the state of the form itself.
2. we should be validating existing elements, not creating `actionOptions` here in the validation method. 

On the other hand, this does, in fact, resolve UIU-1991, so maybe I should not worry about abusing input validation in this manner. But ... it really feels like the `actionOptions` validation should happen as part of the form's initialization, not its validation. Then it could prevent the form from even displaying when `data` (and hence `actionOptions`) is empty because it is not possible for validation to succeed under that circumstance.

Dunno. Thoughts?

annamelnyk: Sounds right, but `ActionModal` component is using in different cases. And for multiple actions, lets say `payment`, we have the dropdown where `owner` can be selected and then we anyway should validate `payment/actionOptions` for already selected owner. So I suppose this should stay a part of validation neither form's initialization.

mkuklis (approved):

Dmitriy-Litvinenko (approved):

src/components/Accounts/Actions/ActionModal.js
zburke: I disagree that it should stay in validation. Regardless of where this component is used, `actionOptions` is not allowed to be empty (line 288); this is a form-configuration issue, not a data validation issue.

annamelnyk: @zburke I removed this logic from validation, not sure if there is the shorter way to write this.

PR #1597, opened 2020-12-29
UIU-1993: Avoid multiple charge

annamelnyk:
# Description
Hitting 'Charge only' button multiple times results in more than one fee/fine being created. Lets disable the `Charge only` button after click is done. If there exists better approach to prevent multiple form submit please let me know. 
# Issue
https://issues.folio.org/browse/UIU-1993


src/components/Accounts/ChargeFeeFine/ChargeForm.js
```diff
@@ -218,10 +221,11 @@ class ChargeForm extends React.Component {
         </Button>
         <Button
           id="chargeOnly"
-          disabled={pristine || submitting || !valid}
+          disabled={pristine || this.state.submitting || !valid}
           onClick={() => {
             change('pay', false);
             handleSubmit(data => onSubmit(data));
+            this.setState({ submitting: true });
```
maximdidenkoepam, line 228: As I see we can't submit again without leaving the page of refresh. Should we set `submitting` flag back to `false` after  submit was completed?

maximdidenkoepam (changes requested):

src/components/Accounts/ChargeFeeFine/ChargeForm.js
annamelnyk: I guess here we do not need it because after submit we will go to the one route back [here].(https://github.com/folio-org/ui-users/blob/bce2383f54b021a20934b9880d95877d1f7f31ee/src/components/Accounts/ChargeFeeFine/ChargeFeeFine.js#L408)

maximdidenkoepam: It's not a good practice to rely on another component in such cases. If onSubmit implementation will be changed -  we will receive a bug.

zburke: I don't entirely understand the flow here, nor why the previous implementation that relied on `props.submitting` was insufficient. In other words, I would expect the `disabled` implementation for both the "Charge and pay" and the "Pay only" buttons to be the same. Can you explain, @annamelnyk, why they are different?

zburke: (Related since you referenced it, but not really related to this PR: using `history.goBack()` in [`ChargeFeeFine.js`](https://github.com/folio-org/ui-users/blob/bce2383f54b021a20934b9880d95877d1f7f31ee/src/components/Accounts/ChargeFeeFine/ChargeFeeFine.js#L408) feels like the wrong implementation. We might indeed want to process the action and then "return" to the "previous" page, but only in the sense that we are visiting a given location a second time. I would expect `history.push(that previous location)`. That function could really use some comments too, or a reference to a wiki page or a ticket or something explaining the overall flow of the processes for charge, charge and pay, etc. Do you know if we have that kind of documentation anywhere?)

annamelnyk: @zburke as I know there no any documentation exist related to `Fee/fines`, I will ask Holly about  `history.goBack()`.

annamelnyk: Holly agreed to remove history.goBack(). Also with `props.submitting` is possible to click multiple times on button, so I created extra `formSubmitting` flag, and  added disabled implementation for both the "Charge&pay now" and the "Charge only"  buttons

src/components/Accounts/ChargeFeeFine/ChargeFeeFine.js
```diff
@@ -172,7 +174,12 @@ class ChargeFeeFine extends React.Component {
     );
 
     return this.props.mutator.accounts.POST(typeAction)
-      .then(() => this.newAction({}, typeAction.id, typeAction.feeFineType, typeAction.amount, commentInfo, typeAction.remaining, 0, typeAction.feeFineOwner));
+      .then(() => this.newAction({}, typeAction.id, typeAction.feeFineType, typeAction.amount, commentInfo, typeAction.remaining, 0, typeAction.feeFineOwner))
+      .then(() => {
+        this.formSubmitting = false;
+
+        return undefined;
```
maximdidenkoepam, line 181: Why do we need manually return `undefined`?

maximdidenkoepam (changes requested):

src/components/Accounts/ChargeFeeFine/ChargeFeeFine.js
annamelnyk: I did it because eslint throws error - `arrow function cannot return assignment`, and also if add empty `return` - eslint error will be `no useless return`

maximdidenkoepam (approved):

Dmitriy-Litvinenko (approved):

mkuklis:
@annamelnyk I apologize for the question here but I am a bit confused about the need for a new flag `formSubmitting`. Why can't we rely on what final form provides for us (`submitting` flag)?

My understanding was that if the promise is returned from the `onSubmit` handler that final form would correctly handle the value of `submitting` during the pending request. Could this be achieved by adding `return` here: 

https://github.com/folio-org/ui-users/pull/1597/files#diff-df62252520d98521d6857fdf68c645137e30c88f59dfc88a9d067d9ecbcba31bR407 

and here?

https://github.com/folio-org/ui-users/pull/1597/files#diff-df62252520d98521d6857fdf68c645137e30c88f59dfc88a9d067d9ecbcba31bR411

annamelnyk:
@mkuklis, after returning promise in `onSubmit` the react-final-form `submitting`  is changing it's value correctly,  which means the form is currently submitted asynchronously. Unfortunately I did not see this solution by myself, and I am happy that any extra flag didn't have to be added. Thanks much!

mkuklis (approved):

PR #1604, opened 2021-01-05
UIU-2004: Set informative message when error declaring item lost

SergiyVSergiyenko:
https://issues.folio.org/browse/UIU-2004

### Purpose
An error modal appears when an item is declared lost and a lost fee will be charged. In this case, a 
fee/fine owner for the item effective permanent location service point must be created.
This PR resolves the bug by displaying a more informative message explaining the cause and how to fix it.

### Screenshot
Screen Shot 2021-01-06 at 00 38 46


mkuklis (approved):
Nice work @SergiyVSergiyenko!

PR #1628, opened 2021-02-02
UIU-1996: Patron block accordion should be open if patron block exists and closed if not

Dmitriy-Litvinenko:
# Description
- Patron block accordion should be open if patron block exists and closed if not

# Stories
https://issues.folio.org/browse/UIU-1996

# Screenshots
image
image

### DIFF
--- a/src/components/UserDetailSections/PatronBlock/PatronBlock.js
+++ b/src/components/UserDetailSections/PatronBlock/PatronBlock.js
@@ -1,10 +1,10 @@
-import _ from 'lodash';
 import React from 'react';
 import PropTypes from 'prop-types';
 import {
   FormattedMessage,
   injectIntl,
 } from 'react-intl';
+import _ from 'lodash';
 import moment from 'moment';
 
 import {
@@ -137,7 +137,9 @@ class PatronBlock extends React.Component {
   onRowClick(e, row) {
     const {
       history,
-      match: { params }
+      match: {
+        params,
+      },
     } = this.props;
 
     if (!row.type) {
@@ -154,7 +156,7 @@ class PatronBlock extends React.Component {
   getPatronFormatter() {
     const {
       intl: {
-        formatMessage
+        formatMessage,
       },
     } = this.props;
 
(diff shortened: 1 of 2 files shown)


src/components/UserDetailSections/PatronBlock/PatronBlock.js
```diff
@@ -225,22 +232,25 @@ class PatronBlock extends React.Component {
...
         <FormattedMessage id="ui-users.settings.patronBlocks" />
         {(hasPatronBlocks) ? <Icon size="medium" icon="exclamation-circle" status="error" /> : ''}
       </Headline>;
 
     return (
       <Accordion
-        open={expanded}
+        data-test-patron-blocks-accordion
         id={accordionId}
+        open={!_.isEmpty(patronBlocks)}
```
annamelnyk, line 247: I am almost sure that current changes will broke the accordion. It will no be toggled. Please recheck it

src/components/UserDetailSections/PatronBlock/PatronBlock.js
```diff
@@ -183,16 +185,17 @@ class PatronBlock extends React.Component {
...
     const {
-      expanded,
       onToggle,
       accordionId,
       patronBlocks,
       hasPatronBlocks,
-      match: { params },
+      match: {
+        params,
+      },
```
annamelnyk, line 194: If we take out one field only, it should stay in one line, shouldn't it?

mkuklis (approved):

src/components/UserDetailSections/PatronBlock/PatronBlock.js
zburke: @annamelnyk is correct here; this change causes the accordion to be locked-open when `patronBlocks` is non-empty. I thought we could use the [`<Accordion>` prop `closedByDefault`](https://github.com/folio-org/stripes-components/tree/6087a2c1cb763163950f04b5e70086e1c7116bc1/lib/Accordion) to get what you want here, but I'm not having luck with that either. 

From reading the Jira it seems like there is a larger problem of accordion-state carrying over from one user to the next. Maybe we should put this off until a larger refactor can be done here?

zburke (changes requested):
Setting `open` to a constant value doesn't do what you want here. Additionally, the dozens of lines of reformatting throughout the file are a major distraction from the one line that is actually changing. 

If you want to reformat to obey a styleguide, thank you! But please do it in a separate PR.

zburke:
@Dmitriy-Litvinenko, what is your plan here? This has been hanging out for > two weeks with no changes.

Dmitriy-Litvinenko:
> @Dmitriy-Litvinenko, what is your plan here? This has been hanging out for > two weeks with no changes.

From reading the Jira it seems like there is a larger problem of accordion-state carrying over from one user to the next. Maybe we should put this off until a larger refactor can be done here?

PR #1636, opened 2021-02-05
STCOR-516 SWR POC

zburke:
`ControlledVocab` SWR POC from `stripes-smart-components` and
`stripes-core`. 

There's really nothing to see here; the API `<ControlledVocab>` doesn't change any. 

Refs [STCOR-516](https://issues.folio.org/browse/STCOR-516)


zburke:
I'm gonna wait for #1645 to land first, and then re-refactor stripes-smart-components ControlledVocab to SWR. Since the refactor will be totally internal with no outward changes to the API, this PR will become unnecessary.

PR #1637, opened 2021-02-08
UIU-2046 calculate expiration date differential

elsenhans:
https://issues.folio.org/browse/UIU-2046

There are two use-cases that need to be differentiated when a new expiration date is calculated

(1) Current expiration date is in the past
-> calculate from today
(2) current expiration date is in the future or today
-> calculate from given expiration date

Background:
Especially if there are annual user fees, a user who comes to renew a week before the expiration date would want to have a full new year starting from the current expiration date


mkuklis (approved):

rchr (approved):
Looks good!

PR #1642, opened 2021-02-11
UIU-1915 add app dropdown menu

elsenhans:
https://issues.folio.org/browse/UIU-1915

As a librarian
I want to use shortcut keys to navigate a Folio app(s)

**Screenshot use Agreements as a pattern**
image

**PR**
See https://github.com/folio-org/stripes-core/pull/598

**Requirement**
To the right of Users that appears to the left hand side of the screen, display a carat
Carat enables a dropdown
Dropdown should include the following option: Keyboard shortcuts
Selecting Keyboard shortcuts should open Keyboard shortcuts modal
Clicking on Users should behave no differently


elsenhans:
UIU-1915 will be done in UIU-1916

PR #1646, opened 2021-02-12
UIU-2050: Fix wrong value on reset data for Refunds to process manually report (CSV)

Dmitriy-Litvinenko:
# Description
- Fix wrong value on reset data for Refunds to process manually report (CSV)

# Stories
https://issues.folio.org/browse/UIU-2050

# Screenshots
image

### DIFF
--- a/src/views/UserSearch/UserSearch.js
+++ b/src/views/UserSearch/UserSearch.js
@@ -4,8 +4,12 @@ import React, { createRef } from 'react';
 import PropTypes from 'prop-types';
 import { matchPath } from 'react-router';
 
-import noop from 'lodash/noop';
-import get from 'lodash/get';
+import {
+  assign,
+  get,
+  each,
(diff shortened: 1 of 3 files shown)


src/views/UserSearch/UserSearch.js
```diff
@@ -433,8 +436,24 @@ class UserSearch extends React.Component {
       return ids;
     }, []);
 
+    const requestData = (startDate, endDate, feeFineOwners) => {
+      let requestData = {
```
annamelnyk, line 440: Why function and object have the same name? It looks confusing. Please change function name. 
And objects better declare with `const`.

Dmitriy-Litvinenko: Fixed

annamelnyk (approved):

zburke (approved):

PR #1647, opened 2021-02-15
UIU-1916 add dropdown show shortcuts

elsenhans:
https://issues.folio.org/browse/UIU-1916

As a librarian
I want to display a list of shortcut keys that I can use to navigate a Folio app(s) by keyboard.

**Requirements**
- IF a librarian clicks on Keyboard shortcuts menu option THEN display Keyboard shortcuts modal and focus should be on the modal.
- IF a librarian types/enters shortcut key ctrl / cmd + / THEN display Keyboard shortcuts modal and focus should be on the modal.
- Pane header should display <<selected App icon>> Keyboard shortcuts. So for Users app, it will be [Users app icon] Keyboard shortcuts
- Display shortcut keys based on user's operating system. If the operating system cannot be detected then show both OS and Windows compatible shortcut keys.
- This implementation will serve as the UX pattern for each app's shortcut keys display

**Keyboard shortcuts to display - Display in this order**
Action | OS shortcut | Windows shortcut
-- | -- | --
Create a new record | Option + n | alt + n
Edit a record | cmd + Option + e | ctrl + alt + e
Save a record | cmd + s | ctrl + s
Expand or collapse an accordion | spacebar | spacebar
Expand all accordions | cmd  + Option + b | ctrl + alt + b
Collapse all accordions | cmd  + Option + g | ctrl + alt + g
Go to Search & Filter pane | cmd + Option + h | ctrl + alt + h
Close a modal or pop-up | esc | esc
Copy | cmd + c | ctrl + c
Cut | cmd  + x | ctrl + x
Paste | cmd + v | ctrl + v
Find | cmd + f | ctrl + f


rchr (commented):
As this implementation will become a candidate to be a centralized component, I think we need to make it more general. For instance, making the modal an own component.

In addition, I don't think the actual shortcuts need to be translated. Is it possible to use the app-specfic shortcuts from [commands.js](https://github.com/folio-org/ui-users/blob/master/src/commands.js)?
Then maybe the possible modal component can use the app-specific shortcuts and just add the general shortcuts that are valid for most browsers (like ctrl + c, ctrl + f...).

elsenhans:
The order of the list with shortcuts is not exactly like in the chart in the description.
This is because the commands-file is defining all user-specific-shortcuts and for the general shortcuts a new file has been added.

rchr (commented):
Looks good!

However, I am wondering if it is possible to unify the presentation of the shortcuts? And is it possible to display the first letters of the shortcuts in uppercase?

As shown in the screenshot only some of the shortcuts have spaces before and after the "+" sign :
modal

rchr (approved):

mkuklis (approved):
The PR looks nice and clean but my knowledge about the keyboard shortcuts is limited so we should really wait for @JohnC-80's approval before this is merged.

src/components/KeyboardShortcutsModal/KeyboardShortcutsModal.js
```diff
@@ -0,0 +1,85 @@
+import _ from 'lodash';
```
mkuklis, line 1: @elsenhans could you please import just what you need from lodash?

`import { upperFirst } from lodash`;

should work do it here.

src/components/KeyboardShortcutsModal/KeyboardShortcutsModal.js
```diff
@@ -0,0 +1,85 @@
...
+  Button,
+  Modal,
+  ModalFooter,
+  MultiColumnList,
+} from '@folio/stripes/components';
+
+import commands from '../../commands';
+import commandsGeneral from '../../commandsGeneral';
+
+class KeyboardShortcutsModal extends React.Component {
```
mkuklis, line 15: @elsenhans this looks like a good candidate for a functional component.

src/components/KeyboardShortcutsModal/KeyboardShortcutsModal.js
elsenhans: I changed _lodash_ import to _upperFirst_ import.

src/components/KeyboardShortcutsModal/KeyboardShortcutsModal.js
elsenhans: I changed class component to functional component.
@mkuklis  Thanks for your feedback!

src/components/KeyboardShortcutsModal/KeyboardShortcutsModal.js
```diff
@@ -0,0 +1,85 @@
...
+import { FormattedMessage } from 'react-intl';
+import {
+  Button,
+  Modal,
+  ModalFooter,
+  MultiColumnList,
+} from '@folio/stripes/components';
+
+import commands from '../../commands';
+import commandsGeneral from '../../commandsGeneral';
```
JohnC-80, line 13: Apologies for delay - This looks great!! The only thing I'll think we'll need to do here to get this 'centralized' is to use some of these module-specific imports as props passed into the 'KeyboardShortcutsModal component where it is instanciated in `src/index.js` . - probably have an 'allCommands' prop that would accept the value you currently assemble internally from concatenating `commands` and `commandsGeneral`.

JohnC-80 (changes requested):
Looking great... I think this will only take just a little bit more to make it generic enough for centralization - basically moving the logic/command list assembly outside of the component itself and excepting it as a prop. Thank you!

src/components/KeyboardShortcutsModal/KeyboardShortcutsModal.js
elsenhans: Thanks for your your feedback.
The allCommand list is now passed into KeyboardShortcutsModal as a prop.

JohnC-80 (approved):

PR #1649, opened 2021-02-17
UIU-1645 and UIU-1988 Claim returned-1.5: Effect on SET COST fees for item status Aged to Lost and Multiple transfer accounts

iselagb:
https://issues.folio.org/browse/UIU-1645
https://issues.folio.org/browse/UIU-1988


zburke:
Can you please explain what is actually happening here? The PR description has no real content and the diff is overwhelmed with whitespace changes that make it really hard to tell what legitimately significant things are changing.

In general, if you need to reformat to deal with lint errors, it's helpful to do that in a separate PR where the formatting changes are the only changes.

iselagb:
@zburke, the indentation problems have already been fixed

PR #1653, opened 2021-02-24
UIU-2058 rewording for text on expiration date modal 

elsenhans:
https://issues.folio.org/browse/UIU-2058

The text on the **expiration data modal**  needs to be changed.
The new text should be as follows:

**Heading:**
"Set expiration date?"

**Text:**
"Library accounts with patron group group expire in offset days. Do you want to set this user’s account to expire on date?"

**Button:**
"Set" instead of "Recalculate"

 In addition, the text of the button in the users **edit form** needs to be changed:
**Button:** 
"Re-set" instead of "Recalculate".

 **Important:**
The SIG wishes the previewed expiration date in the modal to have the format like in "March 1, 2023".

The important parts of the text should be in bold - The UM SIG agreed on this in yesterday's meeting. So the text should be like this:

Library accounts with patron group **group** expire in **offset days**. Do you want to set this user’s account to expire on **date**?

For internationalisation purposes it is important that the date format will be consistent with the rest of dates displayed in FOLIO. It will also need to change according to the settings of the locale (tenant > language and localization > locale).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,6 +66,7 @@
 * Add Callout for Refund report. Refs UIU-2035.
 * Allow 0 as valid entry for Patron Block Limits. Refs UIU-1998.
 * Add "Save and close" button should be active immediately on refund report criteria modal. Refs UIU-2034.
+* Rewording for text on expiration date modal. Refs UIU-2058.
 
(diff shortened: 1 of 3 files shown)


mkuklis (approved):

rchr (approved):
LGTM!

PR #1654, opened 2021-02-24
UIU-2059 expiration date modal open bug

elsenhans:
https://issues.folio.org/browse/UIU-2059

**Overview:**
**Steps to Reproduce:**

1. Go to users app
2. Choose a user record with a patron group that has an expiration date offset defined in the settings
3. Open this user record for editing

**Expected Results:** The modal should only open on patron group change.
**Actual Results:** The modal opens right away, for every edit

This is a new bug, the behaviour has been correct before.


mkuklis (approved):
Thanks for catching this @elsenhans. I probably introduced this while working on refactoring to final form.

rchr (approved):
Looks good!

PR #1655, opened 2021-02-25
UIU-1954: Patron blocks: Allow for override when Renewing

Dmitriy-Litvinenko:
# Description
- Patron blocks: Allow for override of Renewing when logged in user has credentials

# Stories
https://issues.folio.org/browse/UIU-1954

# Screenshots
Implemented for
image
image
Patron Block Modal
image
Override Modal
image


annamelnyk (approved):

Dmitriy-Litvinenko:

src/components/PatronBlock/PatronBlockModalWithOverrideModal.js
```diff
@@ -0,0 +1,81 @@
...
+          open={overrideModal}
+          additionalInfo={additionalInfo}
+          onSetAdditionalInfo={this.onSetAdditionalInfo}
+          onClose={() => {
+            this.onCloseOverrideModal();
+            onOpenPatronBlockedModal();
+          }}
+          onSave={async () => {
+            this.onCloseOverrideModal();
+            this.onSetAdditionalInfo();
```
mkuklis, line 55: @Dmitriy-Litvinenko is there a reason `additionalInfo` is reset to an empty string here?

mkuklis (approved):
@Dmitriy-Litvinenko this looks good. I noticed that the `additionalInfo` is being passed via many components. It's probably because the way the modals are currently structured. I wonder if in the future we could refactor it a bit to use react context. 

I left one little question for you.

src/components/PatronBlock/PatronBlockModalWithOverrideModal.js
Dmitriy-Litvinenko: We need clear current message after we use renew  loan(s) by barcode. For case that we don't have possible for renew loan(s) by barcode we need use current message for override renewal by barcode. This was discussed with PO.

PR #1660, opened 2021-03-02
UIU-2045: Display Fees/fines on Loans pages

annamelnyk:
# Description
On SNAPSHOT, the Fine incurred is no longer displaying on Loan Details. It is supposed to be a link to the Fees/Fines History page (if more than one fee/fine has been billed) or the Fee/Fine Details page (if only one fee/fine has been billed). Right now in SNAPSHOT there is no way to link from a loan to the associated fees/fines.

Changes were added:
- rename `Fee/fine` label to `Fees/fines incurred` (on pages `Loans` and `Loan Details`)
- show the correct fee/fine amount (on pages `Loans` and `Loan Details`) --> show originally charged fee/fine's amount or if there few fees/fines -> show general sum of fees/fines
Handle correct navigation to fees/fines pages:
- if  loan occures with single fine --> nav to `Fee/Fine Details` page
- if  loan occures with few fines --> nav to `Fees/Fines History` page to the correct tab (`Open/Closed/All`) 

# Issue
https://issues.folio.org/browse/UIU-2045
# Screenshots
**Before**
image
image

**After**
image
image


zburke (approved):
LGTM! Thank you for refactoring the column-key to `feefineIncurred` to be consistent with the other keys while you were in there.

src/components/util/util.js
```diff
@@ -137,3 +137,7 @@ export function hasEveryLoanItemStatus(loans, itemStatus) {
 export function hasAnyLoanItemStatus(loans, itemStatuses) {
   return _.every(Object.values(loans), loan => itemStatuses.includes(loan?.item?.status?.name));
 }
+
+export function checkFeesFinesStatus(accounts, status) {
+  return accounts.every((account) => account.status.name.toLowerCase() === status.toLowerCase());
+}
```
zburke, line 143: Maybe rename to something like `accountsMatchStatus()` or `validateAccountsHaveStatus()? Functions named `check...` always kinda bug me because it isn't clear what happens if the check succeeds (nothing? side effect? return true?) or fails (throw? return false?). NBD though; just an observation.

mkuklis (approved):

src/components/util/util.js
annamelnyk: `accountsMatchStatus` sounds good. Renamed method

PR #1663, opened 2021-03-09
UIU-2068: fix eslint error in LoanDetails.js

Dmitriy-Litvinenko:
# Description
- Fix eslint error in LoanDetails.js

# Stories
https://issues.folio.org/browse/UIU-2068

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -72,6 +72,7 @@
 * Rewording for text on expiration date modal. Refs UIU-2058.
 * Patron blocks: Allow for override of Renewing when logged in user has credentials. Refs UIU-1954.
 * Loan Details no longer displays Fines incurred. Refs UIU-2045.
+* Fix eslint error in `LoanDetails.js`. Refs UIU-2068.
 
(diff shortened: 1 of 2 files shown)


zburke (changes requested):
~Unless you think this rule has merit, I'm inclined [turn it off directly in `eslint-config-stripes`](https://github.com/folio-org/eslint-config-stripes/pull/77). What do you think?~

zburke (commented):
Now that I've read [the ESLint rule](https://eslint.org/docs/rules/no-return-await) and [the articles it references](https://jakearchibald.com/2017/await-vs-return-vs-return-await/) 100 times, I'm inclined to say the "right" thing here is to simply remove `await` from the picture:
```
return this.onRenew();
```
rather than disable the lint rule, unless somebody can explain to me why `await` is necessary here. 

(Sorry for the whiplash; thank you for your patience while I'm learning.)

Dmitriy-Litvinenko:
@zburke @mkuklis @annamelnyk Could you please re-review?

mkuklis (approved):

zburke (approved):

annamelnyk (approved):

PR #1670, opened 2021-03-18
UIU-2076: Adapt shortcut modal

elsenhans:
https://issues.folio.org/browse/UIU-2076

**Expected Results:**

- IF a librarian types/enters shortcut key cmd + option + lowercase k (for Mac users) OR ctrl + alt + lowercase k (for PC users) THEN display Keyboard shortcuts modal and focus should be on the modal.
- Pane header should display <<selected App icon>> Keyboard shortcuts. So for Users app, it will be [Users app icon] Keyboard shortcuts

**Additional Information:**
https://issues.folio.org/browse/UIU-1916


elsenhans:
I am not sure if the shortcut Ctrl+shift+7 is the best choice. With ubuntu and chrome the shortcut is already occupied by selecting everything.

mkuklis (approved):
@elsenhans this looks good to me but I think we should wait to hear from @JohnC-80 to double check.

JohnC-80 (approved):

PR #1671, opened 2021-03-18
UIU-1619 Claim returned-2.8: Do not allow "actions" to be made on fees/fines for items that are "Claimed returned"

ccruz17:
https://issues.folio.org/browse/UIU-1619


ccruz17:
Hi @zburke 

Can you check my pull request?
Add the big test, but I don't understand why some lines of code are not covered. Thanks.

src/components/Accounts/ViewFeesFines/ViewFeesFines.js
```diff
@@ -209,51 +211,42 @@ class ViewFeesFines extends React.Component {
...
       ? accounts.reduce((memo, a) => (Object.assign(memo, { [a.id]: a })), {})
       : {};
+
+    this.updateToggle(checkedAccounts);
+    this.setState(({ allChecked }) => ({
+      allChecked: !allChecked
+    }));
+  }
+
+  updateToggle(checkedAccounts) {
```
zburke, line 229: It is difficult to guess what the `updateToggle` function will do. Typically, functions named like `toggleFoo()` will flip the boolean value they are associated with, but `updateToggle` doesn't quite fit that naming-pattern, and isn't as simple as flipping a boolean value.

zburke (approved):
Unrelated to this PR but I noticed it along the way: 

The [official react docs specifically warn against using `JSON.stringify()` in `shouldComponentUpdate`](https://reactjs.org/docs/react-component.html#shouldcomponentupdate).

PR #1675, opened 2021-03-24
UIU-2052: Add changes to indicate clickable for cursor

Dmitriy-Litvinenko:
## Purpose
We should  add clickable indicate  for cursor when we navigate to manual block.

## Approach
I according that we don't have possible add css class to row we add content wrapper for cell. Cursor should appear only for clickable block(manual block).

## Screenshot
image

# Stories
https://issues.folio.org/browse/UIU-2052

# Test
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 * Configure Jest/RTL. Refs UIU-2112.
 * Show user-readable message when user is not found. Fixes UIU-2081.
 * Fix Custom Fields error message by adding a missing permission. Fixes UIU-2104.
+* Add changes to indicate clickable for cursor. Refs UIU-2052.
 
 ## [6.0.0](https://github.com/folio-org/ui-users/tree/v6.0.0) (2021-03-18)
 [Full Changelog](https://github.com/folio-org/ui-users/compare/v5.0.9...v6.0.0)
--- a/src/components/UserDetailSections/PatronBlock/PatronBlock.js
+++ b/src/components/UserDetailSections/PatronBlock/PatronBlock.js
@@ -20,6 +20,8 @@ import { stripesConnect } from '@folio/stripes/core';
 
 import { calculateSortParams } from '../../util';
 
+import css from './PatronBlock.css';
+
 class PatronBlock extends React.Component {
   static manifest = Object.freeze({
     manualPatronBlocks: {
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Change history for ui-users
 
+## [7.0.0] (IN PROGRESS)
```
annamelnyk, line 3: We do not know if there will be major changes yet, so probably it should be `v6.1.0`, right?

src/components/UserDetailSections/PatronBlock/PatronBlock.js
```diff
@@ -158,9 +160,23 @@ class PatronBlock extends React.Component {
       },
     } = this.props;
 
+    const pointerWrapper = (content, isManual) => (
+      isManual
+        ? <div className={css.pointerWrapper}>content</div>
```
annamelnyk, line 165: It should be `{content}` here with curly braces or we will receive `content` as plain text

annamelnyk (changes requested):

CHANGELOG.md
Dmitriy-Litvinenko: @zburke @mkuklis I use next unreleased version from jira. Could you please clarify what version we should use?

src/components/UserDetailSections/PatronBlock/PatronBlock.js
Dmitriy-Litvinenko: Fixed

annamelnyk (approved):

CHANGELOG.md
mkuklis: @Dmitriy-Litvinenko I think `6.1.0` should be good here.

src/components/UserDetailSections/PatronBlock/PatronBlock.js
```diff
@@ -158,9 +160,23 @@ class PatronBlock extends React.Component {
...
+
     return {
-      'Type': f => f?.type ?? <FormattedMessage id="ui-users.blocks.columns.automated.type" />,
-      'Display description': f => f.desc || f.message,
+      'Type': f => {
+        const type = f?.type ?? <FormattedMessage id="ui-users.blocks.columns.automated.type" />;
+
+        return pointerWrapper(type, f?.type);
+      },
+      'Display description': f => {
```
mkuklis, line 175: @Dmitriy-Litvinenko apologies for asking you about it as it looks like the original code had that already too but I'm wondering if we could use some sort of key here instead of english versions for column names like `Display description` etc. 

You can review this code:
https://github.com/folio-org/ui-erm-comparisons/blob/aee68fa3eda3c81b1ba4058d54d86424a7ae0de0/src/components/EntitlementsAgreementsList/EntitlementAgreementsList.js#L21-L38

I think it can be achieved with the `columnMapping`

mkuklis (approved):
@Dmitriy-Litvinenko this looks good. I left one comment for you to consider.

zburke (approved):

CHANGELOG.md
zburke: True, we don't know. This is just the CHANGELOG so NBD either way. Arguably, the person who made the last release should have then pushed another PR to update `#master` to the next expected version, and adjusted the IN PROGRESS version here. I'll go find that person and remind them about their responsibilities. Oh, wait ... 😅

src/components/UserDetailSections/PatronBlock/PatronBlock.js
Dmitriy-Litvinenko: On my opinion we should do it out of scope current pull request. Technical task was created https://issues.folio.org/browse/UIU-2091 .
Could you please review it?
CC: @zburke @annamelnyk

CHANGELOG.md
Dmitriy-Litvinenko: Thank you, i change fix version to 6.1.0

zburke (approved):

PR #1676, opened 2021-03-24
UIU-2087: Note Edit Page - doesn't return to record page after click on Save&Close

BogdanDenis:
## Description
Redirect to User details page after a Note was edited

## Issues
[UIU-2087](https://issues.folio.org/browse/UIU-2087)

## Coverage
image


Vladyslav-Velytskyi (approved):

src/views/Notes/NoteEditPage.js
```diff
@@ -18,19 +18,7 @@ export default class NoteEditRoute extends Component {
...
-    } = this.props;
-
-    const { id } = match.params;
-    const noteViewUrl = `/users/notes/${id}`;
-
-    history.replace({
-      pathname: noteViewUrl,
-      state: location.state,
-    });
+    this.props.history.goBack();
```
Yuliia-Dovhal, line 21: Don't you think would be better to rename the function according to the updated logic?

Yuliia-Dovhal (approved):

zburke (commented):
`history.goBack()` has led to problems in the past ([UIU-1412](https://issues.folio.org/browse/UIU-1412)/#1094). Are you confident it is the correct implementation here, or would `history.push('/some/path')` be better?

BogdanDenis:
> `history.goBack()` has led to problems in the past ([UIU-1412](https://issues.folio.org/browse/UIU-1412)/#1094). Are you confident it is the correct implementation here, or would `history.push('/some/path')` be better?

You're right, I've updated my PR. Thanks for noticing the issue!

PR #1677, opened 2021-03-24
UIU-2077: Fix behavior of Confirm button when 'Claimed returned' item is resolved as 'Mark as missing'

SergiyVSergiyenko:
https://issues.folio.org/browse/UIU-2077

### Purpose
The cause of the bug was the attempt to read the `id` prop of the _undefined_ `user` when updating the local resource `activeRecord`.
This PR fixes this by checking the presence of the `user` before accessing its `id`.


mkuklis (approved):
Nice catch @SergiyVSergiyenko!

zburke (approved):

PR #1678, opened 2021-03-30
UIU-2098 Move Tenant/Bursar exports to Users/Transfer criteria

NikitaSedyx:
https://issues.folio.org/browse/UIU-2098


axelhunn (approved):

AndreiShumski2 (approved):

aliaksei-chumakou (approved):

uladislausamets (approved):

zburke (approved):

zburke:
@NikitaSedyx, it looks like this PR got stalled when BigTest started acting up in CI last month. How do you want to proceed? There are at least three options, maybe more:

* merge as-is; we can ignore the BigTest changes because they don't really matter any longer
* pull out the BigTest config changes; folks are still using BigTest locally and we shouldn't break that
* replace new BigTest test with Jest/RTL tests; long term, this is the right move, but I feel like it's unfair to ask for that here

NikitaSedyx:
@zburke to be honest I'm not sure it's required for R2 at this moment, scout team is disbanded and I don't who should proceed with it and when

PR #1681, opened 2021-03-31
UIU-2078: Reset form values after saving at New Fee/Fine

annamelnyk:
# Description
New Fee/Fine Creation Form Does Not Clear Out When You Charge the Fee/Fine. The success toast appears, the fee/fine is created, but the create form remains active with the data entered for the previous fee/fine still entered.

Was done as well in this part of code:
- replace `SafeHTMLMessage` with `FormattedMessage` because of the console warnings
- rewrite placeholder in form fields to get rid of console errors 
- clean up
# Issue
https://issues.folio.org/browse/UIU-2078


Dmitriy-Litvinenko (approved):

annamelnyk:
This changes do not need any more. Issue was replaced with https://issues.folio.org/browse/UIU-2117.

PR #1687, opened 2021-04-06
UIU-2081: Show user-readable message when user is not found

mkuklis:
https://issues.folio.org/browse/UIU-2081

user-not-found

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 6.1.0 (IN PROGRESS)
 
 * Configure Jest/RTL. Refs UIU-2112.
+* Show user-readable message when user is not found. Fixes UIU-2081.
 
 ## [6.0.0](https://github.com/folio-org/ui-users/tree/v6.0.0) (2021-03-18)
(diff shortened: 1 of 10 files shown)


axelhunn (approved):

axelhunn:
Great!
Will it cover the user's loan pages too?
Like https://folio-snapshot.dev.folio.org/users/df7f4993-8c14-4a0f-ab63-93975ab01c76/loans/view/cf23adf0-61ba-4887-bf82-956c4aae2260

zburke (approved):
👏 👏 👏

mkuklis:
@axelhunn I don't think  https://folio-snapshot.dev.folio.org/users/df7f4993-8c14-4a0f-ab63-93975ab01c76/loans/view/cf23adf0-61ba-4887-bf82-956c4aae2260 will be covered with this work since it looks like this is a different route/container:

https://github.com/folio-org/ui-users/blob/master/src/index.js#L181

But the pattern to fix it should be similar to the work in this PR.

mkuklis:
It looks like we have the same pattern in many places :/

https://github.com/folio-org/ui-users/search?q=getUser

I think some bigger refactoring will be required to handle this in clean way.

Baroquem (approved):

PR #1692, opened 2021-04-09
UIU-2111: Fix Notify patron box behavior for New fee/fine when default patron notice is set

Dmitriy-Litvinenko:
## Purpose
"Notify patron" box should be check marked for "New fee/fine" when default patron notice is set.

## Approach
When we have initial values for Fee/fine owner or we select them should work the same logic.

## Screenshot
image

# Stories
https://issues.folio.org/browse/UIU-2111

# Test
image


annamelnyk (approved):

Dmitriy-Litvinenko:
Hello @zburke @mkuklis Could you please review current pull request?

mkuklis (approved):

PR #1693, opened 2021-04-09
UIU-2117: Go back after fee/fine created

annamelnyk:
# Description
User should go back after creating new "fee/fine" page.

# Issue
https://issues.folio.org/browse/UIU-2117

image

### DIFF
--- a/src/components/Accounts/ChargeFeeFine/ChargeFeeFine.js
+++ b/src/components/Accounts/ChargeFeeFine/ChargeFeeFine.js
@@ -218,6 +218,12 @@ class ChargeFeeFine extends React.Component {
     });
   }
 
+  goBack = () => {
+    const { history } = this.props;
+
+    history.goBack();
+  }
+
   onClosePayModal() {
     this.setState({
       pay: false,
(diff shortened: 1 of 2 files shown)


Dmitriy-Litvinenko (approved):

mkuklis (approved):

src/components/Accounts/ChargeFeeFine/ChargeFeeFine.js
```diff
@@ -218,6 +218,12 @@ class ChargeFeeFine extends React.Component {
     });
   }
 
+  goBack = () => {
+    const { history } = this.props;
+
+    history.goBack();
```
ViktorSoroka07, line 224: See [comment](https://github.com/folio-org/ui-users/pull/1676#pullrequestreview-632384632) from @zburke in one of the PRs regarding goBack().

annamelnyk: Thanks for review, @ViktorSoroka07. But looks like here we can not just replace `history.goBack()` with `history.push(path)`, because there 3 places from where we can go to `Create fee/fine page`, and we need to bring user back to prev screen. As the solution, here we can:
- to leave `history,goBack()`
- to ask PO (Holly) if we should change acceptance criteria, and navigate to the one particular page (for example, `Fee/fines history`)  
 @zburke, what do you think?

Dmitriy-Litvinenko (approved):

PR #1694, opened 2021-04-09
UIU-1617 - Claim returned-4: Display billed fees/fines as "Suspended" on User Information when item claimed returned (plus add "Refund" total)

ccruz17:
### Story:
https://issues.folio.org/browse/UIU-1617

### Description 

Need to display total number of suspended fees/fines due to the patron claiming the items has been returned. Also display number of refunds the patron has received

### Result

image

### Tests:

image


src/components/UserDetailSections/UserAccounts/UserAccounts.js
```diff
@@ -51,14 +63,28 @@ const UserAccounts = ({
   useEffect(() => {
     const open = records.filter(account => account?.status?.name !== accountStatuses.CLOSED);
     const closed = records.filter(account => account?.status?.name === accountStatuses.CLOSED);
-    const openTotal = open.reduce((acc, { remaining }) => (acc + parseFloat(remaining)), 0);
+    const refunded = records.filter(account => (account?.paymentStatus?.name === refundStatuses.RefundedFully || account?.paymentStatus?.name === refundStatuses.RefundedPartially));
+    const loansClaim = get(resources, ['loansHistory', 'records'], []).filter(loan => loan?.action === loanActions.CLAIMED_RETURNED);
```
zburke, line 67: As in the line above with `account?`, you can avoid `_.get(...)` with code like `resources.loansHistory.records?.filter(...)`. There is no need to refactor existing code, but in general we are moving toward this [optional chaining syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining).

ccruz17: Ok Zack I will remember and I preferred to update the code.

zburke (approved):

zburke (approved):

PR #1697, opened 2021-04-14
UIU-2123: Fix disabling "Save & close" button for Manual patron block

Dmitriy-Litvinenko:
## Purpose
When we have invalid values Manual patron block "Save & close" button should be disabled

## Screenshot
image

## Stories
https://issues.folio.org/browse/UIU-2123

## Test
image

### DIFF
--- a/src/components/PatronBlock/PatronBlockForm.js
+++ b/src/components/PatronBlock/PatronBlockForm.js
@@ -350,7 +350,10 @@ class PatronBlockForm extends React.Component {
 export default stripesFinalForm({
   initialValuesEqual: (a, b) => _.isEqual(a, b),
   navigationCheck: true,
-  subscription: { values: true },
+  subscription: {
+    invalid: true,
+    values: true,
+  },
   mutators: { setFieldData },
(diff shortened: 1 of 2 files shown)


src/components/PatronBlock/PatronBlockForm.js
```diff
@@ -350,7 +350,12 @@ class PatronBlockForm extends React.Component {
 export default stripesFinalForm({
   initialValuesEqual: (a, b) => _.isEqual(a, b),
   navigationCheck: true,
-  subscription: { values: true },
+  subscription: {
+    pristine: true,
+    submitting: true,
```
mkuklis, line 355: @Dmitriy-Litvinenko `submitting` and `pristine` are turned on by default so you don't need to list them here:

https://github.com/folio-org/stripes-final-form/blob/master/lib/StripesFinalFormWrapper.js#L97-L98

Dmitriy-Litvinenko: Fixed

mkuklis (approved):

PR #1709, opened 2021-04-21
WIP: pane switching keyboard shortcuts - 

JohnC-80:
This implements keyboard shortcuts to cycle focus through the visible panes in the ui. It will ultimately be set up to focus the first-focusable element within the pane content. If none are available, it will focus the pane header. It will ultimately be set up to allow for a custom focus target within the Pane. This is a prototype. It isn't centralized. The API is guaranteed to change - this is just to see how it feels to use.

Implement `shift + pagedown` and `shift + pageup` as pane switchers.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -3,6 +3,7 @@ import { FormattedMessage } from 'react-intl';
 import PropTypes from 'prop-types';
 import queryString from 'query-string';
 import { hot } from 'react-hot-loader';
+import { getFirstFocusable } from '@folio/stripes-components/util/getFocusableElements';
 
 import { AppContextMenu, Route, Switch, IfPermission } from '@folio/stripes/core';
 import {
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -3,6 +3,7 @@ import { FormattedMessage } from 'react-intl';
 import PropTypes from 'prop-types';
 import queryString from 'query-string';
 import { hot } from 'react-hot-loader';
+import { getFirstFocusable } from '@folio/stripes-components/util/getFocusableElements';
```
zburke, line 6: Can we export `getFirstFocusable` from stripes-components so we don't have to reach in and get it by private path here?

src/index.js
```diff
@@ -26,6 +27,18 @@ import {
   NoteEditPage
 } from './views';
 import KeyboardShortcutsModal from './components/KeyboardShortcutsModal';
+import { PatronBlockMessage } from './components/PatronBlock';
+
+const modifiedCommands = [...commands,
+  {
+    name: 'nextPane',
+    shortcut: 'shift + pagedown'
```
zburke, line 35: FYI, no `pagedown` or `pageup` keys on a MacOS laptop keyboard. I know I know, WIP, POC, etc.

JohnC-80: Thank you for re-iterating this on this rather ancient POC... For persistence of the conversation, Does `fn up/down arrow` perform the same function/key as pageup/down? can you `shift + fn + arrow` on a mac?

PR #1707, opened 2021-04-20
UIU-2140 make circulation and feesfines more fully optional

skomorokh:
https://issues.folio.org/browse/UIU-2140

PR #1715, opened 2021-04-23
correctly include sectionInner

zburke:
Use `{sectionInner}` within JSX to evaluate the variable; without the
braces, it's just a string.

Bugfix for #1707 

Refs [UIU-2140](https://issues.folio.org/browse/UIU-2140)


Baroquem (approved):

skomorokh (approved):
Ooof, that was really dumb, sorry. Thanks Zak!

PR #1719, opened 2021-04-28
UIU-2151: Adopt import for keyboard shortcut modal

elsenhans:
https://issues.folio.org/browse/UIU-2151

With STCOM-811 the keyboard shortcut modal was implemented as a stripes component to make it easiliy available for all apps.
This component should now be adopted in the users app and the current code changed accordingly.


rchr (approved):
Looks good!

mkuklis (approved):

zburke (approved):

PR #1727, opened 2021-05-04
UIU-1555 Transfer Accounts listed in drop-down for Transfer Fee/Fine should be by Fee/Fine Owner

ccruz17:
Story:
https://issues.folio.org/browse/UIU-1555

Description
Add Fee/Fine Owner to the Transfer Fee/Fine modal as a drop-down selection.


src/components/Accounts/Actions/FeeFineActions.js
```diff
@@ -644,10 +644,12 @@ class Actions extends React.Component {
     const currentFeeFineType = feefines.find(({ feeFineType }) => feeFineType === account?.feeFineType);
     const currentOwnerId = servicePointOwnerId || currentFeeFineType?.ownerId || account?.ownerId;
     const currentOwner = owners.find(o => o.id === currentOwnerId) || {};
+    const accountTypes = this.props.accounts.map(a => a.feeFineType);
+    const checkSomeNotify = feefines.filter(feeFine => feeFine.actionNoticeId && accountTypes.includes(feeFine.feeFineType));
     const initialValues = {
       ownerId: currentOwnerId,
       amount: calculateSelectedAmount(this.props.accounts),
-      notify: !!(currentFeeFineType?.actionNoticeId || currentOwner?.defaultActionNoticeId),
+      notify: actions.transferMany ? checkSomeNotify.length > 0 : !!(currentFeeFineType?.actionNoticeId || currentOwner?.defaultActionNoticeId),
```
zburke, line 652: The nested conditions within the true/false conditions of the ternary are pretty brutal to wrap your head around. It works, but it might be easier to parse with something like this: 
```
let notify = false;
if (actions.transferMany) {
  if (checkSomeNotify.length > 0) {
    notify = true;
  }
} else {
  if (!!(currentFeeFineType?.actionNoticeId || currentOwner?.defaultActionNoticeId)) {
    notify = true;
  }
}

File: src/components/Accounts/Actions/ActionModal.js [hunk 2]
```diff
@@ -336,9 +345,16 @@ class ActionModal extends React.Component {
...
 
     const selected = calculateSelectedAmount(accounts, this.isRefundAction(action), feeFineActions);
-    const ownerOptions = owners.filter(o => o.owner !== 'Shared').map(o => ({ value: o.id, label: o.owner }));
+    const ownersServicePoint = [];
+    owners.forEach((item) => {
+      if (item.servicePointOwner.filter(s => s.value === curServicePoint.id).length > 0) {
+        ownersServicePoint.push(item);
+      }
+    });
+    const ownerOptions = ownersServicePoint.filter(o => o.owner !== 'Shared').map(o => ({ value: o.id, label: o.owner }));
```
zburke, line 354: There are a lot of these kinds of list filters throughout the fee-fine codebase. It would be really helpful if we had comments explaining what they do, and what they're for, e.g. 
```
// only fee-fine owners affiliated with the current SP should be displayed.
// filter out all the others. 
const ownersServicePoint = [];
...
```
as it is, we have to read the code to figure out what's happening, and then find the associated ticket to figure out why.

zburke (approved):
As you noted on the ticket, the "Fee/fine owner desk" menu will only ever have one option because it's tied to the current service point. I'm not sure why it's a menu then, but I guess that's up to Holly, not us!

PR #1750, opened 2021-05-21
Use useMemo() correctly

zburke:
Refactor `useMemo()` to use it correctly:
* it does not return a promise
* include all dependencies in its arguments

~Also, lint all the things, which I should have done in a separate PR. Bad developer! No biscuit!~

@annamelnyk, does this refactor of `useMemo` look correct to you?

~@JohnC-80, any idea why `dismissible` seems to trip up ESLint but `open` doesn't?~

I'm moving linting to a separate PR, #1751.


mkuklis (approved):
Nice cleanup @zburke!

src/components/ReportModals/CashDrawerReportModal.js
```diff
@@ -65,24 +65,20 @@ const CashDrawerReportModal = (props) => {
...
+        .then(response => {
+          const responseSources = response?.sources ?? [];
+          const formatSources = responseSources.map((sourceName, index) => ({
+            value: `source-${index}`,
+            label: sourceName,
+          }));
+          setSources(formatSources || []);
+        });
+    }
+  }, [cashDrawerReportSources, servicePointsValue]);
```
annamelnyk, line 81: This works correct @zburke! I feel bad that you did all this clean up work after me! I was going to discuss if lint rule ("React Hook useMemo has a missing dependency: 'getSources'. Either include it or remove the dependency array  react-hooks/exhaustive-deps") should be off. And I learned that we need to include all dependencies in its arguments.
Thanks, Zak!

annamelnyk (approved):

src/components/ReportModals/CashDrawerReportModal.js
zburke: @annamelnyk, no worries. "all this clean up" was really just moving one function, and I only learned that it was necessary from reading the test and lint output. We are both learning things here!

src/components/ReportModals/CashDrawerReportModal.js
```diff
@@ -65,24 +65,20 @@ const CashDrawerReportModal = (props) => {
...
-      label: sourceName,
-    }));
-    setSources(formatSources || []);
 
-    return undefined;
-  };
-
-  useMemo(() => getSources(servicePointsValue), [servicePointsValue])
-    .then(response => response);
+  useMemo(() => {
```
ViktorSoroka07, line 69: In this case, I would expect to see `useEffect` then.  `useMemo` is more for storing the value which will be passed somewhere later (e.g. as props to child component)

PR #1752, opened 2021-05-24
UIU-2166: When resolving a claimed returned situation, the lost item …

iselagb:
When resolving a claimed returned situation, the lost item fees are not cancelled if they have had a transfer made before they are claimed returned

https://issues.folio.org/browse/UIU-2166


zburke (approved):
I understand what is happening at the feature level, i.e. what the Jira ticket says, but it is difficult to see how the code changes here implement. In future tickets, it would be very helpful if comments in the code or in the PR description describe not just what is changing (which we can see with `git diff`) but why the changes have the effect they do, e.g. why does `createRefunds` need to be `async`, why does the `CREDITED_ACTION` need to happen before the `REFUNDED_ACTION` and what was the implication of how it was implemented before?

PR #1756, opened 2021-05-26
UIU-2157: Fix selecting current fee fine type

Dmitriy-Litvinenko:
## Purpose
Fix selecting current fee fine type

## Stories
https://issues.folio.org/browse/UIU-2157

## Screenshot
error
charge   pay now
waive


annamelnyk (approved):

mkuklis (approved):

zburke (approved):
Wow; sharp eyes; surgical fix, @OleksandrVidinieiev! Nice work! This failure of searching by name instead of ID should serve as a cautionary tale we should all be aware of.

PR #1763, opened 2021-06-04
UIU-2184 CashDrawerReconciliationReportCSV should handle empty data

zburke:
An empty data array would cause an NPE in `parse()` because it did not
validate the return value from `buildReport()`.

Refs [UIU-2184](https://issues.folio.org/browse/UIU-2184)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,7 @@
 * Added `NotePopupModal` to User Details page. Refs UIU-2008.
 * Fix selecting current fee fine type. Refs UIU-2157.
 * Fix validation error with END DATE for `Cash drawer reconciliation report` modal. Refs UIU-2175.
+* Handle empty report data in `CashDrawerReconciliationReportCSV`. Refs UIU-2184.
 
 ## [6.0.0](https://github.com/folio-org/ui-users/tree/v6.0.0) (2021-03-18)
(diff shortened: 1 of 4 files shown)


Baroquem (approved):

annamelnyk:
Hm, but we handle case when we receive empty reportData with error callout [here](https://github.com/folio-org/ui-users/blob/82529763377a97c3837c398d4b6f04e9c677c80b/src/views/UserSearch/UserSearch.js#L532).
Do we need to handle this in `CashDrawerReconciliationReportCSV` as well?

zburke:
> Hm, but we handle case when we receive empty `reportData` with error callout here.
> Do we need to handle this in `CashDrawerReconciliationReportCSV` as well?

Aha, thanks for pointing that out, @annamelnyk! I stumbled on this accidentally by noticing that `ReportCSV` contained only [one un-tested condition](https://github.com/folio-org/ui-users/blob/ac12dd5627c1698a1f6e0a414eabfd525d5d787a/src/components/data/reports/cashDrawerReconciliationReportCSV.js#L63-L65), which happened to be the "empty `reportData` condition", and when I wrote the test I found the NPE. Probably, `ReportCSV` will never be called separately from `UserSearch`, but since it could be, it seems worthwhile to correctly handle the condition in `ReportCSV` too. 

Besides, 100% test coverage makes the gods happy :)

annamelnyk (approved):

PR #1768, opened 2021-06-09
UIU-1971: Delete user with transactions check

elsenhans:
https://issues.folio.org/browse/UIU-1971

Description:
Given that a specific user record has been selected by a permissioned staff person when the record is opened in the third pane, the action menu should show a function "check for open transactions/delete user"

Given a staff person selects "check for open transactions/delete user,"
and the user does not have any open transactions
then the pop up will say: 

"Are you sure you want to delete this user **[last name], [first name]**?
Yes No"

If "no" is selected, the display will go back to the user view.
If "yes" is selected, the user will be deleted.

Given the user has open transactions,
when the check is done,
the popup modal will say:

"User **[last name], [first name]** has the following open transactions:

- loans
- requests
- fines/fees
- blocks
- unexpired proxy

Please resolve the transactions to proceed to delete the user.

Ok

This modal will rely on the data provided by the list defined in MODUSERBL-115.


src/views/UserDetail/UserDetail.js
```diff
@@ -309,6 +338,53 @@ class UserDetail extends React.Component {
...
+  }
+
+  showDeleteUserModal(json) {
+    this.setState({
+      showDeleteUserModal: true,
+      openTransactions: json,
+    });
+  }
+
+  doClose = () => {
```
rchr, line 355: Please use a proper name specifying what is going to be closed. Currently, it may suggest that the UserDetail view is closed.

src/views/UserDetail/UserDetail.js
```diff
@@ -309,6 +338,53 @@ class UserDetail extends React.Component {
...
+
+  selectModal(transactions) {
+    if (!transactions.hasOpenTransactions) {
+      this.showDeleteUserModal();
+    } else {
+      this.showOpenTransactionsModal(transactions);
+    }
+  }
+
+  doFetch() {
```
rchr, line 370: Please use a proper name specifying what is going to be fetched.

src/views/UserDetail/UserDetail.js
```diff
@@ -431,6 +523,8 @@ class UserDetail extends React.Component {
     } = this.state;
 
     const user = this.getUser();
+    const username = getFullName(user);
+    const userId = this.props.match.params.id;
```
rchr, line 527: `match` is already destructured. You can use it directly here: `const userId = match.params.id;`

src/views/UserDetail/UserDetail.js
```diff
@@ -309,6 +338,53 @@ class UserDetail extends React.Component {
...
+      this.showOpenTransactionsModal(transactions);
+    }
+  }
+
+  doFetch() {
+    const { stripes } = this.props;
+    const okapiUrl = stripes.okapi.url;
+    const userId = this.props.match.params.id;
+
+    fetch(`${okapiUrl}/bl-users/by-id/${userId}/open-transactions`, {
```
rchr, line 375: Is it possible to specify the GET open-transactions request as a mutator in \<UserRecordContainer\>? 

It may look like:
 ```
openTransactions: {
      type: 'okapi',
      path: 'bl-users/by-id/:{id}/open-transactions',
      accumulate: true
    },'
```
Then all requests are specified in one place. In addition, you wouldn't need to specify the `httpHeaders` in this file.

**Review by rchr — CHANGES_REQUESTED** [review]
Is it possible to write tests for \<UserDetail\> to test that the correct modal is shown in correct circumstances?
This may also increase the test coverage. At the moment it is very low at 17%.

I also made some comments inline.

**Comment by elsenhans** [comment]
@mkuklis  I think the code coverage is going to be measured for the complete UserDetail.js, isn't it?

File: src/routes/UserSearchContainer.js [hunk 5]
```diff
@@ -35,7 +35,7 @@ class UserSearchContainer extends React.Component {
       records: 'users',
       resultOffset: '%{resultOffset}',
       perRequest: 100,
-      path: 'users',
+      path: 'users?unused=%{usersReloadToggle}',
```
mkuklis, line 38: @elsenhans this should not be required. The mutations should refresh corresponding resources automatically. The refresh can be fined tuned with `shouldRefresh` if required in a similar fashion to:

https://github.com/folio-org/ui-users/blob/74220d574ab43d24fe9c3d654de665c3ee7dcca2/src/routes/LoanDetailContainer.js#L31

mkuklis (changes requested):
@elsenhans `usersReloadToggle` should not be required here. stripes-connect should already handle the refresh when a mutation happens.

src/routes/UserSearchContainer.js
elsenhans: Thanks! Your solution is much better.

mkuklis (approved):
Great work @elsenhans!

src/views/UserDetail/UserDetail.js
```diff
@@ -431,6 +505,8 @@ class UserDetail extends React.Component {
     } = this.state;
 
     const user = this.getUser();
+    const username = getFullName(user);
```
zburke, line 508: This is petty, but given that "username" has a specific meaning in this context (i.e. `username` is an attribute of a user) and that's not the value being stored, it's a poor choice for this variable.

zburke (changes requested):
Please add translations for the new permissions, `permission.delete` and `permission.opentransactions` and make sure the values accurately reflect the permissions being granted.

src/views/UserDetail/UserDetail.js
```diff
@@ -309,6 +328,45 @@ class UserDetail extends React.Component {
...
+  doFetchOpenTransactions() {
+    const { mutator } = this.props;
+    const userId = this.props.match.params.id;
+
+    mutator.openTransactions.GET({ userId })
+      .then((response) => {
+        this.selectModal(response);
+      });
+  }
+
```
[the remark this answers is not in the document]
elsenhans: Thanks, I did rename `doFetchOpenTransactions` to `handleDeleteClick`.

src/views/UserDetail/UserDetail.js
elsenhans: I did rename `username` to `fullNameOfUser`

elsenhans:
Thanks a lot for your response @zburke !
I completely forget about the translation of the permissions, but I did add it know and also adapt the displayName.

src/routes/UserRecordContainer.js
```diff
@@ -27,6 +27,17 @@ class UserRecordContainer extends React.Component {
...
+      type: 'okapi',
+      path: 'bl-users/by-id/:{id}',
+      shouldRefresh: () => false,
+    },
+    openTransactions: {
+      type: 'okapi',
+      accumulate: 'true',
+      path: 'bl-users/by-id/:{id}/open-transactions',
+      fetch: false,
+    },
```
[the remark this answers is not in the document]
elsenhans: Well, I think the decision for adding this new endpoint was that the check need to be done in mod-users-bl. The requirement and decisions are described here:
https://wiki.folio.org/display/DD/Check+if+user+can+be+deleted

elsenhans:
@zburke Any suggestions how to fix the build?

src/routes/UserRecordContainer.js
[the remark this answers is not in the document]
elsenhans: The response of the new endpoint is just looking like this:
```
{
    "userID": "uuid-1234",
    "deletable": false,
    "loans": 1,
    "requests": 0,
    "fees/fines": 2,
    "proxies": 0,
    "blocks": 0
}
```
So, we are not going to fetch all the data twice.

zburke:
@elsenhans, build fixed via #1778 (UIU-2190)

zburke (approved):

PR #1769, opened 2021-06-09
Add missing claimed-returned button

mkuklis:
Add missing claimed-returned button


zburke (approved):
Yeah, this does what https://github.com/folio-org/ui-users/commit/f1cdb2754e79e522fb026855413b1e479a577414 was intended to do, and what the original UIU-2125 PR (#1708) did. 

Should we also want to remove the duplicate/un-permission-checked `export-refunds-report` button that was created in f1cdb2754e79e?

mkuklis:
Good catch @zburke! I just removed that extra button.

PR #1771, opened 2021-06-10
Fixed accordion open/close in user form

doytch:
The Expand All button wasn't working in the User form because
the proxy accordion had the wrong ID/state pairing. I went
ahead and just ported it over to the new `AccordionStatus`
component so that the individual accordion state didn't need
to be kept track of and risk stuff like this again.

Note: The diff view is cluttered because of indenting, so hiding whitespace changes makes it more obvious what changed.


mkuklis (approved):
Nice cleanup 😃!

PR #1779, opened 2021-06-23
UIU-2191: Prevent fetching delUser

elsenhans:
https://issues.folio.org/browse/UIU-2191

Steps to Reproduce:
    Select a user

Expected Results:
User details should be loaded without fetching data of resource 'delUser'.

Actual Results:
ERROR: in module @folio/users, operation GET on resource 'delUser' failed, saying: Index 0 out of bounds for length 0.


zburke (approved):
Whoops; good catch. Sorry I missed this!

mkuklis (approved):

PR #1792, opened 2021-07-08
UIU-2204: Increase report modals height

annamelnyk:
# Description
The `date picker` for report modals is cut off and user is not able select date
# Issue
https://issues.folio.org/browse/UIU-2204
# Screenshot
**Before**
image

**After**
image


src/components/ReportModals/ReportModal.css
```diff
@@ -0,0 +1,3 @@
+.content {
+  height: 350px;
```
Dmitriy-Litvinenko, line 2: Can we use min-height instead of height?

mkuklis (approved):

PR #1805, opened 2021-07-22
UIU-2215: Default notice not sent to patron when Transfer done in one of the three ways

Dmitriy-Litvinenko:
## Purpose
We should show "Notify patron" in three ways

## Approach
The transfer of a single fee/fine may be executed in three different ways:
1. Using the TRANSFER button in Fee/Fine Details.
2. Using the TRANSFER option in the ellipsis menu in Fees/Fines History.
3. Using the TRANSFER button in Fees/Fines History after putting a check mark in the box for a single fee/fine.

We have inconsistency behavior in case 3.

After discussing this issue with "PO", we should fix this issue as follows:
1. We must have consistent behavior in all 3 cases.
2. We must have the same behavior as we have on "iris"
3. If we will need provide any changes in behavior for "Notify patron" we should provide them in separate story/task.

## Stories
https://issues.folio.org/browse/UIU-2215

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 * Omit empty `username` during user creation. Fixes UIU-2214.
 * Fix `Total owed amount`/`Total paid amount` on `Fee/Fine Details`. Refs UIU-2211.
 * Whitespace should not mark loan-action forms dirty. Refs UIU-2227.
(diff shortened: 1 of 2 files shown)


Dmitriy-Litvinenko:

mkuklis (approved):

PR #1806, opened 2021-07-22
UIU-2227 whitespace should not make loan-action forms dirty

zburke:
Additional information in loan-action forms is a required field;
whitespace alone should not be acceptable as valid.

Refs [UIU-2227](https://issues.folio.org/browse/UIU-2227)


Baroquem (approved):

Baroquem:
The only problem I see is a typo in the PR title!

mkuklis (approved):

PR #1808, opened 2021-07-23
UIU-2209: Fix Export of fees/fines.

Dmitriy-Litvinenko:
## Purpose
Fix Export of fees/fines

## Approach
We should ignore fees/fines actions that don't have accounts for current user.

## Stories
https://issues.folio.org/browse/UIU-2209


src/components/data/reports/FeeFineReport.js
```diff
@@ -77,6 +77,11 @@ class FeeFineReport {
 
     map(feeFineActions, (action) => {
       const account = accounts.find(({ id }) => id === action.accountId);
+
+      if (!account) {
+        return;
+      }
+
```
zburke, line 84: Maybe use `forEach()` instead of `map()` since we're iterating for the side-effect, not transforming the elements and using the return value? Dunno, and I realize that code isn't even part of the PR here, but that naked `return;` statement just feels kinda odd to me in this context.

zburke (approved):

src/components/data/reports/FeeFineReport.js
annamelnyk: Hm.. I do not think that is matter what we are using `forEach` or `map`, anyway this line is the only place where we work directly with `feeFineActions` here, further we work with entity `reportData`

PR #1809, opened 2021-07-23
UIU-2012: Dept. label shows up in User's UI even when no depts are set up in Settings

mariiaschv:
Hide Department label in User's Detail and Edit views if there are no depts set up in Settings.

Issue: [UIU-2012](https://issues.folio.org/browse/UIU-2012)


Vladyslav-Velytskyi (approved):

zburke (approved):
This looks good to me, but please note the BigTest tests no longer run as part of the CI process and should be replaced by tests written with Jest/RTL.

PR #1812, opened 2021-07-27
UIU-2235 consume mocks from stripes-testing

zburke:
Instead of keeping a local copy of mocks for peer-dependencies, consume
the public mocks from `@folio/stripes-testing`.

Refs [UIU-2235](https://issues.folio.org/browse/UIU-2235)


zburke:
In the end, this just feels too risky: changes here have the potential to break tests in repositories that are totally disconnected from this one. stripes-testing should provide a stable interactors API, but mocks should remain local.

PR #1813, opened 2021-07-27
UIU-2229: Disable renewals for inactive users

Baroquem:
https://issues.folio.org/browse/UIU-2229

When loading the loan detail or loan history views, this checks the user record to see whether the `active` flag is set and the expiration date hasn't expired yet (which is probably redundant).


zburke (approved):
LGTM, but can you added a test for the `checkUserActive()` function? I know it seems trivial, but when it comes to coverage, every little bit helps. 

You can use [ui-requests's util.test.js](https://github.com/folio-org/ui-requests/blob/996c0ee3bd0c9e69e9ca2b4a58c74a926e6c6b0b/src/utils.test.js) as a template.

Baroquem:
Thanks, I was hunting around for an example to follow -- I haven't done much jesting yet.

PR #1815, opened 2021-07-30
Release v6.1.2

zburke:


zburke:
Alas, it looks like one of those cherry-picks didn't come over cleanly.

PR #1821, opened 2021-08-04
UIU-2217: Fix hyperlink in Financial transactions detail report

Dmitriy-Litvinenko:
## Purpose
Fix hyperlink in Financial transactions detail report

## Approach
We should not render hyperlink for "Item Barcode", "Loan Policy", "Overdue Policy", "Lost Item Policy", "Loan Details" when we don't have all necessary data.

## Stories
https://issues.folio.org/browse/UIU-2217

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@
 * Show dates in local time when generating CSV reports. Fixes UIU-2224.
 * Fix user Departments value is not visible in the user view. Fixes UIU-2238.
 * Filter out non existing service points. Fixes UIU-2245.
+* Fix hyperlink in `Financial transactions detail report`. Refs UIU-2217.
 
 ## [6.1.0](https://github.com/folio-org/ui-users/tree/v6.1.0) (2021-06-18)
 [Full Changelog](https://github.com/folio-org/ui-users/compare/v6.0.0...v6.1.0)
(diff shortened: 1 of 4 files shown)


mkuklis (approved):

annamelnyk (approved):

zburke (approved):
Intuitively, this fix looks great. 

That said, it would be great to pull that `parse()` function, or the `reportData.map(row, function)` code out to a separate file so you can write a simple unit test for it. There is [already a test file](https://github.com/folio-org/ui-users/blob/7f584a563a855943cbb28c2a2c1ab5d25e02bc25/src/components/data/reports/FinancialTransactionsReport.test.js) so you just have to add handling for these new conditionals.

Dmitriy-Litvinenko:
> Intuitively, this fix looks great.
> 
> That said, it would be great to pull that `parse()` function, or the `reportData.map(row, function)` code out to a separate file so you can write a simple unit test for it. There is [already a test file](https://github.com/folio-org/ui-users/blob/7f584a563a855943cbb28c2a2c1ab5d25e02bc25/src/components/data/reports/FinancialTransactionsReport.test.js) so you just have to add handling for these new conditionals.

We moved business logic for hyperlink creation to separate file financialTransactionsReportHelpers.js
We add tests for all cases

PR #1828, opened 2021-08-11
UIU-2242: Fix hyperlink with missing `Patron barcode`

Dmitriy-Litvinenko:
## Purpose
Fix hyperlink with missing `Patron barcode`

## Approach
We should not render hyperlink for `Patron barcode` when we don't have all necessary data.
We should render `Patron barcode` hyperlink with `no barcode` label when Patron don't have barcode.

## Stories
https://issues.folio.org/browse/UIU-2242


zburke (approved):
nice work!

mkuklis (approved):

PR #1829, opened 2021-08-16
UIU-2251: Fix Invalid date for Financial transaction detail report

Dmitriy-Litvinenko:
## Purpose
Fix `Invalid date` for `Financial transaction detail report`

## Approach
We get date from BE already formatted or empty string and we need render it as is

## Stories
https://issues.folio.org/browse/UIU-2251


zburke (approved):
This process of storing formatted data on the backend feels wrong to me, like we are going at this whole thing upside down, but I recognize that changing that process is not on the table here and all we need to do is stop feeding the date values through formatters multiple times.

annamelnyk (approved):

PR #1848, opened 2021-09-07
UIU-2411: Automatic fees/fines are appearing in New Fee/Fine "Fee/fine type" drop-down

annamelnyk:
# Description
There are now  is displaying the automated fee/fine types when you create a new fee/fine.  But we cannot have users creating manual fees/fines using the automated fee/fine types. So let's check if `fee/fine type` is automatic when we form `feefineList`.
# Issue
https://issues.folio.org/browse/UIU-2411
# Screenshots
**Before**
image
**After**
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@
 * Support `feesfines` interface version `17.0`. Refs UIU-2248.
 * Update sub permissions in `ui-users.edituserservicepoints` permission set. Fixes UIU-2244.
 * Replace `babel-eslint` with `@babel/eslint-parser`; import global babel config. Refs UIU-2253, UIU-2254.
+* Automatic fees/fines are appearing in New Fee/Fine `Fee/fine type` drop-down. Refs UIU-2411.
 
 ## [6.1.0](https://github.com/folio-org/ui-users/tree/v6.1.0) (2021-06-18)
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -33,6 +33,7 @@
 * Support `feesfines` interface version `17.0`. Refs UIU-2248.
 * Update sub permissions in `ui-users.edituserservicepoints` permission set. Fixes UIU-2244.
 * Replace `babel-eslint` with `@babel/eslint-parser`; import global babel config. Refs UIU-2253, UIU-2254.
+* Automatic fees/fines are appearing in New Fee/Fine `Fee/fine type` drop-down. Refs UIU-2411.
```
Stanislau-Snapok, line 36: should it be anything like "Fix issue when ..."?

Stanislau-Snapok (approved):

Dmitriy-Litvinenko (approved):

CHANGELOG.md
annamelnyk: I think it could be this way. As I know we do not have any requirements for changelog file, and we can compose by ourselves the changelog entry or I prefer to use JIRA's title.

Stanislau-Snapok: ok. asked just for information

mkuklis (approved):

PR #1849, opened 2021-09-07
UIU-2207 Suspended balance for page & Outstanding balance for page, is displaying total balance for the patron rather than the page

iselagb:
JIRA: https://issues.folio.org/browse/UIU-2207

image


zburke:
Replaced by #1865 due to  [FOLIO-3290](https://issues.folio.org/browse/FOLIO-3290)

PR #1851, opened 2021-09-08
UIU-2128: When item Claimed Returned, transfers are refunded but patron payments are not

ymacedor:
[UIU-2128](https://issues.folio.org/browse/UIU-2128): When item Claimed Returned, transfers are refunded but patron payments are not.


zburke:
Replaced by #1866 due to [FOLIO-3290](https://issues.folio.org/browse/FOLIO-3290)

PR #1874, opened 2021-09-22
UIU-2187: add POC fix for patron block accordion open/closed state

pavelspichonak:
## Purpose
POC fix for patron block accordion open/closed state
**Do not merge**

## Refs
https://issues.folio.org/browse/UIU-2187

### DIFF
--- a/src/components/UserDetailSections/PatronBlock/PatronBlock.js
+++ b/src/components/UserDetailSections/PatronBlock/PatronBlock.js
@@ -5,7 +5,6 @@ import {
   FormattedMessage,
   injectIntl,
 } from 'react-intl';
-import moment from 'moment';
 
 import {
   Row,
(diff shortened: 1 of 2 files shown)


src/components/UserDetailSections/PatronBlock/PatronBlock.js
```diff
@@ -29,32 +28,18 @@ const PATRON_BLOCKS_COLUMNS = {
 };
 
 class PatronBlock extends React.Component {
-  static manifest = Object.freeze({
```
pavelspichonak, line 32: Looks like it is not used at all. It is not connected to `stripes-connect`.

src/components/UserDetailSections/PatronBlock/PatronBlock.js
```diff
@@ -98,33 +71,35 @@ class PatronBlock extends React.Component {
...
-      .then(records => {
-        const blocks = records.filter(p => moment(moment(p.expirationDate).endOf('day')).isSameOrAfter(moment().endOf('day')));
-        if ((blocks.length > 0 && !expanded) || (!blocks.length && expanded)) {
-          onToggle({ id: accordionId });
-        }
-      });
+    const { isToggledOnBlocksLoaded } = this.state;
+
+    if (!isToggledOnBlocksLoaded && arePatronBlocksLoaded) {
+      if (!expanded && patronBlocks.length) {
```
pavelspichonak, line 85: If we should not toggle patron blocks if user already clicked "**Expand all**" button we have to pass additional props here which says that user clicked "**Expand all**" button and we have to prevent toggling in this place.

src/components/UserDetailSections/PatronBlock/PatronBlock.js
```diff
@@ -29,50 +28,23 @@ const PATRON_BLOCKS_COLUMNS = {
...
 class PatronBlock extends React.Component {
-  static manifest = Object.freeze({
-    manualPatronBlocks: {
-      type: 'okapi',
-      records: 'manualblocks',
-      path: 'manualblocks',
-      accumulate: 'true',
-      fetch: false,
-      DELETE: {
-        path: 'manualblocks/%{activeRecord.blockId}',
```
Dmitriy-Litvinenko, line 40: Could you please double check can we remove delete here?

Dmitriy-Litvinenko (changes requested):

PR #1876, opened 2021-09-23
UIU-2250: Increment stripes to v7

mkuklis:
https://issues.folio.org/browse/UIU-2250 

Increment stripes to v7


zburke (approved):
Thank you!

PR #1889, opened 2021-10-04
UIU-1996: User Information page 'Patron block' accordion should be op…

pavelspichonak:
## Purpose
Fix issue User Information page 'Patron block' accordion should be open if patron block exists and closed if not

## Refs
https://issues.folio.org/browse/UIU-1996

## Notes
While fixing the issue one more issue was found.
When you click on manual patron block in the patron block list

Screen Shot 2021-10-04 at 1 39 41 PM

patron block page is opened with this patron block data

Screen Shot 2021-10-04 at 1 39 57 PM

if you refresh the page, patron block data disappear

Screen Shot 2021-10-04 at 1 40 09 PM

This issue is fixed in this MR as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,7 @@
 * Create Jest/RTL test for `contactTypes`. Refs UIU-2368.
 * Create Jest/RTL test for `withMarkAsMissing`. Refs UIU-2265.
 * Create Jest/RTL test for `ActionsBar`. Refs UIU-2301.
+* Fix issue User Information page 'Patron block' accordion should be open if patron block exists and closed if not. Refs UIU-1996.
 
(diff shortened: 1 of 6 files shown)


Stanislau-Snapok (approved):

Dmitriy-Litvinenko (approved):

Dmitriy-Litvinenko:

mkuklis:
@pavelspichonak this looks good. Would it be possible for you to add coverage for what was fixed in UIU-1996 ('Patron block' accordion should be open if patron block exists and closed if not).

pavelspichonak:
> @pavelspichonak this looks good. Would it be possible for you to add coverage for what was fixed in UIU-1996 ('Patron block' accordion should be open if patron block exists and closed if not).

Yeah sure, I've added unit tests to check if `expanded` prop is true or false based on loaded patrons blocks amount.
Could you please review again?

PR #1894, opened 2021-10-05
UIU-2179 replace react-intl-safe-html with react-intl

kuplink:
Refactor away from `react-intl-safe-html` since its features
are now supported natively by `react-intl`.

Refs UIU-2179


zburke:
@ankitasen-ubmainz, this is a test of [FOLIO-3290](https://issues.folio.org/browse/FOLIO-3290).

PR #1613, opened 2021-01-21
UIU-1616 Claim returned-5: Display billed fees/fines as Suspended 

ymacedor:
Claim returned-5: Display billed fees/fines as Suspended on Open/Closed Loans and Loan Details when item claimed returned
Need to identify fees/fines on loans that have been suspended due to the patron claiming the item has been returned.

PR #1904, opened 2021-10-07
UIU-2446 correctly show fee/fine to unprivileged user

zburke:
This was subtle. In JSX, `<>{value}</>` returns a component, but in
plain ol' JS, `{value}` returns an object. Here, we need either a
component or a primitive, but we got an object. Whoops.

Introduced in #1613.

Refs [UIU-2446](https://issues.folio.org/browse/UIU-2446)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change history for ui-users
 
+## 7.something (IN PROGRESS)
+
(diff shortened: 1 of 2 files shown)


Baroquem (approved):

mkuklis (approved):
Nice catch!

PR #1915, opened 2021-10-18
UIU-2284 create Jest/RTL test for ModalContent.js

Mathansivavelmurugesan:
Code coverage and unit testing for the issue : https://issues.folio.org/browse/UIU-2284

UIU-2284


mkuklis (approved):
@Mathansivavelmurugesan this looks great. It would be good to update Changelog and cleanup elsint issues.

PR #1923, opened 2021-10-28
disable BigTest workflows and reports

zburke:
We don't run the BigTest workflows any longer so there is no value in
reporting the results of the test we didn't run. :laugh:


zburke:
@ankitasen-ubmainz , any objections, any reason not to merge similar PRs in other repos where we have completely disabled BigTest?

ankitasen-ubmainz (approved):

PR #1958, opened 2021-11-30
UIU-2408 create Jest/RTL test for Owners.js

Mathansivavelmurugesan:
Code coverage and unit testing for Owners.js


mkuklis (approved):

zburke (changes requested):
@Mathansivavelmurugesan , I see that you're passing different `dataOptions` values in each of these tests, but the `expect` function is identical for all of them, which surprised me. e.g. in the first test, shouldn't we check that `test1256` is _not_ rendered since `filterShared` is true? 

@zburke , My bad checked those with debug but didnt add the cases, now I have added those scenarios.

zburke (approved):

mkuklis (approved):

PR #1960, opened 2021-12-02
UIU-2370 create Jest/RTL test for PatronBlockModalWithOverrideModal.js

Mathansivavelmurugesan:
Code coverage and unit test cases for PatronBlockModalWithOverrideModal.


src/components/PatronBlock/PatronBlockModalWithOverrideModal.test.js
```diff
@@ -0,0 +1,56 @@
...
+import '__mock__/stripesCore.mock';
+import '__mock__/stripesSmartComponent.mock';
+
+jest.unmock('@folio/stripes/components');
+
+const renderPatronBlockModalWithOverrideModal = (props) => renderWithRouter(<PatronBlockModalWithOverrideModal {...props} />);
+
+describe('UserPermissions component', () => {
+  it('Checking patronBlock Modal', () => {
+    const props = {
```
mkuklis, line 15: @Mathansivavelmurugesan it looks like the same structure is repeated in other tests. Would it be possible for you to move it maybe maybe to `beforeEach` and just reuse it between tests?

src/components/PatronBlock/PatronBlockModalWithOverrideModal.test.js
```diff
@@ -0,0 +1,56 @@
...
+      onRenew: jest.fn(),
+      onOpenPatronBlockedModal: jest.fn(),
+      onClosePatronBlockedModal: jest.fn(),
+      patronBlockedModalOpen: true,
+      patronBlocks: [],
+      viewUserPath: '/Users/Test'
+    };
+    renderPatronBlockModalWithOverrideModal(props);
+    fireEvent.click(screen.getByText('ui-users.blocks.overrideButton'));
+    fireEvent.click(document.querySelector('[data-test-override-patron-block-modal-close="true"]'));
```
zburke, line 36: Note that [fireEvent only captures a small piece of what happens](https://testing-library.com/docs/guide-events/)  during, say, a click event. So, there's a tradeoff: it's likely faster than modeling the whole event, but it's also not very representative of a real user interaction. As noted in the [fireEvent API doc](https://testing-library.com/docs/dom-testing-library/api-events/), consider using [@testing-library/user-event](https://testing-library.com/docs/ecosystem-user-event) unless you have a specific reason not to. 

If you're following an existing pattern, there's no need to change this now, though if there's a specific reason you chose `fireEvent` over `userEvent` it would be good to leave a comment explaining why.

zburke (approved):

mkuklis (approved):

PR #1967, opened 2021-12-08
UIU-2466: Reset offset when sort values change

mkuklis:
Sorting was not working correctly after "Load more" option was used to fetch more data. This PR resets the offset when sort is used so only the first sorted page is loaded. 

https://issues.folio.org/browse/UIU-2466


zburke (approved):
How I love seeing a comment explaining why a particular has parachuted into the codebase out of the blue! 😍

PR #1969, opened 2021-12-09
UIU-1995: 'Created at' on Fee/Fine Details should be a '-' 

ymacedor:
[UIU-1995](https://issues.folio.org/browse/UIU-1995): 'Created at' on Fee/Fine Details should be a '-'  **if it doesn't exist**
imagen


src/components/util/util.js
```diff
@@ -43,10 +43,11 @@ export const formatDateAndTime = (date, formatter) => {
   return date ? formatter(date, { day: 'numeric', month: 'numeric', year: 'numeric' }) : '';
 };
 
+
 export const getServicePointOfCurrentAction = (action, servicePoints = []) => {
   const servicePoint = servicePoints.find(sp => sp.id === action.createdAt);
-
-  return servicePoint ? servicePoint.name : action.createdAt;
+  const createAtValue = servicePoint ? servicePoint.name : action.createdAt;
+  return createAtValue || '-';
```
zburke, line 50: Please use [`<NoValue>`](https://github.com/folio-org/stripes-components/tree/master/lib/NoValue) instead of `-` for accessibility reasons.

zburke (changes requested):

src/components/util/util.js
ymacedor: @zburke the change has been done.

zburke:
Replaced by #1983

PR #1970, opened 2021-12-10
UIU-2482 Update layout of Fee/Fine Details

ccruz17:
### Story:
https://issues.folio.org/browse/UIU-2482

### Result: 
Captura de Pantalla 2021-12-10 a la(s) 3 19 47


src/views/AccountDetails/AccountDetails.js
```diff
@@ -568,28 +570,10 @@ class AccountDetails extends React.Component {
...
-            >
+            <Col xs={1.5} sm={3} md={3} lg={3}>
               <KeyValue
-                label={<FormattedMessage id="ui-users.loans.details.lostItemPolicy" />}
-                value={lostItemPolicyId
-                  ? <Link to={`/settings/circulation/lost-item-fee-policy/${lostItemPolicyId}`}>{lostItemPolicyName}</Link>
-                  : '-'
-                }
+                label={<FormattedMessage id="ui-users.feefines.details.dateClose" />}
+                value="-"
```
zburke, line 576: Please use [`<NoValue>`](https://github.com/folio-org/stripes-components/tree/master/lib/NoValue) instead of `-`; it's better for accessibility.

zburke: The screenshots attached to the story imply the "Date fee/fine closed" field should display a timestamp but that isn't the case here. Shouldn't `value` be a ternary, like the others, e.g. something like
```
value={loan.closedDate ? <FormattedTime value={loan.closedDate} /> : <NoValue />}
```

src/views/AccountDetails/AccountDetails.js
```diff
@@ -670,6 +654,48 @@ class AccountDetails extends React.Component {
               />
             </Col>
           </Row>
+          <Row>
+            <Col xs={1.5} sm={1.5} md={1.5} lg={1.5}>
```
zburke, line 658: Since `xs` is the default and things scale up for there, the other `<Col>` styles (`sm`, `md`, `lg`) can be omitted when they match.

zburke (changes requested):
This is off to a good start but doesn't appear to implement all the changes requested in the ticket.

src/views/AccountDetails/AccountDetails.js
ccruz17: Apply the recommendations or fixes in my new commit

src/views/AccountDetails/AccountDetails.js
ccruz17: Apply the recommendations or fixes in my new commit

ccruz17:
@zburke, I already applied the recommendations or corrections in my new commit

zburke:
Now that I'm looking closer at the "Date fee/fine closed" field, it isn't clear what value that should correspond to. The field used here does not exist. On the one hand, that's harmless and will just always display as `<NoValue>`; on the other, it's clearly wrong. 

I'll merge this as-is since all the other changes look perfect, but the Jira ticket should remain open until you hear from the PO what value should actually be in that field.

zburke:
Replaced by #1986

PR #1972, opened 2021-12-12
UIU-2409 confirm editing of 'visible: false' permissions

zburke:
The permissions dialog only shows permissions with `visible: true`
attributes, but sometimes permissions with `visible: false` are assigned
to users. Editing such a user or permission set will cause the `visible:
false` values to be removed (see [UIU-2410](https://issues.folio.org/browse/UIU-2410)).

Show a prompt in this situation explaining the situation so the user can
choose whether to continue with the edit action, losing the hidden
permissions, or cancel, leaving things as they are.

Refs [UIU-2409](https://issues.folio.org/browse/UIU-2409)


mkuklis (approved):
Nice and clean!

Baroquem (approved):

MikeTaylor:
What an incredible amount of code just for an "are you sure?" message!

More constructively: I think the message should mention the option of setting "Show invisible permissions" in the Developer settings' Configuration area, and then proceeding.

zburke:
@MikeTaylor: providing a link to the dev-tools setting is a great idea, but it's tricky. ui-users doesn't know anything about ui-developer. We've basically ignored this fact with links among ui-users, ui-requests, and ui-inventory, but somehow the ice feels even thinner here. Related, not everybody likely has permission to access ui-developer, but managing access with `<IfPermission>` won't work since those permissions aren't defined in this module.

MikeTaylor:
Interesting, @zburke. Everything you say makes sense. And yet, and yet ... Bottom line, were I in that position, I would want to know about the one option that would let me actually do what I wanted.

PR #1985, opened 2021-12-16
UIU-1462: Add Staff Info: Change "Created at" to Service Point

ymacedor:
Refs.[ UIU-1462](https://issues.folio.org/browse/UIU-1462)
imagen


zburke:
Replaced by #1999

PR #1989, opened 2021-12-18
UIU-2488 use supported uuid version

zburke:
Use a supported `uuid` release that doesn't cause build warnings.

Refs [UIU-2488](https://issues.folio.org/browse/UIU-2488)


mkuklis:
Since you are doing this cleanup maybe it would be valuable to switch to nanoid: https://github.com/ai/nanoid

zburke:
@mkuklis, `nanoid` is a nice idea, but UUID is baked into the backend modules so a POST request with a non-conformant `id` would be rejected. Alas.

mkuklis:
That makes sense @zburke! I didn't realize that their format is actually different. I thought nanoid is just faster.

PR #1991, opened 2021-12-21
UIU-2490: Do not push to history if the url didn't change.

mkuklis:
https://issues.folio.org/browse/UIU-2490

Only use `history.push` if the url is actually different from the current one.

I noticed this bug when working on lazy loading. The ui-users would just go into infinitive loop and keep re-rendering `UserSearchContainer` reloading all resources. This doesn't happen when everything is bundled into one single module in webpack. It almost seems like a timing issue.


zburke (commented):
Nice catch!

PR #1995, opened 2021-12-28
UIU-2406 create Jest/RTL test for OwnerSettings.js

Mathansivavelmurugesan:
Unit test cases and code coverage for owner settings component.


src/settings/OwnerSettings.test.js
```diff
@@ -1,35 +1,196 @@
...
+    renderOwnerSettings(propData);
+    expect(screen.getByText('ui-users.owners.columns.desc')).toBeTruthy();
+    expect(screen.getByText('ui-users.owners.columns.asp')).toBeTruthy();
+  });
+  it('Delete functionality in component', () => {
+    act(() => {
+      renderOwnerSettings(propData);
+    });
+    act(() => {
+      fireEvent.click(document.querySelector('[id="clickable-delete-settings-owners-1"]'));
```
mkuklis, line 171: @Mathansivavelmurugesan could you please replace `fireEvent` with `click` from `https://testing-library.com/docs/ecosystem-user-event` (that lib should already in place in ui-users). That should allow you to remove all these `act` wrappers.

mkuklis: Also here is a good resource about the `act` and the warnings we sometimes see and how to approach it:

https://kentcdodds.com/blog/fix-the-not-wrapped-in-act-warning

mkuklis (approved):
@Mathansivavelmurugesan nice work 👏 !

PR #1998, opened 2022-01-05
UIU-1160 and UIU-2001 Modify how default Fee/Fine Owner is set

zburke:
Refs [UIU-1160](https://issues.folio.org/browse/UIU-1160) Modify how default 'Fee/Fine Owner' is set on New Fee/Fine page, [UIU-2001](https://issues.folio.org/browse/UIU-2001) Modify how default 'Fee/Fine Owner Desk' is set on Pay Fee/Fine modal

Replaces #1982


zburke:
I thought we could only run SonarCloud checks on PRs from non-forked branches because some secret tokens couldn't be passed through, but now #1982 is showing that all checks have passed. Maybe that's because I also pushed this branch? Dunno. In any case, I'm merging the original PR from the original branch.

PR #2002, opened 2022-01-06
UIU-2256: Fix permissions problems

Baroquem:
https://issues.folio.org/browse/UIU-2256

This is followup to previous work to address two problems that were spotted with the changes to ui-users.loans.all:
1. The "claim returned" action button wasn't protected by an `<IfPermission>`.
2. Renewals fail because of a missing `manualblocks.collection.get` permission (needed, presumably, to see whether there are blocks on the patron before an item can be renewed). I'm not sure why this suddenly turned up as a problem -- maybe we haven't tested with this particular set of limited permissions before.


mkuklis (approved):
👏

PR #2008, opened 2022-01-12
UIU-2473: Properly show service point name in fee/fine details

Baroquem:
https://issues.folio.org/browse/UIU-2473

In order to get the correct service point name, we need to pull in the full list of service points to match against. As Zak stated in the issue comments, this code was previously trying to match based on the user's assigned service points, which could be just a small subset of the full list of SPs where the fine could have been assigned.

Screen Shot 2022-01-12 at 1 32 01 PM


mkuklis (approved):
👏 Nice detective work @Baroquem!

PR #2013, opened 2022-01-13
UIU-2499 suppress edit of some users

zburke:
Suppress edit of users whose IDs are specified in an array stored in mod-configuration. 

Create a configurations entry by sending a POST request to `/configurations/entries` with a body like
```
{
  "module": "@folio/users",
  "configName": "suppressEdit",
  "default": true,
  "enabled": true,
  "value": "[\"SOME_UUID\", \"OTHER_UUID\"]",
}
```
Users whose UUIDs match array values will not have an "Edit" button when viewing the "Actions" menu from the user-details pane. This is a bit kludgey; all it does is remove the "Edit" button from the Action menu, so a determined user could still directly access the edit screen by manually entering a URL like `/users/SOME_UUID/edit`. I could add a check there too if folks think this needs to be more robust. 

Refs [UIU-2499](https://issues.folio.org/browse/UIU-2499)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@
 * Also support `circulation` `12.0`. Refs UIU-2480.
 * Also support `request-storage` `4.0` (TLR). Refs UIU-2495, UIU-2480.
 * Fix problems with permissions for claim returned, renewals. Refs UIU-2256.
+* Suppress edit of users stored in a configuration entry. Refs UIU-2499.
 
 ## [7.0.1](https://github.com/folio-org/ui-users/tree/v7.0.1) (2021-10-07)
 [Full Changelog](https://github.com/folio-org/ui-users/compare/v6.1.0...v7.0.1)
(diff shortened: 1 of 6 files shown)


mkuklis (approved):
This is nice @zburke!
Is this related to what we discussed yesterday?

zburke:
@mkuklis, yes, related to preventing some records from being changed in the community environments. I was trying to figure out "what's the smallest amount of work that would still have a positive impact?" I think this is pretty close :) Of course, I should pull `actionMenuEditButton` into a totally separate component in order to write a test for it so SonarCloud can stop shaming me on this PR.

zburke:
@funkymalc, @wafschneider, @dcrossleyau: this would allow us to prevent `diku_admin` (and any other accounts) from being edited in the ref envs if we could add a curl request to the pipeline as in the description.

Baroquem (approved):

dcrossleyau (approved):
Oh nice. Thanks.

PR #2019, opened 2022-01-24
UIU-2506: Fix unexpected increase of fee/fine remaining balance

Stanislau-Snapok:
## Purpose
Fix unexpected increase of fee/fine remaining balance.

## Approach
`this.props.balnce` was a overall calculated amount for all user fee/fines.
When we use `New staff info`, we update currnet fee/fine, so overall balance is not what we need.
`account.remaining` - is correct field.
I tested new behaviour with all posible actions - everything looks fine. 

## Refs
https://issues.folio.org/browse/UIU-2506

## Screenshots
Before:
image

Аfter:
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@
 * Properly show service point name in fee/fine details. Fixes UIU-2473.
 * Suppress edit of users stored in a configuration entry. Refs UIU-2499.
 * Also support `circulation` `13.0`. Refs UIU-2483.
+* Fix unexpected increase of fee/fine remaining balance. Refs UIU-2506.
 
 ## [7.0.1](https://github.com/folio-org/ui-users/tree/v7.0.1) (2021-10-07)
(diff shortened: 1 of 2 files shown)


Dmitriy-Litvinenko (approved):

zburke (approved):
@Stanislau-Snapok, it's great that such a small fix can address such a serious bug! 👏 👏 👏 

Would you consider adding one or more unit tests to this component? I know it may seem to be beyond the scope of just fixing the bug, but this kind of bug cries out for a unit test to me.

mkuklis (approved):

Dmitriy-Litvinenko:
Hello @zburke  Could you please confirm that we can merge this pull request as is? We will create task for add jest/rtl test for current case and try implement it in current sprint?

pavelspichonak (approved):

zburke:
Yes, @Stanislau-Snapok , you're good to merge. We already have a [ticket to add test coverage to this component](https://issues.folio.org/browse/UIU-2283).

PR #2020, opened 2022-01-24
UIU-2280 create Jest/RTL test for CommentModal.js

Mathansivavelmurugesan:
Code coverage and unit testing for comment Modal.

### DIFF
--- a/src/components/Accounts/Actions/CommentModal.test.js
+++ b/src/components/Accounts/Actions/CommentModal.test.js
@@ -1,0 +1,49 @@
+import { waitFor, screen, cleanup } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+
+import renderWithRouter from 'helpers/renderWithRouter';
+import CommentModal from './CommentModal';
+
+jest.unmock('@folio/stripes/components');
+jest.unmock('@folio/stripes/smart-components');
+
+const renderCommentModal = (props) => renderWithRouter(<CommentModal {...props} />);
+
+const MockFunc = jest.fn();
+
+
+const propData = {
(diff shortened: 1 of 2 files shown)


mkuklis (approved):

src/components/Accounts/Actions/CommentModal.test.js
```diff
@@ -0,0 +1,49 @@
...
+
+import renderWithRouter from 'helpers/renderWithRouter';
+import CommentModal from './CommentModal';
+
+jest.unmock('@folio/stripes/components');
+jest.unmock('@folio/stripes/smart-components');
+
+const renderCommentModal = (props) => renderWithRouter(<CommentModal {...props} />);
+
+const MockFunc = jest.fn();
```
zburke, line 12: Pedantic: `s/MockFunc/mockFunc/`. Only classes should have initial caps.

src/components/Accounts/Actions/CommentModal.test.js
```diff
@@ -0,0 +1,49 @@
...
+
+
+const propData = {
+  form: { reset: MockFunc },
+  pristine: false,
+  submitting: false,
+  onClose: MockFunc,
+  handleSubmit: MockFunc,
```
zburke, line 20: More pedantic: don't re-use `MockFunc` here. Different attributes should receive different values, otherwise, we can't validate that they have been called correctly, i.e. the close-handler is supposed to submit the form and close the modal you don't have a way to test that here.

zburke (changes requested):
The tests themselves look good, but please follow conventions for casing of function names and use separate mocks for separate props.

PR #2023, opened 2022-01-26
UIU-2498: calculation expiration date

elsenhans:
https://issues.folio.org/browse/UIU-2498

**Overview:** 
With UXPROD-2389, a default expiration date can be filled in when a new user is created, a user is changed from inactive to active, re-set or, the case here, a patron group is changed.

The default expiartion period for the patron group is either added to the current expiration date or the current date. 

In the case of a change of patron group, it should be added to the current date.

**Steps to Reproduce:**
1.  Set default expiration date for a patron group X.
2. Open a user with a set expiration date.
3. Change the patron group

**Expected Results:**
The expiration date should be changed to expiration date = today + expiration period.

**Actual Results:**
It is changed to expiration date = current expiration date + expiration period.

**Additional Information:**
 See Uschis comment (https://issues.folio.org/browse/UXPROD-2389?focusedCommentId=116868&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-116868). 

**Case 1: new User**
expiration date = today + expiration period

**Case 2/3: set inactive / set active**
expiration date does not change

**Case 4: Change Patron group**
should not be: expiration date = current expiration date + expiration period
should be like new user: expiration date = today + expiration period

**Case 5: Click Re-set**
expiration date = current expiration date + expiration period 
For Click Re-set see also:
https://issues.folio.org/browse/UIU-2046
(1) Current expiration date is in the past
-> calculate from today
(2) current expiration date is in the future or today
-> calculate from given expiration date

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,12 +24,12 @@
 * Also support `request-storage` `4.0` (TLR). Refs UIU-2495, UIU-2480.
 * Fix problems with permissions for claim returned, renewals. Refs UIU-2256.
 * Settings > Users > change focus. Refs UIU-2036.
-* Correct calcualtion for expiration date. Refs UIU-2498.
 * Add custom fields filters. Refs UIU-2170.
 * Properly show service point name in fee/fine details. Fixes UIU-2473.
 * Suppress edit of users stored in a configuration entry. Refs UIU-2499.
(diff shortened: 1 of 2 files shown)


mkuklis (approved):
@elsenhans nicely done!

PR #2024, opened 2022-01-26
UIU-2075: Preserve invisible permissions during edit

Baroquem:
https://issues.folio.org/browse/UIU-2075

This will keep invisible permissions from being deleted, but note that there's now a conflict with https://issues.folio.org/browse/UIU-2409, which warns users that invisible perms _will_ be deleted!


src/components/PermissionsAccordion/components/PermissionsModal/PermissionsModal.js
```diff
@@ -240,12 +240,19 @@ class PermissionsModal extends React.Component {
...
     } = this.props;
     const permissions = get(this.props, 'resources.availablePermissions.records', []);
     const filteredPermissions = permissions.filter(({ id }) => assignedPermissionIds.includes(id));
 
-    addPermissions(filteredPermissions);
+    // Invisible permissions assigned to a user are lost in the permissions selection process,
+    // so we want to make sure that they get saved along with any visible selections. This renders
+    // it impossible to *remove* invisible permissions in the UI, but that's probably better
+    // than making it impossible to *keep* invisible permissions in the UI.
+    const invisiblePermissions = assignedPermissions.filter(p => !p.visible);
```
zburke, line 253: Thanks for this perfect comment explaining not only what is happening, but why. So often, we fail to document our reasoning behind making a change instead of just saying, "changed from foo to bar". Thank you!

zburke (approved):
WRT conflicting with [UIU-2409](https://issues.folio.org/browse/UIU-2409), as your comment states, I think this behavior is better. I think we should either (1) open a new ticket to undo UIU-2409 (#1972), or (2) decide it's prudent to undo UIU-2409 in the scope of this ticket and just do it right in this PR. I think both are totally acceptable.

Baroquem:
Undoing it in this PR is probably more convenient; at least, disabling the warning modal should be straightforward enough if we don't need to remove it altogether.

mkuklis (approved):
👏

PR #2034, opened 2022-02-04
UIU-2496: Fix problems with permissions duplication in PermissionsModal

Baroquem:
https://issues.folio.org/browse/UIU-2486
https://issues.folio.org/browse/UIU-2496

Most of the bugs and weird behaviors around permissions selection and assignment in `<PermissionsModal>` were the result of double-dipping from the `availablePermissions` resource. I've fixed this by adding a new state variable to hold the complete list of permissions; the existing state `permissions` variable actually changes during searches, which was confusing the issue.

I've also fixed another (unreported?) bug: when the modal was opened for a user who already had assigned permissions, none of the assigned perms were checked/selected. Then, unless you carefully re-selected all of the extant assignments before saving, you'd lose the old ones because they weren't selected in the list. (I hadn't realized at first that that _was_ a bug and thought it was by design, so I found the behavior baffling!)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,7 +34,7 @@
 * Column selector dropdown does not match column headings. Refs UIU-2504.
 * Fee/Fine Type not showing/saving for first manual fee/fine created. Refs UIU-2508.
 * Add Jest/RTL tests for `CommentModal` business logic in `FeeFineActions` component. Refs UIU-2515.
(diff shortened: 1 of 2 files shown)


mkuklis (approved):
@Baroquem I'm so sorry for not getting to this sooner. For some reason it never appeared on my list to review it. 

This is a great approach with the `allPermissions`. 🎉 

I think this will make @zburke very happy too.

PR #2035, opened 2022-02-04
UIU-2492 don't choke when item record is missing

zburke:
Sometimes an item record for an active loan may be inadvertently
deleted. The display should not choke under that circumstance.

Refs [UIU-2492](https://issues.folio.org/browse/UIU-2492)


Dmitriy-Litvinenko (commented):
We should discuss current approach

Dmitriy-Litvinenko (changes requested):
We should discuss current approach

zburke:
This change only addresses a tiny fraction of the NPEs and other bugs that can crop up when a loan's item-record is deleted. I am closing this in favor of taking a broader approach that acknowledges all the associated issues.

PR #2043, opened 2022-02-16
UIU-1697 - Create/Edit Patron Blocks: Required fields are not using the right prop and cannot be read by screenreader

IvanKryzhanovskyi:
## Description
On the create/edit patron block page, these required fields: display description and block actions - should replace the current displayed asterisk with a red asterisk using the prop that supports the indication of a required field

## Approach
- add 'required' and 'aria-required="true"' attributes to display description field
- remove validation messages for 'Block actions' section

## Refs
- https://issues.folio.org/browse/UIU-1697

## Screenshot
after


src/components/PatronBlock/PatronBlockForm.js
```diff
@@ -41,7 +43,7 @@ const showValidationErrors = ({
   expirationDate,
 }) => {
   const errors = {};
-  const patronBlockError = <FormattedMessage id="ui-users.blocks.form.validate.any" />;
+  const patronBlockError = ' ';
```
mariia-aloshyna, line 46: Is this variable still needed?

OleksandrHladchenko1 (approved):

PR #2047, opened 2022-02-17
UIU-2262 create Jest/RTL test for CreateResetPasswordControl.js

Mathansivavelmurugesan:
Code coverage and unit testing for CreateResetPasswordControl Component.


mkuklis (approved):

src/components/EditSections/EditExtendedInfo/CreateResetPasswordControl/CreateResetPasswordControl.test.js
```diff
@@ -0,0 +1,54 @@
...
+    expect(screen.getByText('testemail@email.com')).toBeInTheDocument();
+  });
+  it('If it redirects after POST fails', async () => {
+    const mockFunc = jest.fn(() => new Promise((_, reject) => {
+      const error = 'Something Went Wrong';
+      reject(error);
+    }));
+    await waitFor(() => renderCreateResetPasswordControl(propData(mockFunc)));
+    await userEvent.click(screen.getByText('ui-users.extended.sendResetPassword'));
+    await screen.debug();
```
zburke, line 52: It's odd to me that there is no assertion in this last block. Is there nothing visible in the UI to indicate the POST request fails? If not, then I think we should comment this test for now and file a new UIU ticket to surface the error message.

src/components/EditSections/EditExtendedInfo/CreateResetPasswordControl/CreateResetPasswordControl.test.js
```diff
@@ -0,0 +1,54 @@
...
+    renderCreateResetPasswordControl(propData(mockFunc));
+    expect(screen.getByText('ui-users.extended.sendResetPassword')).toBeInTheDocument();
+  });
+  it('If reset password renders on click', async () => {
+    const mockFunc = jest.fn(() => new Promise((resolve, _) => {
+      resolve({ ok: true, link: 'bl-users/password-reset/link' });
+    }));
+    await waitFor(() => renderCreateResetPasswordControl(propData(mockFunc)));
+    await userEvent.click(screen.getByText('ui-users.extended.sendResetPassword'));
+    expect(screen.getByText('testemail@email.com')).toBeInTheDocument();
```
zburke, line 43: Are you confident this assertion is only valid after the click? Given the user-edit form is populated, I would expect the email address to be in the document both before and after clicking the button. Is there a different/better value we can assert on, i.e. one that is dependent on the click?

zburke (commented):
It isn't clear to me that these tests are specific enough.

src/components/EditSections/EditExtendedInfo/CreateResetPasswordControl/CreateResetPasswordControl.test.js
Mathansivavelmurugesan: Changed the assertion to copylink which would be rendered only on the click

PR #2055, opened 2022-02-23
UIU-1161: Newly created address record should be in focus

Javokhir12:
## Purpose
- Newly created `address` record should be in focus
- Refs: https://issues.folio.org/browse/UIU-1161

## Approach
- Add `autoFocus` prop to `Select` component to move focus on it on `mount`


mkuklis (approved):

zburke:
@JohnC-80 , any thoughts on the use of `autofocus` here?

zburke:
@Javokhir12, the problem is that tests have become flaky in many repos over the past few weeks and we do not understand the reasons why. Most likely, it is _not_ a problem with any of the work done here. I'll try re-running the master-branch build to see if we have better luck the second time.

PR #2058, opened 2022-03-01
UIU-2337 create Jest/RTL test for PatronBlock.js

Mathansivavelmurugesan:
Code coverage and unit testing for PatronBlock component


src/components/UserDetailSections/PatronBlock/PatronBlock.test.js
```diff
@@ -0,0 +1,64 @@
...
+  beforeEach(() => {
+    renderPatronBlock(props);
+  });
+  it('it must be rendered', () => {
+    expect(screen.getByText('ui-users.settings.patronBlocks')).toBeInTheDocument();
+  });
+  it('Checking if it navigates to patronBlock', () => {
+    userEvent.click(document.querySelector('[data-row-inner="0"]'));
+    expect(screen.getByText('Manual')).toBeInTheDocument();
+  });
```
zburke, line 59: It isn't immediately obvious to me how the action and the assertion are connected. Can you add a brief explanation about what element you're clicking, and how the document changes to display the text in the assertion?

src/components/UserDetailSections/PatronBlock/PatronBlock.test.js
```diff
@@ -0,0 +1,64 @@
...
+  it('checking for sort order', () => {
+    userEvent.click(document.querySelector('[id="clickable-list-column-blockedactions"]'));
+    expect(screen.getByText('Sample')).toBeInTheDocument();
```
zburke, line 62: Same as above. Additionally, do you think it's expected that changing the sort (I think that's what you mean by 'checking for sort order') would change the content? I would expect the content to remain the same, but the order to change, and thus we would need to check for position in the document, not just presence in it.

zburke (commented):
I need more context (maybe just comments) to be certain the assertions are actually testing the results of the actions rather than properties of the document that were always true.

src/components/UserDetailSections/PatronBlock/PatronBlock.test.js
Mathansivavelmurugesan: Changed the assertion to verify the router function being called

src/components/UserDetailSections/PatronBlock/PatronBlock.test.js
Mathansivavelmurugesan: Created a issue UIU-2358 for this and added in the test file as comment

zburke (approved):

mkuklis (approved):

PR #2060, opened 2022-03-01
UIU-2290 create Jest/RTL test for patronBlocks.js

Mathansivavelmurugesan:
Code coverage and Unit testing for patronBlocks component.

### DIFF
--- a/src/components/util/patronBlocks.test.js
+++ b/src/components/util/patronBlocks.test.js
@@ -1,0 +1,53 @@
+import getRenewalPatronBlocksFromPatronBlocks from './patronBlocks';
+
+const patronBlocks = [{
+  borrowing: true,
+  desc: 'Sample',
+  id: 'f1e0d3e2-fa48-4a82-b371-bea4e44178ab',
+  patronMessage: '',
+  renewals: true,
+  requests: true,
+  staffInformation: '',
+  type: 'Manual',
+  userId: 'e6dc87a3-591b-43e0-a768-d3552b44878b',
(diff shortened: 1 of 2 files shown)


src/components/util/patronBlocks.test.js
```diff
@@ -0,0 +1,53 @@
...
+
+describe('patronBlocks Filter component', () => {
+  it('if it filters', () => {
+    const data = getRenewalPatronBlocksFromPatronBlocks(patronBlocks);
+    expect(data).toStrictEqual(patronBlocks);
+  });
+  it('if it filters renewal blocked data', () => {
+    const data = getRenewalPatronBlocksFromPatronBlocks(patronBlocks2);
+    expect(data).toStrictEqual([]);
+  });
```
zburke, line 52: Based on the second test, I would have expected `data` in the first to exclude `Test2`. If I am mistaken, can you add a comment explaining why this is the correct behavior? And if I'm not, then I think you found a bug, and if you agree we should file a new UIU ticket.

Mathansivavelmurugesan: It is checking for the blockedrenewal scenario so it returns both data.

zburke: Ah, OK. It wasn't immediately obvious from the tests that `getRenewalPatronBlocksFromPatronBlocks` checks the truthiness of `renewals` _or_ `blockRenewals`. Man, we really hurt for documentation about our business logic :(

mkuklis (approved):

PR #2064, opened 2022-03-04
UIU-1256: Using enter key should open Service points modal when Add service points button in focus

Javokhir12:
## Purpose
- Edit User Record : Using Enter key should Open Add Service points when focus is on the Add Service points button
- Refs: https://issues.folio.org/browse/UIU-1256

## Approach
- Get ref to `Add service points` button and add it to `allowed controls` list for `Enter` key handler.


mkuklis (approved):

zburke (commented):
Hmmmmm, this works, but the `+Add service points` button isn't the only button with this problem: the `Add` buttons for Proxy/Sponsor, `Add permissions`, `Add address`, and likely others also suffer from it. 

What do you think about an implementation where `UserForm` passes a prop-callback like `addButtonRef` when it invokes `EditContactInfo`, `EditProxy`, ... and `EditServicePoints` and they set the ref? Then,  `allowedControls` in `ignoreEnterKey` can be implemented as 
```
    const allowedControls = [
      this.closeButton.current,
      this.saveButton.current,
      this.cancelButton.current,
      [...buttonRefsAddedByAllThoseOtherComponents],
    ];
```
This would also allow us to avoid coupling `UserForm` to the private implementation detail (`#add-service-point-btn`) of `EditServicePoints`. 

If you don't want to fix all those other buttons in the scope of this PR, that's OK. But implementing it this way would at least allow those buttons to be fixed. 

What do you think?

Javokhir12:
@zburke That is actually good idea. I also thought something like forwarding ref or adding custom prop but it would be too many props so I went this way. I think I can implement in your way so I won't have to fix another issue like this)

Javokhir12:
@zburke Made some changes but I couldn't change `Proxy/Sponsor` and `Contact Info` section buttons because they are rendered by components which is from `stripes-core` and `smart-components`. We need to close this ticket by Monday so maybe we'll do follow-up ticket for changes to stripes modules. What do you say?

Javokhir12:
@zburke Could you please review it?

zburke (approved):
This implementation is great and sets the stage for handling other button clicks (e.g. Proxy, Sponsor, Contact, etc) in the future, even if we're not in a position to do it now because of how they are implemented. 

Sorry it took me so looooooooong to look at this again!

PR #2065, opened 2022-03-11
UIU-2541 warn if user-edit actions fail

zburke:
If a user-edit action such as updating user attributes, changing
permissions, or changing request attributes fails, show an error callout
instead of failing silently.

Refs [UIU-2541](https://issues.folio.org/browse/UIU-2541)


Baroquem (approved):

mkuklis (approved):
@zburke nice cleanup and tests :).

mkuklis:
@zburke I forgot to mention. Would you mind updating the CHANGELOG and adding UIU-2541 there?

PR #2067, opened 2022-03-16
Include perms.users.assign in ui-users.editperms UIU-2549

adamdickmeiss:
Thus, users with ui-users.editperms can continue to grant
any permission to any user - except okapi permissions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,10 @@
 # Change history for ui-users
 
-## [7.2.0] IN PROGRESS
+## [8.0.0] IN PROGRESS
 
 * Show an error toast when saving user-changes fails for any reason. Refs UIU-2541.
 * Unable to edit Manual Patron Blocks. Refs UIU-2548.
+* *BREAKING* Require okapi interface `permissions` `5.5` for permission-assignment permissions. Refs UIU-2549.
 
 ## [7.1.0](https://github.com/folio-org/ui-users/tree/v7.1.0) (2022-03-03)
 [Full Changelog](https://github.com/folio-org/ui-users/compare/v7.0.1...v7.1.0)
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@folio/users",
-  "version": "7.2.0",
+  "version": "8.0.0",
   "description": "User management",
   "repository": "folio-org/ui-users",
   "publishConfig": {
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     "okapiInterfaces": {
       "users": "15.0",
       "configuration": "2.0",
-      "permissions": "5.0",
+      "permissions": "5.5",
       "login": "6.0 7.0",
       "users-bl": "5.0 6.0"
     },
@@ -148,7 +148,9 @@
           "ui-users.viewperms",
           "perms.users.item.put",
           "perms.users.item.post",
-          "perms.users.item.delete"
+          "perms.users.item.delete",
+          "perms.users.assign.immutable",
+          "perms.users.assign.mutable"
         ],
         "visible": true
       },


zburke (changes requested):
Dropping support for an interface requires a major version bump, so `7.2.0` should be `8.0.0` if those permissions didn't exist until `5.5` and are now required. 

Do we want to change the description (here and in [the translation file](https://github.com/folio-org/ui-users/blob/master/translations/ui-users/en.json#L928)) to something like "Can assign and unassign any permission to any user" and/or add a description?

adamdickmeiss:
> Dropping support for an interface requires a major version bump, so `7.2.0` should be `8.0.0` if those permissions didn't exist until `5.5` and are now required.
> 
> Do we want to change the description (here and in [the translation file](https://github.com/folio-org/ui-users/blob/master/translations/ui-users/en.json#L928)) to something like "Can assign and unassign any permission to any user" and/or add a description?

No. Adding a permission is not a major bump.

julianladisch:
I agree with Adam. Semantic versioning is about the exposed API of a module, not about the dependencies of a module that are an internal implementation detail.
However, if front-end developers have different version policies we should bump the version.

adamdickmeiss:
> @adamdickmeiss, this PR drops support for an interface. That is a breaking change. Either change the interface declaration to `5.0 5.5` or bump the version to `8.0`.

No. There is no drop for support. Btw "5.5 5.0" would be the same as "5.0" in Okapi. What we want is to allow the use of a new feature a new facility in mod-permissions.

At this point (before this PR ui-users is interface compatible with mod-permissions and permissions 5.0. It is not after this PR as it uses a permissino name that appeared in interface 5.5.

As an example, mod-permissions also started to offer new parameters offset, limit as an alternative to the deprecated count, start. This also is only available in permissions 5.5. But since the new parameters are still obeyed that was not a major bump either.

Since the module versions really don't mean much to Okapi as long as they are sortable it's fine. I just think that bumping ui-users majorly would also send a bad signal for the Lotus release. So I'd call it 7.2.1

mkuklis:
Let's keep `5.0 5.5` and release `7.1.1`. Would that work for you @adamdickmeiss?

adamdickmeiss:
> Let's keep `5.0 5.5` and release `7.1.1`. Would that work for you @adamdickmeiss?

Did you read that `5.0 5.5` makes no sense?

Here's what I want to achieve: that the permission sets at any given time (that goes for all modules) refers to (sub) permissions
that exist. That's not achieved by saying `5.0`.

7.1.1 is fine.

zburke:
> There is no drop for support

This statement is false. This PR drops support for permissions v5.0. The implication of including this change in a patch or minor release is that it will run in the same environment and present the same API as other 7.1.x releases which only require permissions v5.0. After this PR, the module will require permissions v5.5. 

If leveraging the new permissions available in 5.5 is "not required" but provides additional functionality/increased security/whatever, fine, but then leave the dependency at 5.0.

adamdickmeiss:
> > There is no drop for support
> 
> This statement is false. This PR drops support for permissions v5.0. The implication of including this change in a patch or minor release is that it will run in the same environment and present the same API as other 7.1.x releases which only require permissions v5.0. After this PR, the module will require permissions v5.5.
> 
> If leveraging the new permissions available in 5.5 is "not required" but provides additional functionality/increased security/whatever, fine, but then leave the dependency at 5.0.

The interface is: permission. The version is 5.0 or 5.5. Please read what semantic versioning means. When we in Okapi write: permissions >= 5.5, it means all interfaces 5.5 .. 5.indefinite, that's the requirement. A requirement / condition is not the same as breaking change.

When we refer to something new or use a new facility, that is not a breaking change.

It would be called version 6.0 if the permission(s) were *removed*. But they weren't.

adamdickmeiss:
> Guys, good reasons from both sides. But I wonder -- does it even matter what ui-users version is? I mean, does anything ever depend on UI modules?

Perhaps there are technical reasons why it's good to hint a breaking change on UI. Admitted, for mod-permissions the module was bumped to 6.0.0, so if that's desired to for ui-users I'm fine. The module version is a hint that something big is underway and while the interface was kept, there were side-effects - for bootstrap procedures ensuring the admin user had proper permissions etc.

skoczko:
@adamdickmeiss @zburke @mkuklis We need this fix out ASAP. I understand that the simplest way to get this fix merged is to bump Users version to 8.0 correct? If so, should Adam do it as part of the request or Michal after the merge?

zburke:
If `perms.users.assign.immutable` and `perms.users.assign.mutable` exist in `permissions` `5.0` then the interface should remain at `5.0` and the version can remain as-is. We'll merge this on `#master` and publish a patch release, `7.1.1`. If those permissions are not available until `5.5`, then it's a little messier: we will want to release `8.0.0` from a branch and update the version on `#master` to `8.1.0`. 

Comparing the [5.14.4 and 6.1.1 tags in mod-permissions](https://github.com/semver/semver/issues/502#issuecomment-466501843), which I think corresponds to the difference between interfaces 5.3 and 5.5, we are in the latter situation. Deploying the MD generated from this repo against an okapi instance without `permissions` interface `5.5` will, I believe, result in a 422 "Attempting to add non-existent permissions..." when it tries to instantiate the permission-set. 

I'm sorry if this conversation has felt contentious. I have read quite a bit about semantic versioning which is why I'm being a stickler about it. For a UI module, its Okapi interface dependencies function analogously to its peer-dependencies. They are _not_ internal implementation details (sorry, @julianladisch) and they are _not_ optional (sorry, @adamdickmeiss). Together, they form the part of the public API that describes the external requirements of the module. [Bumping them requires a major release.](https://github.com/semver/semver/issues/502#issuecomment-466501843)**

** I'm not just quoting some random guy on the internet. @ljharb is a [member](https://github.com/orgs/tc39/people?query=ljharb) of [the committee defining and maintaining JS itself](https://tc39.es/).

zburke (approved):

PR #2068, opened 2022-03-16
UIU-2548: fix the issue when unable to edit manual patron blocks

Stanislau-Snapok:
## Purpose
We should be able to edit Manual Patron Blocks.

## Approach
This issue appears when we have a more than 10 active manual patron blocks. Before changes we fetch them all, find neccesary by id and prefill data from responce. But when we have more than 10 records, and our block doesn't appears in them, there are no any prefilled data. I added patron id to request, and raize records limit. It helped.

## Refs
https://issues.folio.org/browse/UIU-2548

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [7.2.0] IN PROGRESS
 
 * Show an error toast when saving user-changes fails for any reason. Refs UIU-2541.
+* Unable to edit Manual Patron Blocks. Refs UIU-2548.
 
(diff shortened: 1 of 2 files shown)


zburke (approved):
Perfect, thank you. It's hard to believe we didn't notice the `query` clause was missing on that manualblocks request until now!

Please squash when you merge. I suspect will be cherry-picking this onto a patch branch for a bugfix release.

mkuklis (approved):

PR #2074, opened 2022-03-18
UIU-2537 remove permissions that are no longer available

adamdickmeiss:
Fix permission names.. mostly due to other modules just renaming them.

Refs [UIU-2537](https://issues.folio.org/browse/UIU-2537)


zburke (commented):
As noted in #2067, dropping support for an interface is a breaking change. If it is necessary to pick this change onto the `b8.0` (Lotus) release branch, then effectively `v8.0.0` was a buggy release that forgot to include this breaking change and we can leave the version on master/in this PR at `8.1.0`. If this is targeted for Morning Glory, however, it should bump the version to `9.0.0`.

package.json
```diff
@@ -480,9 +478,7 @@
         "subPermissions": [
           "patron-block-conditions.collection.get",
           "patron-block-conditions.item.get",
-          "patron-block-conditions.item.post",
-          "patron-block-conditions.item.put",
-          "patron-block-conditions.item.delete"
+          "patron-block-conditions.item.put"
```
zburke, line 481: Without post and delete, the display name now feels incorrect. If this is correct, the `displayName` here and [the translation value](https://github.com/folio-org/ui-users/blob/5d5c67cdb46c7856ffc02fd0004b06b34e00e01c/translations/ui-users/en.json#L950) should be updated as well.

adamdickmeiss:
It's funny that only mod-login cared to bump to the interface to signal things being removed. The other modules didn't. In other words, it is impossible to match the definitions when you look at the interfaces only. That's how bad the situation is - an impossible situation. Had the other modules done the same, there would be more bumps (and more breakage with your definition).

Since mod-login cared to bump the interface properly, how that affects the *module* version is beyond me. Just tell me what you'd like the module version to be. Not my job to decide.

package.json
zburke: Is there a corresponding `okapiInterfaces` change here too? I don't know which interface these permissions are associated with, but it would surprise me if it's `login`, the only change indicated here.

package.json
```diff
@@ -415,11 +413,11 @@
...
-          "transfertypes.collection.get",
-          "transfertypes.item.delete",
-          "transfertypes.item.get",
-          "transfertypes.item.post",
-          "transfertypes.item.put",
+          "transfer-criterias.collection.get",
+          "transfer-criterias.item.delete",
+          "transfer-criterias.item.get",
+          "transfer-criterias.item.post",
+          "transfer-criterias.item.put",
```
zburke, line 420: Is there a corresponding `okapiInterfaces` change here too? I don't know which interface these permissions are associated with, but it would surprise me if it's `login`, the only change indicated here.

zburke (changes requested):
I'm glad to see this cleanup, but I think there is likely a missing translation change here, and I suspect some missing interface changes too.

package.json
adamdickmeiss: Good question. They were removed here: https://github.com/folio-org/mod-patron-blocks/commit/849911db4d0b496347a1aa37a11c368e00469701

package.json
adamdickmeiss: https://github.com/folio-org/mod-feesfines/commit/7905485237f1d647f3f0dcb7fbbc69fa44a3ccf1

At this point feesfines was version 15.0.  At later stage feesfines version was bumped to 15.1, 15.2.. (they should have gone straight to 16.0 as this truly a breaking change).

ui-users allows "16.1 17.0", so the change happened before the interfaces that ui-users currently allow.

package.json
adamdickmeiss: I've updated the display name.

mkuklis (approved):
It looks like 6.0 was added back in so this all looks good. @adamdickmeiss do you know if this should go to Lotus?

Baroquem (approved):

adamdickmeiss:
> It looks like 6.0 was added back in so this all looks good. @adamdickmeiss do you know if this should go to Lotus?

I'm not to judge. For many other modules it makes it into Lotus.

package.json
zburke: I agree "Can view and edit..." is better. Please make this change [in the translation file]([the translation value](https://github.com/folio-org/ui-users/blob/5d5c67cdb46c7856ffc02fd0004b06b34e00e01c/translations/ui-users/en.json#L950) as well since that's the value that is actually displayed in the UI.

zburke (approved):
Thanks for seeing this through, @adamdickmeiss!

PR #2085, opened 2022-03-28
UIU-2569: fix error message appearing

Stanislau-Snapok:
## Purpose
Error message "Enter comment" appears erroneously when entering New Staff Info on Fee/Fine Details.

## Approach
Form was reseting before component fully unmout, that why we was able to see error message. With flag `destroyOnUnregister` we no more need to reset form manualy. This resolve first part of the issue. The second problem that when we type something, and than delete and try to click on cancel, in first turn works validation on blure, button runs away from under our cursor and we should click on it one more time to close the window. Regarding this I think we can remove our validation, because we can't save empty message with or without validation.

## Refs
https://issues.folio.org/browse/UIU-2569


artem-blazhko (approved):

zburke (approved):
I love it when we can fix bugs by removing code! Nice job troubleshooting this. Thank you!

PR #2086, opened 2022-03-28
UIU-2557 - Missing interface dependency: tags

IvanKryzhanovskyi:
## Overview
The ui-users package.json file doesn't declare a dependency on the tags interface.  When opening up the users app, you can clearly see a call being made to GET /tags in the browser tools.  It appears this call is made in order to populate the tags filter drop-down.

## Approach
Missing interface dependency has been added into the okapiInterfaces


zburke (approved):
The outcome (add a missing interface to `package.json`) is correct but the reasoning (look in the network inspector and see the query) is flawed. In this case, ui-users itself [does contain a query](https://github.com/folio-org/ui-users/blob/master/src/routes/UserSearchContainer.js#L122-L131) but could also bleed into an app via stripes-smart-components or a plugin.

PR #2094, opened 2022-04-04
UIU-2553 user details patron block table headers should not be interactive

Mathansivavelmurugesan:
Removed headerClick of patron block MCL to make it non clickable


zburke (changes requested):
Just remove the code instead of commenting it; we have git for source control and history so we don't need to maintain it in the file too.

PR #2098, opened 2022-04-06
pin stripes-cli branch to test it

zburke:


zburke:
This was a test. This was only a test.

PR #2099, opened 2022-04-06
UIU-2563 users manipulating psets need permission to do so

zburke:
Prior to `permissions` `5.5`, having `perms.users.items.put`, `...post`,
or `...delete` was a security vulnerability because it allowed any user
with the ability to assign permissions to assign permissions they did
not already have, e.g. for a user with `A` and `B` to grant `C` to
somebody else. That _is_ a useful feature, but the two should not be
coupled (assign own permissions; assign non-owned permissions).

Here, that same logic is applied to creation of permission sets: users
with the ability to create permission sets need the ability to create
sets with permissions they may not own.

Refs [UIU-2563](https://issues.folio.org/browse/UIU-2563)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 * Missing interface dependency: tags. Fixes UIU-2557.
 * Error message "Enter comment" appears erroneously when entering New Staff Info on Fee/Fine Details. Refs UIU-2569.
 * Edit User Record: Using Enter key should Open Add Service points when focus is on the Add Service points button. Refs UIU-1256.
+* Users manipulating permissions sets need access to all permissions. Refs UIU-2563.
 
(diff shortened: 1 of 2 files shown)


zburke:
The alternative implementation would be two psets: "Settings (Users): Can create, edit and remove _owned_ permission sets" (_without_ `perms.users.assign...`) and "Settings (Users): Can create, edit and remove _all_ permission sets" (_with_ `perms.users.assign...`). We could do the same for "Users: Can assign and unassign permissions to users". 

Would that be better? Without that, in the UI at least, we're losing the granularity these new permissions were designed to provide.

adamdickmeiss (commented):
Immutable not needed if this is set manipulation.

Baroquem (approved):

zburke:
@adamdickmeiss, according to user tests,  `immutable` _is_ necessary in order to edit run-time sets composed of build-time (`immutable`) sets.

PR #2101, opened 2022-04-06
UIU-2570: Fix inaccurate request counts in various open loan views & modals

Baroquem:
https://issues.folio.org/browse/UIU-2570
https://issues.folio.org/browse/UIU-2574

This fixes a couple of related bugs in which the number of open requests on a loaned item would never show more than 1 request. This only required an adjustment to the `limit` parameter for the query.

Screen Shot 2022-04-06 at 4 03 41 PM
Screen Shot 2022-04-06 at 4 03 32 PM
Screen Shot 2022-04-06 at 4 06 18 PM
Screen Shot 2022-04-06 at 4 06 28 PM


zburke (approved):
Perfect! Thank you for removing the bug I inserted in #2033, and for explaining to me the bug I inserted in #2033 😆  (there is no reason to limit the _unknown_ number of requests on an individual item to the number of loans overall; they are unrelated).

PR #2109, opened 2022-04-19
UIU-2279 create Jest/RTL test for withServicePoints.js

Mathansivavelmurugesan:
Code coverage and unit testing for withServicePoints wrapper component.


zburke (changes requested):
The coverage looks great but there are some unused variables lint is complaining about.

mkuklis (approved):

PR #2112, opened 2022-04-26
UIU-2338 create Jest/RTL test for memoize.js

Mathansivavelmurugesan:
Code coverage and unit testing for memoize functional component.


src/components/util/memoize.test.js
```diff
@@ -0,0 +1,20 @@
+import memoize from './memoize';
+
+describe('Memoize functional component', () => {
+  it('if it filters', async () => {
+    const fn = (args) => {
+      return args;
+    };
+    const data = await memoize(fn);
```
mkuklis, line 8: @Mathansivavelmurugesan the `memoize` is not async so I think we can drop the `await` here.

src/components/util/memoize.test.js
```diff
@@ -0,0 +1,20 @@
+import memoize from './memoize';
+
+describe('Memoize functional component', () => {
+  it('if it filters', async () => {
```
mkuklis, line 4: @Mathansivavelmurugesan we don't need async here.

src/components/util/memoize.test.js
```diff
@@ -0,0 +1,20 @@
...
+    const fn = (args) => {
+      return args;
+    };
+    const data = memoize(fn);
+    const result = data('one');
+    expect(result).toBe('one');
+  });
+  it('if it filters', async () => {
```
mkuklis, line 12: and here.

PR #2123, opened 2022-05-06
UIU-1657: Use single formatted message for successfully callout message

Dmitriy-Litvinenko:
## Purpose
Use single formatted message for successfully callout message

## Refs
https://issues.folio.org/browse/UIU-1657


zburke (changes requested):
Much better! `amount` just needs a little more clean up to make sure it's correctly handled as a numeric value. 

It's fine to just use `{amount, number}` here and leave the currency conversion to a later PR since it's outlined separately in [UIU-2026](https://issues.folio.org/browse/UIU-2026); OTOH, since it's only a tad more effort to handle this The Right Way you could just do it now.

Dmitriy-Litvinenko:
@zburke Could you please rereview?

zburke (changes requested):
Yay! Currency values formatted as currencies!!! Thank you!

But my petty complaint remains: use `of` instead of `Of` in `en.json`.

Dmitriy-Litvinenko:
> Yay! Currency values formatted as currencies!!! Thank you!
> 
> But my petty complaint remains: use `of` instead of `Of` in `en.json`.

@zburke This part of original translation ""charge.messageOf": " fee/fine Of "," Could you please clarify can we keep it as is?

zburke:
@Dmitriy-Litvinenko, yep, I understand the original translation contained "Of". It was always an error :)

Dmitriy-Litvinenko:
@zburke  Thank you. Fixed. Could you please rereview?

zburke (approved):
LGTM!

PR #2131, opened 2022-05-11
UIU-2602: Remove unused usersPerGroup manifest from PatronGroupsSettings

mkuklis:
https://issues.folio.org/browse/UIU-2602


Baroquem (approved):
Hooray!

PR #2144, opened 2022-05-20
UIU-2607 use NodeJS 16 in CI boxes

zburke:
Lots of clean up of lots of tests. In classic style, "How did this ever work?" Our working theory is that some things that were warnings under NodeJS v12 have been converted to errors in v16, thus causing new failures. Still, some of these issues do seem like outright errors, e.g. incorrect selector IDs, so it's hard to say for certain. 

Refs [UIU-2607](https://issues.folio.org/browse/UIU-2607)


mkuklis (approved):
@zburke great work!

PR #2146, opened 2022-05-23
UIU-2422 - create Jest/RTL test for NoteViewPage.js

OybekMamatovEPAMSystems:
Issue : [UIU-2422](https://issues.folio.org/browse/UIU-2422)
create Jest/RTL test for NoteViewPage.js


Godlevskyi (approved):

NikitaSedyx:
hi @mkuklis could you please review this one? thank you

NikitaSedyx (approved):

src/views/Notes/NoteViewPage.test.js
```diff
@@ -0,0 +1,79 @@
...
+  history,
+  location : history.location,
+  match: {
+    params: {
+      id: 'id'
+    },
+  }
+};
+
+const renderNoteViewPage = () => renderWithRouter(
```
Dmytro-Melnyshyn, line 38: Do we really need to use `renderWithRouter` instead of `render`?

src/views/Notes/NoteViewPage.test.js
```diff
@@ -0,0 +1,79 @@
...
+
+    expect(getByTestId('onEdit')).toBeInTheDocument();
+  });
+
+  it('should redirect to the edit page', () => {
+    const { getByRole } = renderNoteViewPage();
+
+    fireEvent.click(getByRole('button', { name: 'edit' }));
+
+    expect(history.push).toHaveBeenCalled();
```
Dmytro-Melnyshyn, line 77: This test will pass anyway because the `history.push` mock is already called on line 61, you need to invoke the mockClear for each test.

OybekMamatovEPAMSystems: I tried adding `afterEach(() => history.push.mockClear())` , but in this case this test fails with error message 'Received number of calls: 0'

Godlevskyi: Clearing the mock should be done in beforeEach

PR #2148, opened 2022-05-26
UIU-2423 - create Jest/RTL test for NoteCreatePage.js

OybekMamatovEPAMSystems:
Issue : [UIU-2423](https://issues.folio.org/browse/UIU-2423)
create Jest/RTL test for NoteCreatePage.js


Godlevskyi (approved):

src/views/Notes/NoteCreatePage.test.js
```diff
@@ -0,0 +1,28 @@
...
+    <NoteCreatePage {...props} />
+    Note Create Page
+  </div>
+);
+
+describe('Note Create Page', () => {
+  it('should render Note Create Page', async () => {
+    const { getByText } = renderNoteCreatePage();
+
+    expect(getByText('Note Create Page')).toBeInTheDocument();
```
mkuklis, line 26: @OybekMamatovEPAMSystems it looks like we are testing something which is provided in `renderNoteCreatePage ` above but not really anything related to `<NoteCreatePage />`. I don't think this is needed and we should instead try to test `NoteCreatePage`.

mkuklis (approved):

CHANGELOG.md
```diff
@@ -52,6 +53,7 @@
 * create Jest/RTL test for src/components/Accounts/Filters. Refs UIU-2356.
 * Handle time zone issues in user info date displays. Fixes UIU-2623.
 * User should be directed to the last open page when working in more than one browser. Refs UIU-2605.
+* create Jest/RTL test for NoteCreatePage.js. Refs UIU-2423.
```
NikitaSedyx, line 56: this line is duplicated

src/views/Notes/NoteCreatePage.test.js
```diff
@@ -0,0 +1,40 @@
...
+
+const renderNoteCreateRoute = (props = {
+  history,
+  location
+}) => render(
+  <NoteCreateRoute {...props} />
+);
+
+describe('Note Create Page', () => {
+  it('should render Note Create Page', async () => {
```
NikitaSedyx, line 35: could you please introduce one more test case for scenario when location.state is defined?

NikitaSedyx: also would be nice to test what passed to referredEntityData prop when state defined and not

PR #2172, opened 2022-06-13
UIU 2620: New Fee/Fine modal, Spread out the fields 

ymacedor:
[UIU-2620](https://issues.folio.org/browse/UIU-2620): "New Fee/Fine" modal: Spread out the fields on the page so selection of fee/fine ower and fee/fine type are easier to do


src/components/Accounts/ChargeFeeFine/ChargeFeeFine.js
```diff
@@ -257,9 +257,9 @@ class ChargeFeeFine extends React.Component {
     this.setState({ ownerId });
   }
 
-  onChangeFeeFine(e) {
+  onChangeFeeFine(feeFineId) {
```
mkuklis, line 260: @ymacedor the `onChagenFeeFine` is attached to `onChange` event:

https://github.com/folio-org/ui-users/blob/d48a3bc02fd5a5e32395544fe0bc801c890a973a/src/components/Accounts/ChargeFeeFine/FeeFineInfo.js#L97

which passes an event object. Is there a reason you changed it here to `feeFineId`?

mkuklis (changes requested):

mkuklis:
@ymacedor it looks like one of the tests started failing after this change:

https://github.com/folio-org/ui-users/runs/7101624308?check_suite_focus=true#step:12:2838

Could you please make sure it passes? Otherwise we won't be able to merge this.

ymacedor:
@mkuklis  I have already fixed the tests

mkuklis:
@ymacedor thank you. I will try to merge this in: https://github.com/folio-org/ui-users/pull/2172 soon.

You should also now have access to `folio-org/ui-users` so in the future please just clone that repo and open PRs from that remote (instead of cloning ui-users). That should simplify merging your PRs.

PR #2191, opened 2022-07-01
UIU-2631 accurately count loans > 1000

zburke:
The open-loan count on the "Loans" accordion could be inaccurate when
there were more than 1000 loans because it used a `limit=1` query
parameter (which still allows for an [estimated `totalRecords` value](https://github.com/folio-org/raml-module-builder#estimated-totalrecords))
instead of `limit=0` which forces an accurate count.

Display the badge value on the closed accordion with `<FormattedNumber>`
so it will be correctly formatted when the value is > 999.

Refs [UIU-2631](https://issues.folio.org/browse/UIU-2631)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 8.2.0 IN PROGRESS
 
 * Spread out the fields on the new Fee/Fine modal. Fixes UIU-2620.
+* Accurately count open-loans when there are > 1000. Refs UIU-2631.
 
 ## [8.1.0](https://github.com/folio-org/ui-users/tree/v8.1.0) (2022-06-27)
 [Full Changelog](https://github.com/folio-org/ui-users/compare/v8.0.0...v8.1.0)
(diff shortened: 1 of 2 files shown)


zburke:
@mkuklis, there are six queries to `circulation/loans` on the patron-details pane:
```
circulation/loans?query=(userId==USER_ID) sortby id&limit=2000
circulation/loans?query=(userId==USER_ID)&limit=1000
circulation/loans?limit=1&query=userId==USER_ID and status.name<>Closed and action==claimedReturned
circulation/loans?query=(userId==USER_ID)&limit=1000
circulation/loans?limit=1&query=(userId==USER_ID and status.name<>Closed)
circulation/loans?limit=1&query=userId==USER_ID and status.name==Closed
```
and one on the open-loans page when you click the "N open loans" link from the "Loans" accordion:
```
circulation/loans?query=(userId==USER_ID) sortby id&limit=100000
```

This PR only addresses the inaccuracy of those `limit=1` queries, but since there is no loan data displayed on the user-details pane, it seems like we ought to be able remove those `limit=1000` and `limit=2000` queries if possible. Do you want me to file a separate issue for that?

mkuklis (approved):

PR #2193, opened 2022-07-06
UIU-2505: Fine detail page throws an error if item is missing

Dmitriy-Litvinenko:
## Purpose
Fine detail page throws an error if item is missing

## Refs
https://issues.folio.org/browse/UIU-2505


zburke (approved):
LGTM

PR #2198, opened 2022-07-18
UIU-2645: Confirm button is not disabled after the click and each click on this button sends the request to the server

artem-blazhko:
## Purpose

Initially "Confirm" button that could be found on Declare lost modal window could be clickable a few times. It happened due to incorrect implementation of "disable" functionality of the button.
In the scope of this PR that issue should be fixed.

## Refs

[UIU-2645](https://issues.folio.org/browse/UIU-2645)


zburke (approved):
This looks good to me, but the `<LoanActionDialog>` API that accepts both `disableButton` and `enableButton` props seems fraught to me. If both functions return true, who wins? I realize these aren't new props in this PR, but if you're tasked with working in this area again it would be worth refactoring this to a single `toggleButton(boolean)` prop, or something else, that removes this potential ambiguity.

artem-blazhko:
@zburke thank you for your comments. Could you take a look one more time?

Dmitriy-Litvinenko:
Hello @zburke @mkuklis Could you please review?

mkuklis (approved):

PR #2201, opened 2022-07-20
UIU-2647 Support notes interface v3.0

BogdanDenis:
`mod-notes` module was updated with breaking changes to `/note-types` endpoint and updated interface version to `3.0`
However, these changes don't affect `/notes` endpoint so `ui-users` can safely use either `2.0` or `3.0`
@zburke should I add anyone else to reviewers?


zburke (approved):

zburke:
@BogdanDenis, generally I include the Dev Lead/Contact from the [team module responsibility matrix](https://wiki.folio.org/display/REL/Team+vs+module+responsibility+matrix) when I'm opening a PR in a repo where I don't regularly do work. Here, that would be @mkuklis. I think this PR is straightforward and OK to merge now, but if you want to wait for his OK then go ahead and add him as a reviewer.

PR #2203, opened 2022-07-26
UIU-2645: Confirm button is not disabled after the click and each click on this button sends the request to the server

artem-blazhko:
## Purpose
Return back `defaultProps` because in some cases components do not receive `toggleButton` prop.

## Refs
[UIU-2645](https://issues.folio.org/browse/UIU-2645)

Related PR - https://github.com/folio-org/ui-users/pull/2198

### DIFF
--- a/src/components/ModalContent/ModalContent.js
+++ b/src/components/ModalContent/ModalContent.js
@@ -89,13 +89,17 @@ class ModalContent extends React.Component {
     onClose: PropTypes.func.isRequired,
     handleError: PropTypes.func.isRequired,
     declarationInProgress: PropTypes.bool.isRequired,
-    toggleButton: PropTypes.func.isRequired,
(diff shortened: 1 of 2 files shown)


src/components/ModalContent/ModalContent.js
```diff
@@ -98,6 +98,10 @@ class ModalContent extends React.Component {
     okapi: PropTypes.object
   };
 
+  static defaultProps = {
+    toggleButton: () => {},
```
ViktorSoroka07, line 102: You can use the `noop` function from `lodash` which does the same.
```suggestion
    toggleButton: noop,
```

ViktorSoroka07 (approved):

src/components/ModalContent/ModalContent.js
artem-blazhko: fixed

PR #2206, opened 2022-07-28
UIU-2610 treat empty-string values as null in name-format

zburke:
When formatting a name object where some values could be present but
empty, treat such values as empty. Previously, the formatter only
checked if values were null or undefined; empty-string was ignored. This
led objects like
```
{
  firstName: "Andrea",
  preferredFirstName: "",
  lastName: "Andromeda"
}
```
to be formatted as "Andromeda" instead of "Andromeda, Andrea" because
`preferredFirstName` was non-null. Now, undefined, null, and empty
string will all be treated as falsy.

Refs [UIU-2610](https://issues.folio.org/browse/UIU-2610)


zburke:
@Mathansivavelmurugesan, can you review this PR? Michał is out this week. Thank you!

Mathansivavelmurugesan (approved):

PR #2215, opened 2022-08-12
UIU-2560 Modify error messages for duplicated fee/fine

artem-blazhko:
## Purpose
When a Fee/Fine Owner enters a Fee/Fine Type that is duplicated by "Shared" we should show a correct message: "_Fee/fine type exists for Shared Fee/fine owner_".
Previously we had a hardcoded message that was used for different cases.

## Approach
Change the existing message and add a new one.
Get rid of unnecessary magic words and use an appropriate constant.

## Refs
[UIU-2560](https://issues.folio.org/browse/UIU-2560)

## Screenshots
image

image


Dmitriy-Litvinenko (approved):

Dmitriy-Litvinenko (approved):

zburke (approved):
Thanks for cleaning up the `Shared` use and replacing it with a constant. That magic word still really bothers me, but the UI code is much cleaner now.

PR #2216, opened 2022-08-17
Fix failed test

artem-blazhko:
## Purpose
Due to the incorrectly implemented mocked date, we have failed test after 2022-08-16.
2022-08-16 date was hardcoded and if the test is executed after this date - it fails.
The mocked date should always be in the future.

## Approach
Next day date was used for mocks. 
As a result, we have an expiration date which is always the next day.

## Refs
[UIU-2668](https://issues.folio.org/browse/UIU-2668)

### DIFF
--- a/src/components/EditSections/EditUserInfo/EditUserInfo.test.js
+++ b/src/components/EditSections/EditUserInfo/EditUserInfo.test.js
@@ -47,7 +47,12 @@ const props = {
   form: {
     change: changeMock,
     getFieldState: jest.fn(() => {
-      return { value: '2100-08-16T02:01:15.606+00:00' };
+      const dateIncrement = 1;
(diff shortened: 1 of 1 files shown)


Dmitriy-Litvinenko (approved):

zburke (changes requested):
There's not enough info in the PR description for me to understand what the problem was, or how this change resolves it. Looking at the change, I'm guessing we need a date in the future, which `2022-08-16` no longer is. Can we implement this that way then, e.g. with a dynamically-configured always-in-the-future-date via faker or something?

artem-blazhko:
@zburke fixed.
Could you check it one more time?

zburke (approved):
LGTM, thank you!

mkuklis (approved):

PR #2217, opened 2022-08-18
UIU-2560: modify translation keys

artem-blazhko:
## Purpose
Use other translation keys for error messages.
In case when we have en_US locale we will not see any changes related to "feefines.errors.existOther" and "feefines.errors.existShared" error messages. 
It happens because en_US translation file will have the following translations:
_"feefines.errors.existOther": "Fee/fine type exists for **other** Fee/fine owner(s)",
 "feefines.errors.existShared": "Fee/fine type exists for **other** Fee/fine owner(s)",_
 
 In this case we can wait for PR from the translation bot or change translation keys (that is done in this PR).
 It is necessary to do it faster and change translation keys because we need to release new changes to MG.

## Refs
[UIU-2560](https://issues.folio.org/browse/UIU-2560)
[Related PR #2215](https://github.com/folio-org/ui-users/pull/2215)


Dmitriy-Litvinenko (approved):

zburke (changes requested):
Sorry, I don't get what's going on here. Changing a translation only in `en.json` will have no effect on what's rendered in the UI because plain `en` [is not exposed as a locale](https://github.com/folio-org/stripes-core/blob/7eb49199a436470d753d99f2204f06c88e665c6b/src/loginServices.js#L29-L58) in stripes-core. If you're trying to update a value for US-English and can't wait for the translation script to run, then you need to change it in `en_US.json`. 

But even then, _you need to actually change the value_. Here, the keys change but not the values:
```
"feefines.errors.existOther":    "Fee/fine type exists for other Fee/fine owner(s)",
"feefines.errors.existForOther": "Fee/fine type exists for other Fee/fine owner(s)",
```
and 
```
"feefines.errors.existShared":    "Fee/fine type exists for Shared Fee/fine owner",
"feefines.errors.existForShared": "Fee/fine type exists for Shared Fee/fine owner",
```

artem-blazhko:
@zburke I will try to explain what the issue is.

Currently, we have `en.json` file which has the following translations:

`"feefines.errors.existOther": "Fee/fine type exists for other Fee/fine owner(s)"`
`"feefines.errors.existShared": "Fee/fine type exists for Shared Fee/fine owner"`

We did not have translations from the translation bot and as a result `en_US.json` has only `"feefines.errors.existShared"` key with old translation:

`"feefines.errors.existShared": "Fee/fine type exists for other Fee/fine owner(s)"`

`en_US.json` does not contain `"feefines.errors.existOther"` key.

As a result when we have locale `en_US` we load translations from `en_US.json` file (see screenshot below).

image

`en_US.json` file has only `"feefines.errors.existShared"` key with old translation which equal `"Fee/fine type exists for other Fee/fine owner(s)"`.
But if `en_US.json` does not have required key, that key with translation will be taken from the default translation file. In our case, it is `en.json`.

As a result, we have the following data in `en_US.json`:

`"feefines.errors.existShared": "Fee/fine type exists for other Fee/fine owner(s)"` - old value that was in `en_US.json`
`"feefines.errors.existOther": "Fee/fine type exists for other Fee/fine owner(s)"` - value that was taken from `en.json`

mkuklis:
@artem-blazhko what @zburke said is correct.

`feefines.errors.existShared` already exists in en_US:

https://github.com/folio-org/ui-users/blob/master/translations/ui-users/en_US.json#L170

You can just add the other one: 

`feefines.errors.existOther` to en_US.json instead of changing the names in en.json.

Please also review the discussion:

https://folio-project.slack.com/archives/C58TABALV/p1660811453591609

artem-blazhko:
@mkuklis `èn_US.json` file has incorrect (old) translation for `feefines.errors.existShared` key.
Also, I am not sure that it is a good idea to change `en_US.json` manually.

cc @Dmitriy-Litvinenko

mkuklis:
@artem-blazhko please review the conversation on slack. It's fine to update en_US.json. You can go ahead and change the translation without changing the keys.

artem-blazhko:
PR was closed because [PR #2218](https://github.com/folio-org/ui-users/pull/2218) was created and merged.

PR #2220, opened 2022-08-23
UIU-2491 Create permission for "Lost items requiring actual costs" processing page

artem-blazhko:
## Purpose
Users with _"Users: Can process lost items requiring actual cost"_ permission should be able to open the _"Lost items requiring actual costs"_ page using the link in action menu (see screenshot below).
If a user doesn't have permission he/she should see the appropriate error page (see screenshot below).
Implementation of _"Lost items requiring actual costs"_ page will be done in the scope of another PR.

## Refs
[UIU-2491](https://issues.folio.org/browse/UIU-2491)

## Screenshots
image

image


Dmitriy-Litvinenko (approved):

zburke (changes requested):
Code LGTM but other permissions are `namedLikeThis` `not-like-this` so you should follow the established convention.

artem-blazhko:
> Code LGTM but other permissions are `namedLikeThis` `not-like-this` so you should follow the established convention.

@zburke thank you for your comment. 
I tried to use the most popular approach in `package.json`  file. I see different approaches there and it seems that `approach-with-dashes` is more popular than `camelCase` approach. 
Could you clarify if there is any documentation with code conventions? 
Thank you in advance.

zburke (approved):
@artem-blazhko, when I reviewed this PR the first time I only looked at the few lines in the diff in `en.json`, and that made this new value look like an outlier. When you back up and look at the whole file, it's clear that I'm wrong about camelCase being the standard. 

In short, ugh, you're right, dash-words is at least as common as any other naming convention in `package.json`. That's definitely unfortunate, but it's also definitely not a problem specifically with this PR. Thanks for setting me straight!

PR #2223, opened 2022-09-01
UIU-2665 Bug: Changing the qindex select updates results

JohnC-80:
See [UIU-2665](https://issues.folio.org/browse/UIU-2665)
Changing the search index dropdown (`qindex`) will currently trigger a search since it mutates the query directly instead of waiting for the user to click the Search button. This is both a useability and accessibility concern, as the user would not expect changing the select dropdown to perform a search.

Until recently, the `SearchField` component from stripes-components did not have the `name` attribute of the `qindex` select exposed.
With that bug fixed, it is possible to wire the `qindex` select up directly with `SearchAndSortQuery`'s SearchChange handlers (since it requires that the changing field have a name attribute)
This also removed the existing `onChangeIndex` method since it was no longer needed.


zburke (changes requested):
Your description of the change makes sense, but lint is unhappy with the solution: 
```
'onChangeSearch' is defined but never used. Allowed unused args must match /^_/u
```
Maybe that's not the right prop to pull out?

JohnC-80:
@zburke you're right. I probably stuck it in there and failed to remove it after I realized the prop I did need to use... all hail lint!

zburke (approved):

PR #2226, opened 2022-09-06
UIU-2236: "Fees/fines incurred" on "Loan History" should show decimal places like it does on "Loan Details"

artem-blazhko:
## Purpose
Add functionality to show decimal places for numbers that do not have value in the decimal place (see screenshot below).
For example instead of `12` we should have `12.00`, instead of `12.2` we should have `12.20`.

## Refs
[UIU-2236](https://issues.folio.org/browse/UIU-2236)

## Screenshots
image


Dmitriy-Litvinenko (approved):

zburke (commented):
Is there a reason not to use [`<FormattedNumber>`](https://formatjs.io/docs/react-intl/components/#formattednumber) for this?
```
<FormattedNumber value={someVar} style="currency" />
```
The `currency` prop (e.g. USD, Euros, etc) will be pulled from the tenant's settings via [react-intl's context](https://github.com/folio-org/stripes-core/blob/a7facb28210f69eec39af70a54c8d27765316c94/src/components/Root/Root.js#L163).

Dmitriy-Litvinenko:
> Is there a reason not to use [`<FormattedNumber>`](https://formatjs.io/docs/react-intl/components/#formattednumber) for this?
> 
> ```
> <FormattedNumber value={someVar} style="currency" />
> ```
> 
> The `currency` prop (e.g. USD, Euros, etc) will be pulled from the tenant's settings via [react-intl's context](https://github.com/folio-org/stripes-core/blob/a7facb28210f69eec39af70a54c8d27765316c94/src/components/Root/Root.js#L163).

Hello @zburke 
How i understood result will be {currency symbol}{someVar} and we don't use currency symbol across all application.
Could you please correct me if i am wrong?

Dmitriy-Litvinenko:
> Is there a reason not to use [`<FormattedNumber>`](https://formatjs.io/docs/react-intl/components/#formattednumber) for this?
> 
> ```
> <FormattedNumber value={someVar} style="currency" />
> ```
> 
> The `currency` prop (e.g. USD, Euros, etc) will be pulled from the tenant's settings via [react-intl's context](https://github.com/folio-org/stripes-core/blob/a7facb28210f69eec39af70a54c8d27765316c94/src/components/Root/Root.js#L163).

Hello @zburke
How i understood result will be {currency symbol}{someVar} and we don't use currency symbol across all application.
@artem-blazhko Could you please double check and correct me if i am wrong?

artem-blazhko:
Hello @zburke.
I think it is not the best place to use `FormattedNumber`.
We do not need to show any currency or unit. 
In my opinion, `FormattedMessage` is more suitable.

zburke:
I don't understand why we wouldn't display this value as a currency, but let's put that aside. We agree this is a numeric value, right, so why not use the number-formatter react-intl hands us to make sure the thousands-separator and decimal characters are correct. What value is there in writing our own formatter?

artem-blazhko:
@zburke I added new changes and used `FormattedNumber` with the default style `decimal`.
I have not used style `currency` because we do not have any requirement to display currency symbols.
Thank you.

zburke (approved):

PR #2228, opened 2022-09-07
UIU-2666: "Financial Transaction Detail Report" title line no longer includes actual service points selected

artem-blazhko:
## Purpose
Initially, we did not have the list of service points in "Financial Transaction Detail Report". 
It happened because we used incorrect properties of some data.

## Refs
[UIU-2666](https://issues.folio.org/browse/UIU-2666)

## Screenshots

After changes we have a list of service points

image


zburke (approved):
Thank you for fixing this. It is really frustrating that some of these properties, especially `createdAt`, don't have better/less confusing/more consistent names.

PR #2234, opened 2022-09-19
lint: it does a body good

zburke:
Provide missing proptypes, correct some spacing, and remove `package-lock.json` since we're not using npm yet which means that even if we _wanted_ a lock file, this would be the wrong one.

After this change, PRs should be quieter because they will no longer contain yarn's litany of complaints about unchanged files with warnings.


zburke:
For some reason I find this stat hilarious: `+38 −41,799`

mkuklis (approved):

PR #2237, opened 2022-09-20
UIU-2540: Fix broken link to item in export fees fines report

Dmitriy-Litvinenko:
## Purpose
Fix broken link to item in export fees fines report

## Approach 
When we have all necessary data we render link 
When we have barcode with out other data, this mean that we can't find and navigate to item and we notify about that
Constant for empty value was added

## Refs
https://issues.folio.org/browse/UIU-2540


mkuklis (approved):
Nice cleanup @Dmitriy-Litvinenko!

zburke (approved):
LGTM

artem-blazhko (approved):

PR #2238, opened 2022-09-20
UIU-2651: Add minimal date validation to date fields

mkuklis:
https://issues.folio.org/browse/UIU-2651

Add minimal date validation to date fields to prohibit users from entering dates like: 0000-01-01.


zburke (approved):
LGTM. I really like the way you implemented `validateMinDate` in such a reusable fashion. There are three date fields on the user-edit screen. Do we want to add validation to those too? 
* "Date Enrolled" in EditExtendedInfo
* "Expiration date" in EditUserInfo
* "Expiration date" in EditProxy

mkuklis:
@zburke I think the answer is yes but it would be good to know what value should be used for the minimum date. Do you think `1900-01-01` is a good start at least?

artem-blazhko (approved):

zburke:
On the one hand, we have no authority to make this kind of decision, right? But on the other, it feels like 1900 is unlikely to cause problems in any of these fields. I say go for it. :shipit:

PR #2240, opened 2022-09-26
UIU-2376 include unit test for ItemLookup.js

VSnehalatha:
Unit test for ItemLookup. 
Refs UIU-2376(https://issues.folio.org/browse/UIU-2376)

### DIFF
--- a/src/components/Accounts/ChargeFeeFine/ItemLookup.test.js
+++ b/src/components/Accounts/ChargeFeeFine/ItemLookup.test.js
@@ -1,24 +1,22 @@
-import { screen, cleanup } from '@testing-library/react';
+import { screen, render } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
-
-import renderWithRouter from 'helpers/renderWithRouter';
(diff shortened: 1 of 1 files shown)


zburke (changes requested):
This is a good start. 

1. Please sign the contributor license agreement so we can accept your PR!
2. Please run `yarn lint` and fix all lint warnings.
3. The third test doesn't really do what it says. What it does is check that `onConfirm` can't be called when a row hasn't been clicked, which means the `Confirm` button is still disabled so clicking on it has no effect. This is good; this is a condition we want to check! But the test should say that, so change "OnConfirm modal check" to something like "Confirm is disabled until a row is selcted" and also assert on that with something like 
```
expect(screen.getByRole(button, { name: /confirm/ })).toBeDisabled()
```
4. Add a fourth test that passes some data shaped like `{ barcode, title }` to the `items` prop, click a row, and assert on what you expect to happen in `onConfirm`.

VSnehalatha:
Required changes are done as per review and committed. Please review the changes.

PR #2242, opened 2022-09-27
ProxyViewList test cases included

KetineniM:
UIU-2372

### DIFF
--- a/src/components/ProxyGroup/ProxyViewList/ProxyViewList.test.js
+++ b/src/components/ProxyGroup/ProxyViewList/ProxyViewList.test.js
@@ -1,0 +1,46 @@
+import { screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+
+import renderWithRouter from 'helpers/renderWithRouter';
+import ProxyViewList from './ProxyViewList';
+
+import '__mock__/stripesCore.mock';
+import '__mock__/stripesSmartComponent.mock';
+
+jest.unmock('@folio/stripes/components');
+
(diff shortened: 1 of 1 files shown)


zburke (changes requested):
This is a good start. 

1. Sign the contributor license agreement so we can accept your contribution!
2. Run `yarn lint` and fix all lint warnings. 
3. Add a test to cover the two missing conditions, i.e. when `items` is non-empty, since displaying `items` is the main purpose of this component. You'll need to populate `records` with a few simple objects like `{ name: 'apple' }` and `{ name: 'banana' }` and `itemComponent` with a dummy component that takes `record` as a prop just as `ComponentToRender` does on `ProxyViewList` line 10, e.g. 
```
itemComponent: ({ record }) => <div>{record.name}</div>,
```
4. correct the spelling errors in the `it` descriptions: `s/sponsers/sponsors/`

src/components/ProxyGroup/ProxyViewList/ProxyViewList.test.js
```diff
@@ -0,0 +1,49 @@
+import { screen } from '@testing-library/react';
+// import userEvent from '@testing-library/user-event';
```
zburke, line 2: Please prune unused code.

zburke (changes requested):
Yay 100% coverage 🎉 Nice work!

For me, this test ran successfully without these three lines:
```
import '__mock__/stripesCore.mock';
import '__mock__/stripesSmartComponent.mock';

jest.unmock('@folio/stripes/components');
```
I think they may have been copied in from other tests. Please remove them and then I'll merge this.

zburke (approved):

PR #2244, opened 2022-09-29
UIU-2375 include unit test for UserLoans.js

VSnehalatha:
Unit test for src/components/UserDetailSections/UserLoans/UserLoans.js,
Refs [UIU-2375](https://issues.folio.org/browse/UIU-2375)

### DIFF
--- a/src/components/UserDetailSections/UserLoans/UserLoans.test.js
+++ b/src/components/UserDetailSections/UserLoans/UserLoans.test.js
@@ -1,0 +1,129 @@
+import { screen } from '@testing-library/react';
+import '__mock__/stripesComponents.mock';
+import UserLoans from './UserLoans';
+import renderWithRouter from 'helpers/renderWithRouter';
+const toggleMock = jest.fn();
+jest.unmock('@folio/stripes/components');
+jest.unmock('@folio/stripes/smart-components');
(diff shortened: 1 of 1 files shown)


zburke (changes requested):
100% coverage! Nice work :tada:

1. Run `yarn lint` and correct all lint warnings.
2. The second and fourth tests make assertions on the same set of conditions. It's OK to include multiple `expect` clauses in a single `it` block. You could change the `it` description to something more descriptive like, "Correctly displays open and closed loan counts". You don't _have_ to put the assertions in a single block, but doing so will speed up the test a little since it will eliminate a render. 
3. The description in the last test, `< 0`, doesn't match the condition, `0`. How about "count is 0" or "count is < 1"?
4. A minor nit: in the object returned by `props()`, `location.search` is misspelled, which suggests you may not need this property at all.  Consider paring down the data in the mock if you don't need it all. e.g. does this component actually access the resources' `records.loans` properties, or just `totalRecords`? You don't _have_ to pare down `props`, but as a rule simpler is better, and less work! OTOH, if providing these "unused" props satisfies the prop-types checker, then it's worth keeping, e.g. it appears `location` _is_ necessary, even though `location.search` is not.

PR #2258, opened 2022-10-07
UIU-2693: A long Instance title climbs out of popover boundaries on the Lost items requiring actual cost page

Dmitriy-Litvinenko:
## Purpose
A long Instance title climbs out of popover boundaries on the Lost items requiring actual cost page

## Refs
https://issues.folio.org/browse/UIU-2693


Dmitriy-Litvinenko:
> [MDN says `word-break: break-word` is deprecated](https://developer.mozilla.org/en-US/docs/Web/CSS/word-break) but can be replaced by
> 
> ```
> word-break: normal;
> overflow-wrap: anywhere;
> ```
> 
> Let's use that instead. Agreed, @JohnC-80?

Hello @zburke thank you. Fixed.

artem-blazhko (approved):

JohnC-80 (approved):

zburke (approved):
LGTM. 

In the future, please do not use `--force-push`; instead, use the "Squash and merge" option when merging your PR. This keeps the history linear for people browsing commits directly in git while also maintaining the context of comments on the PR in GitHub. `--force-push` destroys the context of comments on a PR, e.g., my comments about `break-word` no longer make any sense because there is no reference to it in the commit history.

mkuklis (approved):

PR #2259, opened 2022-10-07
UIU-2673: Change Fees/Fines History to use an ACTIONS menu instead of buttons

ymacedor:
[UIU-2673](https://issues.folio.org/browse/UIU-2673) Change Fees/Fines History to use an ACTIONS menu instead of buttons

imagen


ymacedor:
@mkuklis could you review? thanks

mkuklis (approved):

PR #2261, opened 2022-10-11
UIU-2696: Fix user form validation

mkuklis:
https://issues.folio.org/browse/UIU-2696

It looks like after we refactored the user form to the final form in https://github.com/folio-org/ui-users/pull/1645 the main validation stopped working. This is because the signature of the **validate** changed:

and it only passes a single argument (current values):

https://final-form.org/docs/final-form/types/Config#validate

This makes it so:

https://github.com/folio-org/ui-users/blob/87e611c1c2a79a8dafa528a8528f9b238584439c/src/views/UserEdit/UserForm.js#L50-L52

is always true completely stopping validation.


MikeTaylor (approved):
Satisfyingly simple.

zburke (approved):
LGTM. A changelog entry would be worthwhile. Even if it's just "fix validation regression" it could help future devs who might not have realized the signature has changed. But maybe it's a lot to expect that people are reading these CHANGELOGs that carefully 😆

PR #2264, opened 2022-10-12
UIU-2697: Remove password validation from UserForm

mkuklis:
https://issues.folio.org/browse/UIU-2697

This PR removed password validation from `UserForm`. The password field is no longer accessible from the form.

### DIFF
--- a/src/views/UserEdit/UserForm.js
+++ b/src/views/UserEdit/UserForm.js
@@ -55,10 +55,6 @@ function validate(values) {
     errors.username = <FormattedMessage id="ui-users.errors.missingRequiredUsername" />;
   }
 
-  if (!values.id && (!values.creds || !values.creds.password) && values.username) {
-    errors.creds = { password: <FormattedMessage id="ui-users.errors.missingRequiredPassword" /> };
-  }
-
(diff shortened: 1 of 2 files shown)


src/views/UserEdit/UserForm.js
```diff
@@ -55,10 +55,6 @@ function validate(values) {
     errors.username = <FormattedMessage id="ui-users.errors.missingRequiredUsername" />;
   }
 
-  if (!values.id && (!values.creds || !values.creds.password) && values.username) {
-    errors.creds = { password: <FormattedMessage id="ui-users.errors.missingRequiredPassword" /> };
-  }
-
```
zburke, line 61: @mkuklis, Is there a test case we should add/update here? The fact that this bug leaked through suggests we don't have adequate coverage for submitting the new-user form. That's pretty crucial functionality!

mkuklis: @zburke this bug leaked because this code was never executed (due to the fact that the validation was not functional at all). I do agree that we should probably add more tests to `UserForm`. They do not exist at this point at all.

PR #2265, opened 2022-10-13
UIU-2430 test UserForm

zburke:
Test validation function. I couldn't get the component to render :/

Refs [UIU-2430](https://issues.folio.org/browse/UIU-2430)


mkuklis (approved):
Thank you for this @zburke.

mkuklis:
@zburke I got stuck with that rendering issue too.

PR #2269, opened 2022-10-26
UIU-2415 include unit test for Filters

VSnehalatha:
Jest/RTL test for ./src/views/UserSearch/Filters.js
Refs: [UIU-2415](https://issues.folio.org/browse/UIU-2415)


zburke (approved):
LGTM. 

In the future, prefer interactions imported from [`@testing-library/user-event`](https://testing-library.com/docs/ecosystem-user-event) as they do a better job of representing a true user interaction, e.g. including future events like focus that won't be triggered by `fireEvent`.

PR #2270, opened 2022-10-26
UIU-2415

zburke:
PR #2269 doppelgänger

Jest/RTL test for ./src/views/UserSearch/Filters.js
Refs: [UIU-2415](https://issues.folio.org/browse/UIU-2415)


zburke:
Merged the original instead.

PR #2274, opened 2022-11-01
ProxyEditItem test cases

KetineniM:
JEST/RTL test cases for ProxyEditItem

Reference: https://issues.folio.org/browse/UIU-2371


zburke (commented):
@KetineniM this a great start. Are you able to bump coverage over the 80% threshold by adding form-interactions with the status field?

mkuklis:
@KetineniM please re-open this directly under folio project and we should be able to wrap this up.

PR #2247, opened 2022-10-03
UIU-2688 leverage yarn.lock

zburke:
Leverage `yarn.lock` and `yarn install --frozen-lockfile` so we can choose the cadence of our updates, in particular for dev-deps, instead of being surprised by build failures whenever a third-party lib publishes a buggy release.

Refs [UIU-2688](https://issues.folio.org/browse/UIU-2688), [STRIPES-790](https://issues.folio.org/browse/STRIPES-790)

PR #2275, opened 2022-11-08
UIU-2421 - Increase code coverage

Terala-Priyanka:
## Purpose
Increase the code coverage by writing more tests
Issue: https://issues.folio.org/browse/UIU-2421

## Note
The unit tests already existed for src/views/UserDetail.js
I have improved the coverage to as much as possible by adding few tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## 8.3.0 IN PROGRESS
 
+* create Jest/RTL test for UserDetail.js. Refs UIU-2421
+
 ## [8.2.0](https://github.com/folio-org/ui-users/tree/v8.2.0) (2022-10-24)
 [Full Changelog](https://github.com/folio-org/ui-users/compare/v8.1.0...v8.2.0)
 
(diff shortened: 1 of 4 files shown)


Terala-Priyanka:
I see yarn.lock is not included in .gitignore file and the repository also has one.
Thats probably the reason for having yarn.lock in the commit.
I have no clue why there it is without github:   :(
I just ran yarn install.
Could you help me with the way forward?

zburke:
@Terala-Priyanka: TL;DR: run `git checkout origin/master -- yarn.lock` to restore `yarn lock` from the `#master` branch, then add it, commit, and push. Since the file will now match the one on the `#master` branch, it will no longer be listed on the "Changed files" tab. 

Details: We will begin committing `yarn.lock` files in all repos to reduce the problems caused by accidental-breaking changes in third party libraries (here it was in PR #2247). The reasoning is detailed in [STRIPES-790](https://issues.folio.org/browse/STRIPES-790). To obey the `yarn.lock` file when installing dependencies in local development run `yarn --frozen-lockfile` instead of `yarn` or `yarn install`. 

Alternatively/in addition, instead of staging changes with `git add -a`, use something like `git add src` or `git add <filename>`. You can also remove a staged file with `git reset HEAD -- <filename>` if you accidentally modified `yarn.lock` and ran `git add ...` but haven't yet committed.

Terala-Priyanka:
Thank you @zburke
Restored yarn lock file.

mariia-aloshyna (approved):

PR #2280, opened 2022-11-16
UIU-2418: Include test cases for LoanProxyDetails

VSnehalatha:
Unit test cases for LoanProxyDetails.js
REFS: [UIU-2418](https://issues.folio.org/browse/UIU-2418)


zburke (changes requested):
This is a great start! 

Please add a check for the condition where, given `props.id`, `props.mutator.proxy.GET()` throws and therefore calls `props.showErrorCallout`. You just need to make sure that the function you pass in as `showErrorCallout` is called with expected translation key given on line 39.

zburke (approved):
100% ❤️ 👏 👏 👏

PR #2288, opened 2022-11-29
UIU-2736: Refactor code for disable 'Declare lost' button after click to prevent multiple submissions

Dmitriy-Litvinenko:
## Purpose
Refactor code for disable 'Declare lost' button after click to prevent multiple submissions.

## Approach
UIU-2372 introduced a generic isInProgress prop which should be used in lieu of props like claimedReturnInProgress or declarationInProgress that pollute this generic component with business logic they don't need to know about.

## Refs
https://issues.folio.org/browse/UIU-2736

### DIFF
--- a/src/components/Wrappers/withDeclareLost.js
+++ b/src/components/Wrappers/withDeclareLost.js
@@ -13,7 +13,7 @@ const withDeclareLost = WrappedComponent => class WithDeclareLost extends React.
       declareLostDialogOpen: false,
       loan: null,
       itemRequestCount: 0,
-      declarationInProgress: false,
+      declareLostInProgress: false,
     };
   }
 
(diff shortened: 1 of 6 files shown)


artem-blazhko (approved):

src/components/Wrappers/withDeclareLost.js
```diff
@@ -47,8 +47,8 @@ const withDeclareLost = WrappedComponent => class WithDeclareLost extends React.
       <>
         <WrappedComponent
           declareLost={this.declareLost}
-          declarationInProgress={declarationInProgress}
-          toggleButton={this.toggleButton}
+          declareLostInProgress={declareLostInProgress}
+          setDeclareLostInProgress={this.setDeclareLostInProgress}
```
zburke, line 51: We can remove these props just like in #2289, right? i.e. they're passed directly to `<LoanActionDialog>` via props below, so there's no need to propagate them through the wrapper.

Dmitriy-Litvinenko: Hello @zburke in current case we cannot remove them because we use them in stripes/ui-users/src/views/LoanDetails/LoanDetails.js

artem-blazhko (approved):

zburke (approved):

PR #2289, opened 2022-11-29
UIU-2732: Disable 'Claimed return' button after click to prevent multiple submissions

Dmitriy-Litvinenko:
## Purpose
Disable 'Claimed return' button after click to prevent multiple submissions.

## Approach
Remove unused variables for remove possible get side effect

## Refs
https://issues.folio.org/browse/UIU-2732


artem-blazhko (approved):

zburke (approved):
Checking my understanding: `<LoanActionDialog>` already directly receives `isInProgress: claimReturnedInProgress` and `toggleButton: this.setClaimReturnedDialog` via props, so propagating them through `<WrappedComponent>` has no effect. 

Thanks for cleaning this up.

PR #2301, opened 2022-12-21
UIU-2749: Update versions of interfaces request-storage

artem-blazhko:
## Purpose
Update request-storage interface version due to braking changes in mod-circulation-storage

## Refs
[UIU-2749](https://issues.folio.org/browse/UIU-2749)


zburke (approved):
LGTM. Are you confident that adding the interface to the list, rather than replacing the list with a single version as at folio-org/ui-requests/pull/989, is correct?

artem-blazhko:
> LGTM. Are you confident that adding the interface to the list, rather than replacing the list with a single version as at [folio-org/ui-requests/pull/989](https://github.com/folio-org/ui-requests/pull/989), is correct?

@zburke Yes. I discussed it with @OleksandrVidinieiev. He said that `5.0` version is optional and it can be added to the list of already existing versions.

mariia-aloshyna (approved):

PR #2302, opened 2022-12-21
UIU-2750: Upgrade yarn.lock fixing decode-uri-component DoS CVE

julianladisch:
Upgrading yarn.lock upgrades decode-uri-component from 0.2.0 to 0.2.2 fixing Denial of Service (DoS):

https://nvd.nist.gov/vuln/detail/CVE-2022-38900

For production not this yarn.lock file but the platform yarn.lock file is used. However, the builds from this repository should contain the fixed decode-uri-component version so that any side-effects pop up early and we can be sure that ui-users is ready to use the fixed decode-uri-component version.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -16,9 +16,9 @@
     "@jridgewell/trace-mapping" "^0.3.9"
 
 "@apollo/client@^3.2.1":
-  version "3.7.1"
-  resolved "https://registry.yarnpkg.com/@apollo/client/-/client-3.7.1.tgz#86ce47c18a0714e229231148b0306562550c2248"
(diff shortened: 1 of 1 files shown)


zburke:
@julianladisch, one of the less-fun aspects of dealing with `yarn.lock` files is that we don't have great resources for handling precise updates like "bump the version of a transitive dependency". In point of fact, the updated `yarn.lock` file you included here includes updates for 80 different libraries. One approach is to add a `resolutions` entry to `package.json` for the affected library in order to generate a new `yarn.lock` and then remove the `package.json` change. That approach works here, but it sure is cumbersome. 

The stripes community is exploring different possibilities about how to do this with yarn, how to leverage that effort so we don't waste it as we continue the push to migrate from yarn to npm, and how to incorporate tools like dependabot to keep us from falling behind on this kind of maintenance. I appreciate your attentiveness to security concerns, but it would be helpful to us if you could refrain from opening further security-related PRs based on the versions you find in UI apps' `yarn.lock` files while our processes and policies are still in flux.

zburke (approved):

PR #2309, opened 2023-01-05
UIU-2487: In Settings > Users > Patron Blocks > Conditions, mark "Message to be displayed" as required

UladzislauKutarkin:
In Settings > Users > Patron Blocks > Conditions, mark "Message to be displayed" as required
Refs: [UIU-2487](https://issues.folio.org/browse/UIU-2487).
210304407-f72d2a4b-fa17-4782-a65e-7d3b96de9ace
210304389-aef644b3-37bc-4f91-ba88-b3925d845895

### DIFF
--- a/src/settings/patronBlocks/Conditions/Conditions.test.js
+++ b/src/settings/patronBlocks/Conditions/Conditions.test.js
@@ -40,7 +40,7 @@ const mockPut = jest.fn().mockImplementation(() => Promise.resolve());
 
 const mutator = {
   patronBlockCondition: {
-    DELETE: jest.fn().mockReturnValue(Promise.resolve()),
+    DELETE: jest.fn(),
     POST: jest.fn(),
     PUT: mockPut,
     cancel: jest.fn(),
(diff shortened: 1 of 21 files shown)


vashebsco (approved):

vashjs (approved):

src/settings/patronBlocks/Conditions/Conditions.test.js
```diff
@@ -81,6 +81,6 @@ describe('Conditions', () => {
     act(() => {
       userEvent.click(screen.getByText('stripes-core.button.save'));
     });
-    expect(mockPut).toHaveBeenCalledTimes(1);
+    expect(mockPut).toHaveBeenCalledTimes(0);
```
zburke, line 84: 1. This test is not so much "handle submit" but "prevent submit when required fields are empty". It would be nice to add an `expect` clause looking for the validation warning.
2. It would be nice to add an additional test where required fields are filled in, validation passes, and `mockPut` gets called.

zburke (approved):
LGTM but it would be really nice to test both conditions: failed validation (because a required field is empty) which blocks submit, and successful submit (because required fields are filled in).

uladislausamets (approved):

UladzislauKutarkin:
I have to revert changes due backend not supporting that future yet. Also fixed utils test that always fail, probably that comes from new Date()

zburke:
The implication of this revert, @UladzislauKutarkin, is that this PR should have been accompanied by either (a) changes in `package.json` that bumped the `okapiInterface` version and major version update or (b) wrapping the code changes in `<IfInterface>` checks for a particular interface version.

PR #2320, opened 2023-01-19
refactor brittle tests

zburke:
Tests should not be dependent on dynamically generated element IDs. These values are subject to change for who knows what reasons, say, maybe, a change in the rendering of elements in a third party library when `yarn.lock` changes, which causes those tests to break even though there has been no material change.


mkuklis (approved):
❤️

PR #2337, opened 2023-02-02
UIU-2788: Use updateUser from @folio/stripes-core to update service-point data

mkuklis:
https://issues.folio.org/browse/UIU-2788

The `setServicePoints` and  `setCurServicePoint` were removed from `stripes-core` in https://issues.folio.org/browse/UIU-2788

Instead, a new API was introduced called `updateUser` to modify user session data.

This PR replaces `setServicePoints` and  `setCurServicePoint` with `updateUser` to align with the recent changes in stripes-core.


zburke (approved):
LGTM!

PR #2339, opened 2023-02-08
UIU-2790 show users' service points correctly

zburke:
Correctly update the list of service points for each user when navigating among users by making sure local state is updated whenever the user changes. Previous comparisons between state and resources were insufficient.

Refs [UIU-2790](https://issues.folio.org/browse/UIU-2790)


Dmytro-Melnyshyn (approved):

mkuklis (approved):
Nice sleuthing @zburke!

Thank you for the extra comments!

BogdanDenis (approved):

PR #2347, opened 2023-02-15
UIU-2700: "Refund" option should not be available for cancelled fee/fine that was already refunded

artem-blazhko:
## Purpose
Fee/fine that was refunded with the full amount and then canceled as an error can not be refunded one more time.

## Refs
[UIU-2700](https://issues.folio.org/browse/UIU-2700)


Dmitriy-Litvinenko (approved):

s3fs (approved):

zburke:
Please include the Jira slug (e.g. UIU-2700) in the commit subject when you squash. It makes browsing the history much easier. Thank you!

PR #2364, opened 2023-03-14
UIU-2822: Correctly handle removing all permissions from user

mkuklis:
In cases when all permissions were removed from the given user the `permissions` object was undefined and the `updatePermissions` function was never executed.

This PR is trying to address it.

https://issues.folio.org/browse/UIU-2822


zburke:
@mkuklis, this is exactly the same problem I solved for proxies in #2350. Do you think we should go hunting for other places that manage list-attributes? I hesitate to call two points a pattern, but, well, this feels like a pattern.

mkuklis:
@zburke yes that looks very similar. I think the user update/create looks fine now: https://github.com/folio-org/ui-users/blob/de2d26732fcd82b296b3ae79e50eeb8427846f7f/src/views/UserEdit/UserEdit.js#L188

but we can take a look at other places.

PR #2384, opened 2023-03-30
UIU-2835 - A result is not displayed in list of permissions after clicking on the 'Search' button

Terala-Priyanka:
## Purpose
A result is not displayed in list of permissions after clicking on the 'Search' button

## Refs
https://issues.folio.org/browse/UIU-2835


zburke (approved):
LGTM. In fact the earlier fix had two problems:
1. it didn't work for compiled translations
2. it only worked for exact matches
It would be nice to include a comment explaining why `query` needs to be escaped. Right now, it may not be clear what problem we are solving. i.e. given a label like `Label (with parens): blah blah blah` we want a `query` value like `label (` to match, and that only works if the `(` is treated literally, rather than as a regex special character.

Dmytro-Melnyshyn (approved):

PR #2385, opened 2023-04-02
UIU-2842 run tests on multiple cores

zburke:
Based on comments from https://jestjs.io/docs/troubleshooting#tests-are-extremely-slow-on-docker-andor-continuous-integration-ci-server: 

> While Jest is most of the time extremely fast on modern multi-core computers with fast SSDs, it may be slow on certain setups as our users [have](https://github.com/facebook/jest/issues/1395) [discovered](https://github.com/facebook/jest/issues/1524#issuecomment-260246008).
> [...]
> ...
> If you use GitHub Actions, you can use [github-actions-cpu-cores](https://github.com/SimenB/github-actions-cpu-cores) to detect number of CPUs, and pass that to Jest.

Refs [UIU-2842](https://issues.folio.org/browse/UIU-2842), [STRIPES-854](https://issues.folio.org/browse/STRIPES-854)

### DIFF
--- a/.github/workflows/build-npm-release.yml
+++ b/.github/workflows/build-npm-release.yml
@@ -88,8 +88,12 @@ jobs:
         run: yarn lint
         continue-on-error: true
 
+      - name: Get number of CPU cores
+        id: cpu-cores
+        uses: SimenB/github-actions-cpu-cores@v1
+
       - name: Run yarn test
-        run: xvfb-run --server-args="-screen 0 1024x768x24" yarn test $YARN_TEST_OPTIONS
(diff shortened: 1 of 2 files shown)


zburke:
I don't know where the timing value in the PR decoration (6m 11s) is coming from. Compared to the last GA build on master, adding `--max-workers` based on detected cores reduced the test suite runtime from >450 seconds to ~175 seconds. 🔥 🔥 🔥

mkuklis (approved):

mkuklis:
👏

PR #2424, opened 2023-04-25
UIU-2811 - Unable to close some accordions on the Edit form.

JohnC-80:
https://issues.folio.org/browse/UIU-2811

The offending `<Accordion>`s had a hard-coded `open` prop of `true` being passed in via their wrapping implementations' `expanded` prop. Using this prop overrides the value that they get from their toggled state within the wrapping `<AccordionStatus>` component.

Approach: I just removed the `expanded` prop from the `UserForm` instances, and the accordions work as expected.


Terala-Priyanka (approved):
@JohnC-80 could you please update changelog.md file before merging this PR?

BogdanDenis (approved):

PR #2443, opened 2023-05-09
UIU-2863 - "ui-users.settings.customfields.view" permission insufficient to view custom fields on user settings.

Terala-Priyanka:
## Purpose
"ui-users.settings.customfields.view" permission insufficient to view custom fields on user settings.

## Approach
There is problem in viewing the custom fields when a user is assigned just the permission "ui-users.settings.customfields.view". The errors are documented in the ticket, linked stated below.

Add permission "user-settings.custom-fields.collection.get" to permission set "ui-users.settings.customfields.view"

## Ref
https://issues.folio.org/browse/UIU-2863


BogdanDenis (approved):
Looks good!
One question - is it ok to commit yarn.lock changes in ui-users?

Terala-Priyanka:
> Looks good! One question - is it ok to commit yarn.lock changes in ui-users?

I think yes, yarn.lock file is probably being committed since long by now.

zburke (approved):

PR #2446, opened 2023-05-10
Resolve issue with Edit/Create form accordions remaining open/fix associated prop-types

MikeTaylor:
These were polluting my JS console.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@
 * Show success and Error toasts when Assign/unassign a users affiliations. Refs UIU-2852.
 * "ui-users.settings.customfields.view" permission insufficient to view custom fields on user settings. Refs UIU-2863.
 * Align affiliation assignment with stripes-core updates (switch active affiliation). Refs UIU-2855.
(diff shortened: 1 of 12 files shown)


JohnC-80 (changes requested):
Apologies, @MikeTaylor - setting these to `true` is actually problematic as it doesn't allow the Accordions to be collapsed (also annoying!)
The necessary thing to do here would be to remove the `.required` from these prop-types. It isn't required on the `<Accordion/>` that either of them render internally, so it shouldn't be required in these cases - I suspect it was just a copy/paste mistake.

One of them is in this repo (EditExtendedInfo.js) at src\components\EditSections\EditExtendedInfo\EditExtendedInfo.js
The other is in stripes-smart-components, which I'll submit a PR for in just a moment...

MikeTaylor:
Aha! A triumph for peer-review. Now I am glad I didn't just go ahead and merge this!

(It feels a bit weird that true, false and undefined have three different meanings, but it's what it is.)

Thanks for taking this on.

JohnC-80 (approved):

PR #2449, opened 2023-05-11
UIU-2431 create JEST/RTL test cases for UserSearch.js

ssandupatla:
UIU-2431 create JEST/RTL test cases for UserSearch.js
URL: https://issues.folio.org/browse/UIU-2431

### DIFF
--- a/src/views/UserSearch/UserSearch.test.js
+++ b/src/views/UserSearch/UserSearch.test.js
@@ -1,0 +1,471 @@
+import React from 'react';
+import userEvent from '@testing-library/user-event';
+import {
+  screen,
+} from '@testing-library/react';
+import {
+  IfInterface,
+  IfPermission,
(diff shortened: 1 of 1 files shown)


zburke (changes requested):
16/20 tests here end by taking some action without a follow-up `expect` call to validate that something happened as a result of that action. This validates that the function was callable, but that's it: it doesn't establish that the function behaves correctly. 

Maybe you need to add assertions that a "submit" button is disabled until required fields have been filled out, and then assert that that it's enabled when they are completed, and then assert that clicking the button causes it to to disappear (presumably because the modal containing it closes, which is something else you could assert). Maybe you need to mock an API so you can assert that when somebody types 3/12/2023 into a date-field and then hits "Save" it closes the modal and makes an API request to `some-report?query=startDate=2023-03-12`. etc. etc.

PR #2458, opened 2023-05-16
test clean-up

zburke:
* omit `screen.debug`
* omit private paths to translations in test harnesses


MikeTaylor (commented):
Why doesn't this break tests? Surely the look for specific English-language text that they expect to see rendered?

PR #2462, opened 2023-05-17
UIU-2855 Align affiliation assignment with BE changes

NikitaSedyx:
## Purpose
https://issues.folio.org/browse/UIU-2855

Persist consortia requests to consortium central tenant and check permissions there.

## Approach
- fetch permissions from central tenant to allow consortia operations
- use consortia config to define central tenant
- add unit tests


alisher-epam:
@NikitaSedyx The only thing is updates to the `changelog` is missing

NikitaSedyx:
@alisher-epam it's not missing, @usavkov-epam added it in previous request

alisher-epam:
All LGTM

alisher-epam (approved):

BogdanDenis:
Hey @NikitaSedyx, with more modules (ui-inventory, ui-marc-authorities, ui-quick-marc) now having consortia-related features in backlog I think it would make sense to move the hooks to get consortia permissions, and to check if user has a consortia permission to `stripes-core`. What do you think?
cc: @zburke

zburke:
@BogdanDenis, @NikitaSedyx, I truly appreciate your desire to centralize some of these hooks and permissions checks but I have to be honest: I'm still really uneasy with the architecture of the consortia work. Making ui-users, ui-inventory, ui-_anything_ aware that it's searching a consortium still feels weird to me, but I don't know if that indicates a problem with the design or a problem in my head. If this were separated into MVC architecture, I'd say "plug in new models and views; the controller doesn't need to care/need to know." Here, we don't have strictly separate pieces, but I can't shake the feeling that adding "if consortia" conditions all over the place is deeply deeply wrong. 

It seems to me that consortia should either be fundamentally a part of every model in the system or completely isolated and pulled in via something like a plugin where tenants who care install ui-plugin-fetch-consortia-users and and tenants who don't install ui-plugin-fetch-users. Of course, comments on an already-merged PR are not a helpful place to have this conversation :) 

For now, I guess stripes-core is the best place for these shared things.

PR #2478, opened 2023-05-31
UIU-2880: User doesn't have an access to Limits menu option in Settings->Users menu option with the permission that allows to make changes to patron block limits

artem-blazhko:
## Purpose
User that has "Settings (Users): Can create, edit and remove patron blocks limits" permission does not have access to Limits menu option.

## Approach
Extended already existing solution by adding "settings.users.enabled" permission

## Refs
[UIU-2880](https://issues.folio.org/browse/UIU-2880)

## Notes
Extend changes that were added in the scope of [ui-users 2474 PR](https://github.com/folio-org/ui-users/pull/2474).
`CHANGELOG.md` file was not changed because all necessary changes were added in the scope of  [ui-users 2474 PR](https://github.com/folio-org/ui-users/pull/2474).


package.json
```diff
@@ -657,6 +657,7 @@
         "permissionName": "ui-users.settings.limits.view",
         "displayName": "Settings (Users): Can view patron blocks limits",
         "subPermissions": [
+          "settings.users.enabled",
```
Terala-Priyanka, line 660: It is recommended to move this line to sub permissions of "ui-users.settings.conditions.view" in line#621

Terala-Priyanka (changes requested):

package.json
artem-blazhko: Hello @zburke 
Could take a look at that changes?
In general, my goal is to have a "Settings" menu visible if a user has `"Settings (Users): Can view patron blocks limits"` permissions. As a result, I add  `"settings.users.enabled"` sub-permission to `"Settings (Users): Can view patron blocks limits"`. This approach was approved by PO.

Terala suggests moving`"settings.users.enabled"` sub-permission to `"Settings (Users): Can view patron blocks conditions"` which is a part of `"Settings (Users): Can view patron blocks limits"`. 
And we have some concerns about the correct approach.

Could you help to understand what is the correct way here?

cc @Dmitriy-Litvinenko, @Terala-Priyanka

zburke: `settings.users.enabled` must be part of _every_ settings permission set. You have three choices here:

1. Since permission sets are hierarchical and `ui-users.settings.conditions.view` is part of this pset adding it there kills two birds with one stone but is a little bit indirect since you have to trace the permissions hierarchy to see that all the necessary values are included. This is the most bang for the buck.
2. If you add it only here, it fixes UIU-2880 but leaves open a (not-yet-written) bug analogous to this one: a user with _only_ `ui-users.settings.conditions.view` should have access to settings but doesn't. 
3. Adding it here and on line 621 resolves both UIU-2880 and the not-yet-written bug in (2). It's more verbose than (1) but also a little more explicit/easy to follow. 

I would vote for 1 or 3 but they are all legit solutions.

zburke (approved):
LGTM, but as @Terala-Priyanka notes it would be nice to fix this for both the `limits.view` and `conditions.view` psets.

Terala-Priyanka (approved):

PR #2484, opened 2023-06-08
UIU-2885 - update permissions only when the user has permissions to edit them

Terala-Priyanka:
## Purpose
Users with view-only access to user permissions and access to edit user records get error message when saving a user record

## Refs
https://issues.folio.org/browse/UIU-2885 

## Screenshots
chrome_iEqqBu0uO0


Terala-Priyanka:
Hello @artem-blazhko 
I think this is the maximum achievable coverage here

artem-blazhko (approved):

Dmitriy-Litvinenko (approved):

jeboden:
My institution is still having issues with this.  It has occured when editing proxy information and expiration dates.

PR #2489, opened 2023-06-13
prop-types cleanup

zburke:
Prop-types clean up during some React v18 testing. Could. Not. Resist.

### DIFF
--- a/src/settings/CommentRequiredForm.js
+++ b/src/settings/CommentRequiredForm.js
@@ -45,7 +45,7 @@ const Setting = ({
 );
 
 Setting.propTypes = {
-  label: PropTypes.element,
+  label: PropTypes.string,
   name: PropTypes.string,
   intl: PropTypes.object,
 };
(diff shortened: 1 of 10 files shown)


Terala-Priyanka:
Hey @zburke 
hope you would fix failures.

zburke:
Whoops, sorry, @Terala-Priyanka. I didn't realize how react-intl was mocked in that one test. Fixed!

src/settings/CommentRequiredForm.js
```diff
@@ -45,7 +45,7 @@ const Setting = ({
 );
 
 Setting.propTypes = {
-  label: PropTypes.element,
+  label: PropTypes.string,
```
Terala-Priyanka, line 48: Hey Zak,
I see the reason underlying in here (Select accepts aria-label prop of type string), that made you update label prop type as string.
Wouldn't eslint complaint in this case? How did it go unaddressed till now?

Terala-Priyanka (approved):

src/settings/CommentRequiredForm.js
zburke: ESLint is only going to do static analysis so it won't pick up this kind of thing.

PR #2490, opened 2023-06-14
UIU-2893 Use new WSAPI for adding patron/staff notes to loans

MikeTaylor:
Fixes UIU-2893.

No user-visible changes, just the use of a newer WSAPI.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Change history for ui-users
 
-## 9.1.0 IN PROGRESS
+## 10.0.0 IN PROGRESS
 
 * Add STATUS FILTER to LOST ITEMS REQUIRING ACTUAL COST processing page. Refs UIU-2748.
 * Correctly handle removing all permissions from given user. Fixes UIU-2822.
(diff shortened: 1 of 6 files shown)


zburke (commented):
If this does indeed leverage a "new WSAPI" as indicated in the title, should there be version bump in the okapi interface in `package.json`? Or does this leverage functionality that was already there, and we're only now leveraging it?

MikeTaylor:
> If this does indeed leverage a "new WSAPI" as indicated in the title, should there be version bump in the okapi interface in `package.json`? Or does this leverage functionality that was already there, and we're only now leveraging it?

See [this commit](https://github.com/folio-org/ui-users/pull/2490/commits/7cee4e1876b88b21808e31d4a5a5ece76dc28dbc).

zburke (changes requested):
We are ships passing in the night 😆 

A new required okapi interface is a breaking change. Please bump the package version to 10.0.0 and then this LGTM.

MikeTaylor:
@zburke Re-review, please!

zburke (approved):
LGTM

PR #2494, opened 2023-06-19
Revert "UIU-2896 Expose `<RenderPermissions>`, `<PermissionSetDetails>` and `<PermissionSetForm>` components for reuse"

usavkov-epam:
Reverts folio-org/ui-users#2491


MikeTaylor (approved):
Thank you for doing this. I know the frustration of having to revert work for esoteric-seeming reasons, but this really is the right thing and I appreciate your finding a different way.

NikitaSedyx (approved):

PR #2495, opened 2023-06-20
UIU-2895 correct subpermission for adding patron/loan notes

MikeTaylor:
Since the addition of the `/circulation/loans/{id}/add-info` WSAPI to mod-circulation, we use that to add patron and staff notes instead of doing a PUT to the loan object. As a result, the user-level permissions for adding patron and staff notes no longer require the high-powered `circulation.loans.item.put` subpermission, but do require the permission that guards the new API, `circulation.loans.add-info.post`.

Fixes UIU-2895.


nielserik (approved):
LGTM!

PR #2498, opened 2023-06-23
UIU-2901 handle sparse borrower data in overdue loans report

zburke:
If a user record attached to a loan in the overdue loans report has been removed, the `userId` attribute will be present but the `borrower` object will not be. The parser needs to be robust to this situation.

Curiously, such an NPE causes parsing to silently halt: there is no warning in the console, no stacktrace, nothing at all. `map` just never returns. In local tests, I get an NPE in similar situations, so I don't understand how the error gets swallowed here, but it does.

Refs [UIU-2901](https://issues.folio.org/browse/UIU-2901)


Terala-Priyanka (approved):

BogdanDenis (approved):

BogdanDenis:
Hey @zburke, here's where the error is caught
https://github.com/folio-org/ui-users/blob/master/src/views/UserSearch/UserSearch.js#L251
But because error message is different - it never gets printed to console
image

zburke:
Thanks, @BogdanDenis. @mkuklis pointed out the same to me. I don't know how I missed it! Do you think this PR should include better error reporting in that catch statement, instead/in addition to the null-check I added when parsing the `borrower` record? Suddenly, this PR feels a bit anemic since the scope of the fix is so narrow.

BogdanDenis:
> Do you think this PR should include better error reporting in that catch statement, instead/in addition to the null-check I added when parsing the `borrower` record?

Yes, I think so. I see that you've changed it so the user will at least get notified about an error - looks good to me!

PR #2514, opened 2023-07-14
UIU-2588: Include info about reminder fees in Overdue Loans Report

mkuklis:
https://issues.folio.org/browse/UIU-2588


MikeTaylor (approved):
Looks good!

PR #2521, opened 2023-07-21
UIU-2905 - User settings > Comment required: Disable editing for users with "Setting (Users): View all settings" permission

Terala-Priyanka:
## Purpose
User settings > Comment required: Disable editing for users with "Setting (Users): View all settings" permission

## Approach
Seems like "comments required" page contents/configuration doesn't exist by default. When the page loads for the first time, if the config is not available from GET call response, then the configuration is currently being posted from FE. 

Currently, I see an error popup, stating that a permission 'comment.item.post' is required, when logging in with a user who doesn't have "Settings (Users): Can edit if comment required" but has "Settings (Users): View all settings" permission.

Hence, updating the logic.

**NOTE:** A PR for this same ticket has already been approved and closed. These new changes address the bug observed while testing the changes on snapshot (after merge). Hence there is no need to update changelog.md in this case.

## Refs
https://issues.folio.org/browse/UIU-2905


OleksandrHladchenko1:
Update CHANGELOG.md, please

Terala-Priyanka:
> Update CHANGELOG.md, please

Hey @OleksandrHladchenko1 
I have mentioned the reason for not having the changelog updated. Please re-check the PR description - especially NOTE

UladzislauKutarkin (approved):

PR #2522, opened 2023-07-21
UI u 2902

Terala-Priyanka:


UladzislauKutarkin (approved):

vashjs (approved):

Terala-Priyanka:
@UladzislauKutarkin and @vashjs 
Thank you for your reviews.
I had reconsidered the changes to make them uniform as departments page.
Hence, used canCreate and actionSupressor props of ControlledVocab component rather than editable prop. 

There is not much difference in the result.
Sorry for the trouble.
Could you please re-review the changes?

vashjs (approved):
Please update changelog if required

UladzislauKutarkin (approved):

PR #2530, opened 2023-07-31
UIU-2903 - Add permission check to render copymodal in feefinesettings

Terala-Priyanka:
## Purpose
User settings > Manual Charges: Disable editing for users with "Setting (Users): View all settings" permission.
**NOTE:** This is an improvement PR on top of earlier one for the same ticket. and hence Changelog.md file is not being updated.

## Screenshots
Copy modal do not pop up after changing owner, when logged in with user who has view all user settings permissions.
msedge_2KHQghVQ0d

UlM05dYEyi

## Refs
https://issues.folio.org/browse/UIU-2903


src/settings/FeeFineSettings.js
```diff
@@ -317,13 +317,15 @@ class FeeFineSettings extends React.Component {
...
           hasEditOwnerPerm={hasEditOwnerPerm}
         />
-        <CopyModal
-          {...this.props}
-          openModal={this.state.showCopyDialog}
-          onCopyFeeFines={this.onCopyFeeFines}
-          onCloseModal={this.hideCopyDialog}
-          ownerList={filterOwners}
-        />
+        <IfPermission perm="feefines.item.post feefines.item.put">
```
artem-blazhko, line 320: Should we use a comma to split permission names?
```suggestion
        <IfPermission perm="feefines.item.post,feefines.item.put">
```

artem-blazhko (commented):
Could you add information to CHANGELOG file?

src/settings/FeeFineSettings.js
artem-blazhko, line 320: Should we use a comma to split permission names?
```suggestion
        <IfPermission perm="feefines.item.post,feefines.item.put">
```

Dmitriy-Litvinenko (approved):

artem-blazhko (approved):

PR #2538, opened 2023-08-11
UIU-2873: refactor permission set component and improve caching

alisher-epam:
## [UIU-2873](https://issues.folio.org/browse/UIU-2873) - Assign/unassign users from Permission set

**Purpose:** Due to `grantedTo` values being received from the parent component it caused some query caching issues. That's why, in order to fix those issues and reuse this component in consortium the refactor needed to be done.

**Approach:**
- Add `usePermissionSet` custom hook
- Pass `grantedToIds` to queryKey for better caching

https://github.com/folio-org/ui-users/assets/116072773/ddba2998-a8bd-4fc1-b628-51e6253cc3f7


mariia-aloshyna (approved):

alisher-epam:
@Terala-Priyanka I have tried with State before but it didn't work. I have explained the issue in the Root cause section in the PR description above. Basically, the users' list depends on `grantedTo` ids list. In the perviouse PR, I updated the grantedTo using state. It worked as expected for a while. Once, a user switches from one permission set to another, the `grantedTo` state is set to the initial grantedTo ids from `permissionSet` props. 
Here is the video describing the issue.


https://github.com/folio-org/ui-users/assets/116072773/0d2f3f77-7b2d-4a7e-a4db-ce293e5a62a2

Terala-Priyanka:
@alisher-epam 
When you tried to leverage useState hook, you have initiated state with props, that you could a potential root cause. I suspect that piece of code. Could you try to leverage useEffect hook in setting the state as and when the prop changes. Test and observe how the behavior. 

Initiating state with props is an anti-pattern in most of the cases in React.js. Ref - https://vhudyma-blog.eu/react-antipatterns-props-in-initial-state/#Better-Avoid-Props-In-Initial-State

alisher-epam:
@Terala-Priyanka Is it a suggestion or a request? In both cases, I haven't used state and I have removed since it was causing issues.

Terala-Priyanka:
const { grantedTo, id: permissionSetId, permissionName } = permissionsSet;
const [grantedToIds, setGrantedToIds] = useState(grantedTo);

This is the piece of code that I referred to.

PR #2556, opened 2023-09-15
UIU-2932: Add auto focus to textarea on staff and patron info modal

mkuklis:
https://issues.folio.org/browse/UIU-2932

Add auto focus to textarea on staff and patron info modal


MikeTaylor (approved):
Solution is way too complex.

:-)

PR #2566, opened 2023-10-05
UIU-2959 Accordion collapse/expand all shortcuts not working on Edit form.

JohnC-80:
## Purpose
In addition to the problem mentioned in the title, those particular handlers were coming up as `undefined` on closer inspection...

## Approach
Implemented some handlers using the centralized handler functions from stripes-components.

I needed to update the `yarn.lock` to pull in updated stripes-cli/stripes-webpack changes for handling icons.


Terala-Priyanka (approved):
Changes LGTM.
Do you think it is good to add couple of unit tests here (although expandAllSections and collapseAllSections must have been covered with tests in stripes-components) to increase the coverage number?

BogdanDenis (approved):

JohnC-80:
With `HasCommand` being mocked here, I'm not sure of the best way to test this, since the functionality of the few lines here all flow through that dependency. This is probably best for AQA team to cover in integration/cypress tests.

PR #2614, opened 2024-01-23
UIU-3029: Add patron notice print jobs to action menu

mkuklis:
https://issues.folio.org/browse/UIU-3029
https://issues.folio.org/browse/UIU-3030

Add patron notice print jobs to action menu and the appropriate permission sets.

menu

patron-notice-print-jobs


MikeTaylor (approved):
Same code as before, right?

mkuklis:
@MikeTaylor yes it's almost the same with some addition for permissions.

PR #2660, opened 2024-03-21
UIU-3078: use Save & close button label stripes-component translation key 

UladzislauKutarkin:
https://folio-org.atlassian.net/browse/UIU-3078
We removed local `saveAndClose` label and replace it with `stripes-components.saveAndClose`. Also all related tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [10.2.0] IN PROGRESS
 
+* UX consistency: Use Save & close button label stripes-component translation key. Refs UIU-3078.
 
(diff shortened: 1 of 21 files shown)


BogdanDenis (approved):

Terala-Priyanka (commented):
If this translation("ui-users.saveAndClose) is used nowhere else , it is worth cleanup. Please check.

UladzislauKutarkin:
> Can we remove the translation key from `/translations/ui-users/en.json` then, too? Hopefully, 🤞 that'll cause it to be removed from other locale files, just like adding a new key causes it to be added to other locale files.

Done!

PR #2665, opened 2024-03-28
UIU-2049 - trim empty string input

manvendra-s-rathore:
## Purpose 
[UIU-2049](https://folio-org.atlassian.net/browse/UIU-2049) - Trim input values and convert empty string to null.
Fix :- 
while creating a new user or updating existing user, when an user field (for ex. - externalSystemId) is empty string or string with all spaces and no characters(ex. - "     "), the field should be removed from the JSON on user record save.

## Screencast 
https://github.com/folio-org/ui-users/assets/105021655/930668dc-df69-4662-b5a0-c998665d90eb

## Pre-Merge Checklist


manvendra-s-rathore:
Hey @folio-org/fe-tl-reviewers , can someone please review this PR on priority.

src/views/UserEdit/UserEdit.js
```diff
@@ -154,6 +154,12 @@ class UserEdit extends React.Component {
       .catch((e) => showErrorCallout(e, this.context.sendCallout));
   }
 
+  externalSystemIdValidator = (user) => {
+    if (!user.externalSystemId?.trim()) {
+      delete user.externalSystemId;
+    }
+  };
+
```
Terala-Priyanka, line 162: I think this function can be generalized to work for any property within user. example - lines 171 - 173

Terala-Priyanka (changes requested):

src/views/UserEdit/UserEdit.js
BogdanDenis: Agree with @Terala-Priyanka, this can be a general function like `deleteEmptyUserField(field)`

BogdanDenis (approved):

src/views/UserEdit/UserEdit.js
```diff
@@ -154,6 +154,15 @@ class UserEdit extends React.Component {
...
 
+  deleteEmptyFields = (user) => {
+    if (!user.username) {
+      delete user.username;
+    }
+    if (!user.externalSystemId?.trim()) {
+      delete user.externalSystemId;
+    }
+  };
+
```
Terala-Priyanka, line 165: you may further refine it by 
1. creating an array of user properties to be deleted
2. loop through the array
as the action is same always - delete the property, if empty.

zburke: I would expect the "emptiness" check for these fields to be the same, or for there to be a comment explaining why they are different. 

This impl is easy, but I don't love functions with side-effects, i.e. that modify their arguments. IOW, technically it's OK but philosophically it's sketchy. If you're decide to go with because it's just so so simple (and that is a merit here), please document the side-effect.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 * UX consistency: Use Save & close button label stripes-component translation key. Refs UIU-3078.
 * Fix two "triangle down" icons in select element of "Copy existing fee/fine owner table entries" modal. Refs UIU-2929.
 * Fix incorrect translation key having count-disagreements in Pay fees/fines modal in Fees/Fines Page. Refs - UIU-1097.
+* Trim input values and convert empty string to null. Refs - UIU-2049.
```
zburke, line 8: This is not accurate. `{ foo: "bar", bat: null }` is different than `{ foo: bar }`. You're not setting empty strings to null; when you find a value that is an empty string, you are deleting that property altogether.

zburke (changes requested):
Documentation is inaccurate. 
Implementation is inconsistent.

Terala-Priyanka (approved):

manvendra-s-rathore:
Hello @zburke , can you please re review this PR, whenever you are available.

manvendra-s-rathore:
Hey @zburke, in the meantime, I'll be merging the current pull request. If you encounter any issues with it, please let me know.

zburke:
@manvendra-s-rathore , I'm sorry I didn't get to this in time. I do have concerns about the approach you've taken here. 

One point of order: when dealing with empty string/null/undefined, it's important to get your terms straight. Some words you are using have specific meanings in this context and you are using them incorrectly.
* "trim" means to remove whitespace from a string; if you trim a string that was only whitespace to begin with, the result will be an empty string. 
* "null" means the absence of any value; null and empty string both evaluate false but they are not the same thing.
* "removing" a field from an object removes the key in addition to the value; setting a key's value to null is different than removing the key altogether. 
These examples show some additional differences:
```
let foo = { bar: "bar" }; // foo.bar is "bar"
foo.bar = ""; // empty string, foo.hasOwnProperty('bar') is true, foo.bar.length is 0
foo.bar = null; // null, foo.hasOwnProperty('bar') is true, foo.bar.length throws TypeError
delete foo.bar; // removes the property, foo.hasOwnProperty('bar') is false, foo.bar.length throws TypeError
```

In this context, make sure your documentation and your code say and do the same thing. The title of this PR is "trim empty string input" but what it actually does is more like "remove empty fields" or "remove certain whitespace-only fields". 

Instead of using `field.split` and then a condition to handle the top-level and nested properties, use the lodash methods [`get`](https://lodash.com/docs/4.17.15#get) and [`unset`](https://lodash.com/docs/4.17.15#unset) so you can avoid the condition and handle properties at any depth in one go. As is, if somebody updates `USER_INFO` to include `NESTED: 'deeply.nested.property`, `deleteEmptyFields` will blow up. 

Be wary of the word "check" when naming variables or functions, e.g. `USER_FIELDS_TO_CHECK`. Check for what? It's obvious in the context of this PR that you're checking whether they have data and removing them, but for somebody who is reading `src/constants.js` later, the connection to `UserEdit.js` is not clear. A more descriptive name could be `OPTIONAL_FIELDS` or `DELETABLE_FIELDS`.

PR #2682, opened 2024-05-02
UIU-3133 "Reset password email sent" modal is not opened when clicking on "Send reset password email" while editing a user

ryandberger:
- Fixes [UIU-3133](https://folio-org.atlassian.net/browse/UIU-3133)
- Don't refresh User Edit page when clicking "Send reset password email. We know this is when action name is `refreshPassword`. Community code relies on password reset having a different base URL (`bl-users` for password reset and `users` for the User page). In the case of Eureka, the base URL is the same, so we need to explicitly tell the page not to refresh the content (which blocks the modal from appearing).
- Removed a duplicate file ` EditUserRoles.test.js` which also had a leading space in the filename.
- Also added a missing param `user` back into the Community `<UserPermissions>` component.


zburke (approved):
I'm still amazed you figured this out! Nice work.

PR #2689, opened 2024-05-10
UIU-3105: hide permissionSets from settings/users navigation pane if roles interface is present

aidynoJ:
## Purpose

[UIU-3105](https://folio-org.atlassian.net/browse/UIU-3105) - suppress Settings > Users > Permission sets when `roles` interface is present

## Pre-Merge Checklist


ryandberger (approved):

zburke (changes requested):
This works, but in Eureka, we don't have mod-permissions, so the `permissions` interface won't exist at all, right? If that's correct, lets implement like we did for [fees and fines on line 133](https://github.com/folio-org/ui-users/blob/3c5aa4d1450d9e9357762c2b516937bd33bf46be/src/settings/sections.js#L133), leveraging the filter on [SettingsPage](https://github.com/folio-org/ui-users/blob/3c5aa4d1450d9e9357762c2b516937bd33bf46be/src/settings/components/SettingsPage.js#L45) to implement this in a less obtrusive/more generic way.

zburke (approved):
I hesitate to approve this because `getSettingSections` is just so out of place, and so specific to the `perms` route. True, it gets the job done, but without any context since there are no comments here and the route in question is not defined here. 

I think a better approach may be to handle this directly in `sections.js` where we could use the same conditionals to evaluate whether to unshift this route onto the `settingsGeneral` list. 

I also don't understand how the new tests evaluate the condition we are adding here. All the tests do is check whether the settings page renders, not whether the content it renders changes when the interface list changes.

aidynoJ:
> I hesitate to approve this because `getSettingSections` is just so out of place, and so specific to the `perms` route. True, it gets the job done, but without any context since there are no comments here and the route in question is not defined here. 
> 
> 
> 
> I think a better approach may be to handle this directly in `sections.js` where we could use the same conditionals to evaluate whether to unshift this route onto the `settingsGeneral` list. 
> 
> 
> 
> I also don't understand how the new tests evaluate the condition we are adding here. All the tests do is check whether the settings page renders, not whether the content it renders changes when the interface list changes. 

I'll try to handle it differently.

src/settings/components/SectionPageItem.js
```diff
@@ -38,7 +42,9 @@ SectionPageItem.propTypes = {
     label: PropTypes.element,
     perm: PropTypes.string,
     route: PropTypes.string,
-  })
+    dependsOnNoneInterface: PropTypes.string
+  }),
+  hasInterface: PropTypes.func
```
zburke, line 47: Petty: when a list is sorted alphabetically, add new entries in order.

src/settings/components/SectionPageItem.js
```diff
@@ -5,7 +5,7 @@ import {
   NavListItem,
 } from '@folio/stripes/components';
 
-const SectionPageItem = ({ setting, path }) => {
+const SectionPageItem = ({ setting, path, hasInterface }) => {
```
zburke, line 8: Let's avoid re-implementing `hasInterface` in `SettingsPage` and passing it as a prop. We can just use the original:
```
const stripes = useStripes();
// ...
if (stripes.hasInterface(setting?.dependsOnNoneInterface)) { ... }
```

zburke (approved):
Ooooh, this is very nice! What if we rename `dependsOnNoneInterface` to `unlessInterface`?

PR #2705, opened 2024-06-12
UIU-3125: Make unblocked fee-fine actions if item is claimed returned and loan is opened

artem-blazhko:
## Purpose
When loan was closed by check in and when it has open fees or fines and if item is checked out again and gets claimed returned the previous borrower blocked from paying (waiving/transferring/canceling) their fees or fines.

## Approach
Initially we had disabled pay, waive, transfer, refund, cancel buttons if item had a status "Claimed returned". But now we additionally rely on loan status that should have value 'Closed'. In this case all mentioned buttons will be enabled.
(Approach confirmed by PO).

## Refs
[UIU-3125](https://folio-org.atlassian.net/browse/UIU-3125)

### DIFF
--- a/src/views/AccountDetails/AccountDetails.js
+++ b/src/views/AccountDetails/AccountDetails.js
@@ -43,6 +43,7 @@ import FeeFineReport from '../../components/data/reports/FeeFineReport';
 import {
   itemStatuses,
   refundClaimReturned,
+  loanStatuses,
 } from '../../constants';
 
 import css from './AccountDetails.css';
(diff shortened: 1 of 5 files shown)


Dmitriy-Litvinenko (approved):

src/views/AccountDetails/AccountDetails.js
```diff
@@ -385,7 +387,7 @@ feeFineActions
           <Button
             id="errorAccountActionsHistory"
             buttonStyle="dropdownItem"
-            disabled={disabled || buttonDisabled || isActionsPending || isAccountsPending || !cancelAllowed || isClaimReturnedItem}
+            disabled={disabled || buttonDisabled || isActionsPending || isAccountsPending || !cancelAllowed || isMenuButtonDisabled}
```
JohnC-80, line 390: Could these conditionals be consolidated rather than repeated for each button?

JohnC-80 (approved):

PR #2706, opened 2024-06-12
UIU-3109: Add the possibility to use "pay" and "waive" fee/fine permissions separately

artem-blazhko:
## Purpose
Add the possibility to use "pay" and "waive" fee/fine permissions separately.

## Approach
- change permission sets
- fix Sonar issues
- use showing/hiding approach for fee/fine action buttons instead of disabling/enabling approach

## Refs
[UIU-3109](https://folio-org.atlassian.net/browse/UIU-3109)
[UIU-3156](https://folio-org.atlassian.net/browse/UIU-3156)


Dmitriy-Litvinenko (approved):

JohnC-80 (approved):
Thank you! Approved!

PR #2731, opened 2024-08-07
"add-info" interface is optional to prevent dependency cascades

MikeTaylor:
The interface used for adding patron/staff notes to loans is now optional, and the relevant code guarded with `<IfInterface>`.

Fixes UIU-3182.

### DIFF
--- a/src/views/LoanDetails/LoanDetails.test.js
+++ b/src/views/LoanDetails/LoanDetails.test.js
@@ -526,18 +526,6 @@ describe('LoanDetails', () => {
       });
       expect(screen.getByRole('button', { name:'ui-users.loans.declareLost' })).toBeDisabled();
     });
-    it('should disable "New patron info" button', () => {
-      renderLoanProxyDetails({
-        ...virtualPatronPropsData,
-      });
-      expect(screen.getByRole('button', { name:'ui-users.loans.newPatronInfo' })).toBeDisabled();
-    });
-    it('should disable "New staff info" button', () => {
-      renderLoanProxyDetails({
-        ...virtualPatronPropsData,
-      });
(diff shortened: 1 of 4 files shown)


src/views/LoanDetails/LoanDetails.test.js
```diff
@@ -526,18 +526,6 @@ describe('LoanDetails', () => {
...
-        ...virtualPatronPropsData,
-      });
-      expect(screen.getByRole('button', { name:'ui-users.loans.newPatronInfo' })).toBeDisabled();
-    });
-    it('should disable "New staff info" button', () => {
-      renderLoanProxyDetails({
-        ...virtualPatronPropsData,
-      });
-      expect(screen.getByRole('button', { name:'ui-users.loans.newStaffInfo' })).toBeDisabled();
-    });
```
zburke, line 540: @MikeTaylor , you're correct that default `<IfInterface>` mock is incorrect in the context of this test, but like the `hasInterface` function on line 79, the default can be overridden. 

These tests are poorly named given what they actually assert, which is that the buttons are present but disabled. As you noted, these buttons should not even be present when `add-info` is absent. Can we test for that independently (mock IfInterface to return false and assert the buttons [or their container] is absent), and also keep the existing tests (with IfInterface mocked to return true) in place? 

CC: @Terala-Priyanka

MikeTaylor: Hiya, @zburke. I stand by my earlier conclusion that "the tests are fundamentally valueless and this was not a good use of my time", and I honestly can't summon up any enthusiasm for revisiting this now, especially with eight hundred other things going on which all have functional implications. But if you would like to reinstate some form of those tests, then by all means be my guest.

PR #2748, opened 2024-09-25
[UIU-3212] Error when saving edited user

ryandberger:
- Fixes [UIU-3212](https://folio-org.atlassian.net/browse/UIU-3212) by removing property `assignedRoleIds` from `PUT /users` request if `roles` interface is not present.
- Corrected API calls in `UserRecordContainer.js` to use variable in parameter instead of expecting `stripes-connect` which will only transform static string values and not functions.
- Removed duplicate code in `UserDetail.js`

### DIFF
--- a/src/hooks/useAllRolesData/useAllRolesData.js
+++ b/src/hooks/useAllRolesData/useAllRolesData.js
@@ -20,7 +20,7 @@ function useAllRolesData() {
   const [namespace] = useNamespace();
 
   const { data, isLoading, isSuccess } = useQuery([namespace, 'user-roles'], () => {
-    return ky.get(`roles?limit=${stripes.config.maxUnpagedResourceCount}&query=cql.allRecords=1 sortby name`).json();
+    return stripes.hasInterface('roles') && ky.get(`roles?limit=${stripes.config.maxUnpagedResourceCount}&query=cql.allRecords=1 sortby name`).json();
   });
 
   const allRolesMapStructure = useMemo(() => {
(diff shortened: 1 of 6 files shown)


ncovercash (commented):
I'm happy to review but...am I supposed to be here? lol

ryandberger:
> I'm happy to review but...am I supposed to be here? lol

I need a review to unblock QA. Zak and I discussed most of the changes, but he is unavailable right now. If you are able to spot-check, I'd appreciate it!

ncovercash (approved):
ship it 🚀

src/hooks/useAllRolesData/useAllRolesData.js
```diff
@@ -20,7 +20,7 @@ function useAllRolesData() {
   const [namespace] = useNamespace();
 
   const { data, isLoading, isSuccess } = useQuery([namespace, 'user-roles'], () => {
-    return ky.get(`roles?limit=${stripes.config.maxUnpagedResourceCount}&query=cql.allRecords=1 sortby name`).json();
+    return stripes.hasInterface('roles') && ky.get(`roles?limit=${stripes.config.maxUnpagedResourceCount}&query=cql.allRecords=1 sortby name`).json();
```
zburke, line 23: This works fine. A more conventional approach is to set the `enabled` attribute of the object passed to `useQuery`, e.g. `useQuery({ ..., enabled: stripes.hasInterface('roles')})`.

zburke:
Thanks for fixing all the things I screwed up in c8acc87020014f243cab7f39279ad64f3d117c86! 😄 😅 😬

PR #2770, opened 2024-10-18
UIU-3224 Implement duplicates users view

usavkov-epam:
## Purpose

https://folio-org.atlassian.net/browse/UIU-3224

Display a list of duplicates for pre-stage patron user.

---
The first part of UIU-3224 - #2769 

## Approach

 - Create a separate route for viewing duplicates of pre-stage users;
 - An autosized MCL will occupy the entire height of the parent Pane element, which has a set pixel value. Since, in addition to the MCL, there is also a banner inside the Pane, it is necessary to calculate the height of the container in which the MCL will be stored without an unnecessary scrollbar. For this, ResizeObserver should be used.
-  To preserve information about filters between pages, we will use location state. Since the user information page can be opened in a new tab from the duplicates page, for better UX, you should check the "referrer" property when closing the page. For security, you should also make sure the links comply with the same origin policy.

## Screenshot

https://github.com/user-attachments/assets/69075248-64b8-4beb-8481-bb77680f9a4a

## Pre-Merge Checklist

### DIFF
--- a/src/hooks/usePatronGroups/usePatronGroups.test.js
+++ b/src/hooks/usePatronGroups/usePatronGroups.test.js
@@ -20,7 +20,7 @@ const wrapper = ({ children }) => (
 
 const usergroups = [{ id: 'ug-1' }];
 
-describe('useUserAffiliations', () => {
+describe('usePatronGroups', () => {
   const mockGet = jest.fn(() => ({
     json: () => Promise.resolve({ usergroups }),
   }));
--- a/src/hooks/usePatronGroups/usePatronGroups.test.js
+++ b/src/hooks/usePatronGroups/usePatronGroups.test.js
@@ -1,0 +1,43 @@
+import {
+  QueryClient,
+  QueryClientProvider,
+} from 'react-query';
+
+import {
+  renderHook,
+  waitFor,
+} from '@folio/jest-config-stripes/testing-library/react';
+import { useOkapiKy } from '@folio/stripes/core';
+
+import usePatronGroups from './usePatronGroups';
+
+const queryClient = new QueryClient();
+const wrapper = ({ children }) => (
+  <QueryClientProvider client={queryClient}>
+    {children}
+  </QueryClientProvider>
+);
+
+const usergroups = [{ id: 'ug-1' }];
+
+describe('useUserAffiliations', () => {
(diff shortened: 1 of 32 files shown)


Terala-Priyanka:
Hey @usavkov-epam 
This feature is for LOC which do not have ECS env. So, I think we need not bother about tenantId in the fetch calls.

usavkov-epam:
> Hey @usavkov-epam This feature is for LOC which do not have ECS env. So, I think we need not bother about tenantId in the fetch calls.

I used a common pattern for common hooks that can get `tenantId`. Just because the current function doesn't need to care about tenants doesn't mean that another part of the code won't :)

usavkov-epam:
@folio-org/fe-tl-reviewers please

NikitaSedyx (approved):

alisher-epam (approved):

src/hooks/usePatronGroups/usePatronGroups.test.js
```diff
@@ -0,0 +1,43 @@
...
+const queryClient = new QueryClient();
+const wrapper = ({ children }) => (
+  <QueryClientProvider client={queryClient}>
+    {children}
+  </QueryClientProvider>
+);
+
+const usergroups = [{ id: 'ug-1' }];
+
+describe('useUserAffiliations', () => {
```
Terala-Priyanka, line 23: Rename the test suite

usavkov-epam: done

Terala-Priyanka:
Please update the permission name at https://github.com/folio-org/ui-users/blob/3f3d25b9e7c7cbb1bcf91fb8f630b882b046d162/src/components/LinkToPatronPreRegistrations/LinkToPatronPreRegistrations.js#L15, without which the feature will not be testable, after merge

usavkov-epam:
> Please update the permission name at
> 
> https://github.com/folio-org/ui-users/blob/3f3d25b9e7c7cbb1bcf91fb8f630b882b046d162/src/components/LinkToPatronPreRegistrations/LinkToPatronPreRegistrations.js#L15
> 
> , without which the feature will not be testable, after merge

Done

Terala-Priyanka (approved):
Thank you!

PR #2798, opened 2024-11-13
UIU-3267 correctly import from dompurify

zburke:
Import from `dompurify` as shown in example code; `sanitize` was never deliberatly provided as a named export. Not sure how this ever worked.

Refs [UIU-3267](https://folio-org.atlassian.net/browse/UIU-3267)


Terala-Priyanka (approved):

Terala-Priyanka (commented):
Hey @zburke 
Please push yarn.lock file too

PR #2806, opened 2024-11-21
Show all patron notice print jobs, in correct order

MikeTaylor:
We now show the top 100 patron notice print jobs, not 10.

More importantly, they are now sorted correctly (reverse chronological) on the server side, so even if the list grows past 100 and so is truncated in the UI at the first hundred, they will be the _right_ (most recent) first hundred.

Fixes UIU-3269.


nielserik (commented):
Thanks Mike, looks good to me.

PR #2817, opened 2024-12-02
UIU-3290: Fix issue with `Proxy borrower` field value

artem-blazhko:
## Purpose
 `Proxy borrower` field has incorrect value from time to time. In some cases it shows correct name, in other cases it shows "Unknown user" for the same loan. 

## Approach
Display correct Proxy borrower name for loan. In a case if user is not found display "Unknown user".

## Refs
[UIU-3290](https://folio-org.atlassian.net/browse/UIU-3290)


Dmytro-Melnyshyn (approved):

UladzislauKutarkin (approved):

zburke (approved):
The subtle bugs in the lifecycle methods of these old class-based components are so sneaky. Thank you for the clean up effort here! 

In the PR description, please describe the state of the code after the bug has been fixed, e.g. "Consistently display Proxy borrower field". You don't need to describe the bug (the bug is gone, and we care about what the code looks like now, not what it used to look like).

PR #2829, opened 2024-12-25
UIU-3302: Create/Edit a Patron Block - HTML page title

UladzislauKutarkin:
https://folio-org.atlassian.net/browse/UIU-3302

Here we added `TitleManager` for `PatronBlock` pages.
Screenshot 2025-01-13 at 12 16 13
Screenshot 2025-01-13 at 12 15 45


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 * Leverage API supported sorting of columns on pre-registrations records list. Refs UIU-3249.
 * Fix issue with `Proxy borrower` field value. Refs UIU-3290.
 * Remove duplicates from the keyboard shortcut modal. Refs UIU-3026.
+* Create/Edit a Patron Block - HTML page title. Refs UIU-3302.
```
Terala-Priyanka, line 8: Could you please reword this to "Add HTML page title to add/edit patron block page"?

Terala-Priyanka:
@UladzislauKutarkin 
Changes LGTM. But the requirement mentioned in the ticket description -
"We need to ensure that when a user is on the following setting pages that the HTML page title reads in the following format: <<<App name>> - <<selected page name>> - FOLIO."
talks about a settings page and also mentions about create/edit patron block page, which is not a setting page and the statements are contradicting. I recommend you to please have the ticket description updated.

Dmitriy-Litvinenko (approved):

PR #2848, opened 2025-01-29
UIU-3119: Pronoun Field - User Record Edit

s3fs:
Add `pronouns` field to user edit view and display `pronouns` in its header alongside with user's name if they're present.

https://folio-org.atlassian.net/browse/UIU-3119

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
       }
     ],
     "okapiInterfaces": {
-      "users": "16.3",
+      "users": "16.4",
       "configuration": "2.0",
       "permissions": "5.7",
       "login": "7.3",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@folio/users",
-  "version": "11.0.9",
+  "version": "12.0.0",
   "description": "User management",
   "repository": "folio-org/ui-users",
   "publishConfig": {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Change history for ui-users
 
-## [11.1.0] In progress
+## [12.0.0] In progress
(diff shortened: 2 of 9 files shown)


package.json
```diff
@@ -29,7 +29,7 @@
       }
     ],
     "okapiInterfaces": {
-      "users": "16.3",
+      "users": "16.4",
```
zburke, line 32: Dropping support for an interface version is a breaking change. You must update the major version of this package in `package.json` or surround this new functionality with an `<IfInterface>` conditional that specifies the minimum version.

zburke (changes requested):

package.json
s3fs: Done @zburke. Could you please check if it's correct?

zburke (approved):
Full disclosure: I saw the updated okapiInterface version in the first revision and did not read past that for my initial review. 

Now that I'm looking at the details, this work may be backwards compatible depending how mod-users is implemented. If it ignores unknown fields (as it should IMO) then you're safe and the original commit, sans the interface version update, would be just fine. If mod-users panics when it receives unknown fields, as many of our modules do 🙄 , then we do need to bump the interface version. In any case, the current implementation is correct, and if you're happy with it then so am I.

zburke:
An `okapiInterface` dependency like `16.3` is equivalent to `^16.3` (I have no idea why the syntax is different) so if the new fields were truly optional on the FE and BE, then the original PR could have been OK in the sense that "submitting a form with unsupported but optional fields would not cause an error". Showing fields that wouldn't be saved would not be a great experience, of course, but it wouldn't be an error either.

s3fs:
I have tested it locally with `"users": "16.3"` against `https://folio-dev-volaris-okapi.ci.folio.org` and it works fine. I guess it would be better to remove the version bumps I've done then? @zburke

PR #2850, opened 2025-01-31
UIU-3118: Pronoun Field - User Record View

s3fs:
Show user pronouns within the user view pane if they're present.

https://folio-org.atlassian.net/browse/UIU-3118

### DIFF
--- a/src/views/UserDetail/UserDetail.css
+++ b/src/views/UserDetail/UserDetail.css
@@ -1,0 +1,10 @@
+.NameContainer {
+  display: flex;
+  gap: 0.5rem;
+  flex-wrap: wrap;
+}
+
+.Pronouns {
+  font-weight: normal;
+  word-wrap: break-word;
+}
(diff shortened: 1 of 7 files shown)


src/views/UserDetail/UserDetail.css
```diff
@@ -0,0 +1,10 @@
+.NameContainer {
+  display: flex;
+  gap: 0.5rem;
+  flex-wrap: wrap;
+}
+
+.Pronouns {
+  font-weight: normal;
+  word-wrap: break-word;
+}
```
zburke, line 10: petty: please follow existing convention and use `nameContainer` and `pronouns` for consistency with other style-classes in this repository.

zburke (approved):

src/views/UserDetail/UserDetail.css
s3fs: I thought it was the opposite (https://github.com/folio-org/ui-users/blob/e816e9f37cd76844f5d472acf48e71198b3ae5ed/src/views/UserEdit/UserForm.css#L1).

zburke: 🙄 If there are examples in both directions, I give up. Pick your favorite and be consistent.

zburke (approved):

PR #2851, opened 2025-01-31
[UIU-3179] Add ability to create/edit role assignments for all of a user's affiliations

ryandberger:
- Fulfills [UIU-3179](https://folio-org.atlassian.net/browse/UIU-3179).
- Add the Tenant Select dropdown in User Edit screen. When changing tenant, you can add/remove roles for that user's affiliated tenant. 
- When tenant is changed, this is tracked in state and API calls are re-run to get that tenant's role IDs and role definitions.
- The object used with Final Form switches from `[roleId]` to `{tenantId1: [roleId], tenantId2: [roleId]}` in order to remember changes made to multiple tenants when saving. This also allows for `{tenantId1: [roleId], tenantId2: []}` which would remove all roles from `tenant2`.

image


aidynoJ:
LGTM!

aidynoJ (approved):

PR #2852, opened 2025-02-06
UIU-3324: Pronouns Field - Character Limit Warning

s3fs:
Introduce character limit for pronouns field; fix label.

https://folio-org.atlassian.net/browse/UIU-3324


src/components/validators/validateLength.js
```diff
@@ -0,0 +1,3 @@
+import { FormattedMessage } from 'react-intl';
+
+export const validateLength = (errorMessageKey, maxLen = 300) => value => (value?.length > maxLen ? <FormattedMessage id={errorMessageKey} /> : null);
```
BogdanDenis, line 3: Please create a constant with a name that explains the `300` value

BogdanDenis (changes requested):

PR #2863, opened 2025-02-27
[UIU-3332] New fields in due-date receipt

MikeTaylor:
New fields are included when printing a due-date receipt,
e.g. `{{item.physicalDescriptions}}`

Fixes UIU-3332.


MikeTaylor:
This is disturbing. Some of the test failures are my fault for not updating tests for the new code, but two of them happen in `master`, too.

PR #2867, opened 2025-03-04
[UIU-3333] Update detail page after editing user roles

ryandberger:
Originally, user detail page would not update users roles after they had been added/removed on the user edit screen. By invalidating the GET `/roles` API call with `react-query`, the API is refetched with the latest content when returning to the user detail screen.


aidynoJ:
LGTM. sonarqube is yelling, probably we could use async await with try catch for updateUserRoles function

aidynoJ (approved):

zburke (commented):
To @aidynoJ's point, Sonar is rightly angry because [`forEach` expects a synchronous function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach), i.e. it does not play nice with promises. Maybe it's as simple as labeling the `updateUserRoles` itself async and changing line 64 to a `for ... of`?

PR #2868, opened 2025-03-10
UIU-3335: "Preferred email communications" field clicking causes error

s3fs:
An attempt to access `option.value` within `itemToString` prop fails when `option` is `null` or `undefined`. An optional chaining operator + fallback value is added to resolve this behavior.

https://folio-org.atlassian.net/browse/UIU-3335


zburke (approved):
The code change LGTM, thank you!, but the documentation is weak. The PR title/commit title should describe the change, not the problem. "Fix error" in the changelog is not helpful; it amounts to "change code". Again, please describe what _actually_ changed so when somebody is looking at this PR a year from now, they have all the context they need and don't need to re-triage the bug in order to figure out (a) what the problem was and (b) how this change resolves it.

PR #2874, opened 2025-03-13
UIU-3347 bump jspdf to v3.0.0

JohnC-80:
## [UIU-3347](https://folio-org.atlassian.net/browse/UIU-3347)

Addresses security vulnerability within `jspdf` < 3 dependencies.
This also updates `jspdf-autotable` plug-in to v5

Noteworthy changes for the autotable plugin - 
* the `doc.autotable.previous` api replaced by `doc.lastAutoTable`.
* `title` key in columns object renamed to `header`.

Output pdf:
image


zburke (approved):
Code changes LGTM. I'm glad you noticed the jspdf-autotable API change! 🕵️ 

@s3fs, no, this does not require a major version bump. Changing direct (internal) dependencies does not change the public (external) API either in terms of what is consumed (peer-dependencies, okapi interfaces) or what is provided (public exports, routes, etc). 
* Yes, we should add a CHANGELOG entry, describing why we made these changes.
* @JohnC-80, let's generate `yarn.lock` from scratch (`rm yarn.lock; yarn install`) to de-duplicate things like `dompurify`. Since deps are ultimately resolved at the platform level this doesn't really matter, but it should help keep dependabot off our backs for a little while longer by purging some cruft (e.g. paring `dompurify` down to a single, current version instead of installing three, two of which are outdated).

JohnC-80:
@zburke @s3fs All things updated, changes logged. PDF checked for *yet another adjustment I had to make.

zburke (approved):
I'm glad you caught the additional title/header changes and sad I missed that in my first review.

s3fs (approved):

PR #2875, opened 2025-03-13
[UIU-3331, UIU-3333] Fix Save button not always enabled when saving roles to multiple affiliations; Detail page not immediately updated after editing user roles

ryandberger:
- Fixes [UIU-3331](https://folio-org.atlassian.net/browse/UIU-3331).
- Fixes [UIU-3333](https://folio-org.atlassian.net/browse/UIU-3333).
- Previously, when adding/removing roles in the User edit screen and then switching affiliations, the save button would grey out and would forget changes made to the first affiliation. 
  - This was caused by the object resetting whenever existing roles were fetched from the API for a new affiliation. 
  - Working with @aidynoJ, we determined that trying to maintain changed roles in addition to roles fetched from an API was too confusing to manage and pass to `react-final-form` to determine if changes had been made by the user. The solution is is first fetch roles from all affiliations to set the baseline in `initialRoleIds` and add/subtract in `assignedRoleIds` so `react-final-form` can make a proper comparison.
- Additionally, PR fixes UIU-3333. Refactoring the `updateUserRoles` function by collecting promises into a list and awaiting them all at once resolves the issue [UIU-3333](https://folio-org.atlassian.net/browse/UIU-3333). The reason is that async/await does not work properly within `forEach` iterations. Instead of waiting for the promises to resolve, the loop continues execution, causing the edit window to close prematurely. This results in stale role data appearing in the role details pane.


CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Change history for ui-users
 
+## [12.0.1] In Progress
+
+* Fix bug where save button is not always enabled when saving roles to multiple affiliations. Refs UIU-3331.
+* Fix detail page not immediately updated after editing user roles. Refs UIU-3333.
```
zburke, line 6: This bug is gone now, so I don't care about it. Tell me what the code does now that it's working correctly, not about a problem you used to have, e.g. "Replace `forEach` with `map` in order to await promises and then render details".

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Change history for ui-users
 
+## [12.0.1] In Progress
+
+* Fix bug where save button is not always enabled when saving roles to multiple affiliations. Refs UIU-3331.
```
zburke, line 5: This bug is gone now, so ...

zburke (approved):
PR description is very helpful in terms of describing how these changes address the bugs described in the tickets; thank you!

Lint is kinda mad at you. Please address this before you merge.

aidynoJ (approved):

PR #2885, opened 2025-03-21
WIP FOR DISCUSSION: Conditional Load for external dependency

EthanFreestone:
## Purpose
A recent PR brought 'ui-service-interaction' in as a peer dependency to ui users. In normal operation for FOLIO this would simply resolve itself easily, as both users and ui-service-interaction are parts of the "flower release".

However the intention with NumGen (the feature implemented by UIU-2729) was never intended to bring extra locking dependencies into the frontend and thus entire build process, rather act as an OPTIONAL feature set for modules to implement how they see fit.

Due to constraints when building `ui-service-interaction` the implementations themselves did not end up being completed by me beyond a very small proof-of-concept, and so some of the architectural nuance ended up being lost.

## Approach
I had envisioned something similar to "Pluggable", but instead as a new pattern where modules could surface "public" components to be brought in by other modules should they wish. The "peer dependency" is equivalent to an optional okapi dependency (whichh should ALSO  be defined), and thus it should be treated as such.

To that end I propose something akin to Pluggable: ConditonalLoad (Name absolutely up for debate, I am not the best at naming things)

This works utilising Lazy loading to catch imports at runtime, meaning that functionality for flower releases need not be affected, but any implementations of Users without wanting to bring in service interaction and the _required_ interface dependencies of that can do so.

#### TODOS and Open Questions
*Does this belong in stripes-core?*
My gut says that such a feature absolutely belongs in stripes core, this PR simply acts as a place to begin that conversation.

*Is there anything else required to support this "public" component pattern?*
I know that some developers, especially @zburke have had questions about this pattern from the very start. However it has now been ratified by the tech council and so the genie is basically out of the bottle. It is definitely on me to a large extent that the fallout of this design was not thoroughly considered with measures like this put in place, but I believe this 100% MUST be considered for functionality such as UIU-2729 in Trillium in such a key module as users

## Pre-Merge Checklist
- [X ] I've added appropriate record to the CHANGELOG.md
  - [X ] Code coverage on new code is 80% or greater
  - [X ] Duplications on new code is 3% or less
  - [ X] There are no major code smells or security issues
  - [X ] There are no breaking changes in this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 13.0.0 IN PROGRESS
 * Collect coverage from unit tests. Refs UIU-3356.
 * Use number generator for barcode. Refs UIU-2729.
+  * Implemented optional loading of Number generator. Refs UIU-2729
 
 ## [12.1.0](https://github.com/folio-org/ui-users/tree/v12.1.0) (2025-03-18)
 [Full Changelog](https://github.com/folio-org/ui-users/compare/v12.0.0...v12.1.0)
--- a/src/components/ConditionalLoad/ConditionalLoad.js
+++ b/src/components/ConditionalLoad/ConditionalLoad.js
@@ -1,0 +1,48 @@
+import { lazy, Suspense } from 'react';
+import PropTypes from 'prop-types';
+
+import { Loading } from '@folio/stripes/components';
+
+// This effectively becomes analogous to Pluggable, except happening at the dependency load stage -- should almost certainly belong to stripes-core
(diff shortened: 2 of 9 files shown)


EthanFreestone:
Note... this doesn't seem to work quite as expected currently, I think the approach is still up for debate, but something is breaking WRT the dynamic import of modules in the workspace :(

EthanFreestone:
Of course, the alternative is to use <IfInterface> here, and handle this purely as a peer dep. That might be less disruptive and more palatable?

I thought it worth a conversation though, at least

PR #2888, opened 2025-03-31
UIU-3314: "Affiliations" select in Role/Permission accordion should not be displayed on patron user details pane

UladzislauKutarkin:
https://folio-org.atlassian.net/browse/UIU-3314

Here we fixed issue with `"Affiliations" select `with `Patron user type`

### DIFF
--- a/src/components/RenderRoles/RenderRoles.js
+++ b/src/components/RenderRoles/RenderRoles.js
@@ -37,6 +37,7 @@ class RenderRoles extends React.Component {
         listInvisiblePerms: PropTypes.bool,
       }).isRequired,
     }).isRequired,
+    isAffiliationsVisible: PropTypes.bool.isRequired,
   };
 
   static defaultProps = {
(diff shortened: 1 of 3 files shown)


zburke (approved):
This works, but is it worth prop-drilling `isAffiliationsVisible` when we could just as easily import `.../util/isAffiliationsEnabled` in `RenderRoles` and call it directly there? 

It would be nice to see an updated test that demonstrates different content depending on whether this function returns true or false.

src/components/RenderRoles/RenderRoles.js
```diff
@@ -89,7 +91,7 @@ class RenderRoles extends React.Component {
       >
         <IfConsortium>
           <IfConsortiumPermission perm="consortia.user-tenants.collection.get">
-            {Boolean(affiliations?.length) && (
+            {Boolean(affiliations?.length) && isAffiliationsVisible && (
```
Dmytro-Melnyshyn, line 94: > Workaround: refresh the page

Why does refreshing the page solve this issue?

UladzislauKutarkin: Because after switching the user's type, an array of affiliations starts caching. That's why I added an additional condition as it's done in `Affiliation accordion`
https://github.com/folio-org/ui-users/blob/master/src/views/UserDetail/UserDetail.js#L763

PR #2895, opened 2025-04-16
UIU-3352: Specify timezone to fix incorrect date display

s3fs:
https://folio-org.atlassian.net/browse/UIU-3352

Also, post-release changelog updates


zburke (approved):
Code changes LGTM, though see also [<FormattedUTCDate>](https://github.com/folio-org/stripes-components/tree/master/lib/FormattedUTCDate) which is a shortcut to what you have done here since it is a common scenario. 

Personally, I don't like seeing the unrelated CHANGELOG updates being carried in this PR, and I disagree with way they are being made. UIU-3266 was implemented first on the main branch in v13 (so it belongs there) and then backported to v11.0.12 (so it belongs there too). Removing the v13.0.0 entry is misleading.

zburke (approved):

PR #2901, opened 2025-04-25
UIU-3318: [Follow-up] Fix incorrect date display due to locale parsing issues

s3fs:
Applying 'L' locale-aware formatting didn't help with fixing moment parsing the date as invalid for some reason. A call like `moment('25/04/2025', 'L')` still resulted in `Invalid date`.

It seems that removing the 'L' specifier altogether resolves this (tested with a few locales like British English, Spanish, Czech).

https://folio-org.atlassian.net/browse/UIU-3318

@zburke could you please share an opinion on whether this is the right approach?


zburke (commented):
Without any comments in the code or examples of what is supposed to be happening here, it's hard to say what the "correct" behavior should be. 

In this particular case, I don't understand how the locale is relevant to calculating an expiration date. I will add that if you're going to bother changing code where `moment` is involved, you are highly highly encouraged to convert that code to use `dayjs` as moment has been deprecated both officially and within FOLIO. See details in [UIU-3334](https://folio-org.atlassian.net/browse/UIU-3334) and [STRIPES-967](https://folio-org.atlassian.net/browse/STRIPES-967).

s3fs:
Thanks for the input @zburke! Unfortunately I'm doing this on quite a short notice won't be able to switch to `dayjs` in scope of this PR.

Me and QAs did a bit of testing on this and it seems to work fine. So you're right, the locale within the calculations wasn't needed in the first place.

JohnC-80 (approved):
Thanks for getting the conversion in there! 👏

Dmytro-Melnyshyn (approved):

PR #2907, opened 2025-05-20
UICIRC-1232: Add missing users view sub-permission to view settings (users) patron blocks template

mukhiddin-yusuf:
## Purpose
Requesting `GET /users` fails during viewing settings (users) patron blocks template

## Approach
Add missing `users.collection.get` sub-permission for `ui-users.settings.patron-blocks.view`

## Refs
[UICIRC-1232](https://folio-org.atlassian.net/browse/UICIRC-1232)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1119,7 +1119,8 @@
         "subPermissions": [
           "ui-users.settings.conditions.view",
           "ui-users.settings.limits.view",
-          "ui-users.settings.patron-block-templates.view"
+          "ui-users.settings.patron-block-templates.view",
+          "users.collection.get"
         ],
         "visible": true
       },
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1119,7 +1119,8 @@
         "subPermissions": [
           "ui-users.settings.conditions.view",
           "ui-users.settings.limits.view",
-          "ui-users.settings.patron-block-templates.view"
+          "ui-users.settings.patron-block-templates.view",
+          "users.collection.get"
```
Dmytro-Melnyshyn, line 1123: This permission has already been added to `ui-users.settings.patron-block-templates.view`. Is the permission issue still reproducible?

PR #2908, opened 2025-05-21
UIU-3383: Refactor `updateKeycloakUser` to use `ky.put` instead of `api.put`, because updating current user data requires setting `x-okapi-tenant` to the current tenant.

Dmytro-Melnyshyn:
## Description
When editing user data and assigning roles to a non-current tenant, after saving the record, the error toast message shows up, the assigned roles are saved, but other changes are not saved. The problem is that the request to save user data is using the wrong tenant. It takes the tenant from the last selected tenant during role selection. To fix this, we just need to use the current tenant.

## Issues
[UIU-3383](https://folio-org.atlassian.net/browse/UIU-3383)

## Screencasts

https://github.com/user-attachments/assets/f0be3d60-7114-4f29-8aa1-fe02ee6b4e90

## Pre-Merge Checklist
  - [ ] Code coverage on new code is 80% or greater
  - [ ] Duplications on new code is 3% or less
  - [ ] There are no major code smells or security issues
  - [ ] There are no breaking changes in this PR.


zburke (approved):
IIUC, the issue here is that the `updateUserRoles` API request needs to have its `x-okapi-tenant` value set to the tenant in which that role is being assigned but the `updateKeycloakUser` API request needs to have its `x-okapi-tenant` value set to the current tenant. The bug is that `updateKeycloakUser` may get the wrong tenant in some circumstances, causing the request to fail, resulting in an error toast. 

If that's all correct, great! Please add a comment to the source around both `updateUserRoles` and `updateKeycloakUser` functions explaining why their requests need to monkey with `x-okapi-tenant` at all, since this is unusual. IOW, the code changes look good, but we need to improve the documentation.

PR #2950, opened 2025-08-12
[UIU-3426] Fix User Edit workflow for making edits to non-Keycloak user

ryandberger:
- Addressing comment made by QA in [UIU-3426](https://folio-org.atlassian.net/browse/UIU-3426?focusedCommentId=272678)
- Previously Edit screen remained open. Now it properly returns to User Detail page with updated data by calling `onFinish()`. This is only explicitly called when not assigning roles since `setIsCreateKeycloakUserConfirmationOpen()` calls `onFinish()` when dismissing the dialog.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,7 @@
 * Only show Keycloak user record confirmation when assigning roles to a non-Keycloak record. Refs UIU-3426.
 * Add 'Minor' column to Patrons Pre-Registration List. Refs UIU-3425.
 * *BREAKING* Update the user record view page to display custom fields under the Fees/fines, Loans, or Requests accordions as configured in settings. Refs UIU-3130.
+* Fix User Edit workflow for making edits to non-Keycloak user. Previously Edit screen remained open. Now it properly returns to User Detail page with updated data. Refs UIU-3426.
 
 ## [12.1.7] (https://github.com/folio-org/ui-users/tree/v12.1.7) (2025-06-03)
 [Full Changelog](https://github.com/folio-org/ui-users/compare/v12.1.6...v12.1.7)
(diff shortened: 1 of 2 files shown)


zburke (approved):
This looks like a follow-up to PR #2948, correct? OK, that makes sense. 👌

Unfortunately, the logic here is hard to follow. Your comment "`setIsCreateKeycloakUserConfirmationOpen()` calls `onFinish()` when dismissing the dialog" isn't quite right I think since `setIsCreateKeycloakUserConfirmationOpen` is returned from `useState`. Maybe you mean "`confirmCreateKeycloakUser()` calls `onFinish()`"? That would make sense in the context of this bug -- i.e. we were missing an `onFinish()` call and this PR adds it -- but the purpose of this wrapper and how all its pieces work is very difficult to follow. e.g. why do we even have state for a confirmation-modal here given this component does not, itself, contain a confirmation-modal? Why do we call `onFinish()` in multiple places given a form is only submitted once?

I'm ticking "approve" here based on the fact that this PR gets us to the correct behavior in the UI, but the component itself is kind of a mess. At minimum we need some documentation about what this wrapper does and all these callback props work.

Dmytro-Melnyshyn (approved):

ryandberger:
> This looks like a follow-up to PR #2948, correct? OK, that makes sense. 👌
> 
> Unfortunately, the logic here is hard to follow. Your comment "`setIsCreateKeycloakUserConfirmationOpen()` calls `onFinish()` when dismissing the dialog" isn't quite right I think since `setIsCreateKeycloakUserConfirmationOpen` is returned from `useState`. Maybe you mean "`confirmCreateKeycloakUser()` calls `onFinish()`"? That would make sense in the context of this bug -- i.e. we were missing an `onFinish()` call and this PR adds it -- but the purpose of this wrapper and how all its pieces work is very difficult to follow. e.g. why do we even have state for a confirmation-modal here given this component does not, itself, contain a confirmation-modal? Why do we call `onFinish()` in multiple places given a form is only submitted once?
> 
> I'm ticking "approve" here based on the fact that this PR gets us to the correct behavior in the UI, but the component itself is kind of a mess. At minimum we need some documentation about what this wrapper does and all these callback props work.

Yep, it's a follow-up. `onFinish()` is called in multiple places since they are the final step in each code path. I agree this is confusing and ideally should be done as a final step in all cases (except cancel). I need to merge in the interest of time for this ticket, but I agree this logic could use a good refactor to help make it way less confusing. 😄

PR #2969, opened 2025-09-08
[UIU-3457] Avoid indirection through deleted item in loan

MikeTaylor:
Use `get(entity, 'item.FIELD')` for each field
Instead of `entity?.item.FIELD` as before.

(We could alternatively have used a second `?.`.)

### DIFF
--- a/src/hooks/usePrintData/usePrintData.js
+++ b/src/hooks/usePrintData/usePrintData.js
@@ -59,7 +59,7 @@ export const mapEntityToTemplate = (entity, type, formatTime, patronGroup) => {
       'item.editions': get(entity, 'item.editions')?.join('/'),
       'item.physicalDescriptions': get(entity, 'item.physicalDescriptions')?.join('/'),
       'item.instanceHrid': get(entity, 'item.instanceHrid'),
-      'item.seriesStatements': entity?.item.seriesStatements?.join(' / '),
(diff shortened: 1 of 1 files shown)


src/hooks/usePrintData/usePrintData.js
```diff
@@ -53,14 +53,14 @@ export const mapEntityToTemplate = (entity, type, formatTime, patronGroup) => {
...
-      'item.editions': entity?.item.editions?.join('/'),
-      'item.physicalDescriptions': entity?.item.physicalDescriptions?.join('/'),
-      'item.instanceHrid': entity?.item.instanceHrid,
+      'item.accessionNumber': get(entity, 'item.accessionNumber'),
+      'item.administrativeNotes': get(entity, 'item.administrativeNotes')?.join('/'),
+      'item.datesOfPublication': get(entity, 'item.datesOfPublication')?.join('/'),
+      'item.editions': get(entity, 'item.editions')?.join('/'),
+      'item.physicalDescriptions': get(entity, 'item.physicalDescriptions')?.join('/'),
+      'item.instanceHrid': get(entity, 'item.instanceHrid'),
       'item.seriesStatements': entity?.item.seriesStatements?.join(' / '),
```
Dmitriy-Litvinenko, line 62: Could you please also use `get` for `seriesStatements`?

PR #3032, opened 2026-04-10
UIU-3548 Implement "select all" as "select visible"

zburke:
In the "Select user roles" SAS modal, ticking the "Select all" checkbox in the first column header must follow existing conventions, select only the rows that are visible in the results pane, not all rows.

Previously, conducting a search that returned, say, 10 rows and then ticking the column-header's box would select all rows rather than only the 10 returned by the search. This was misleading and inconsistent with other implementations where "select all" is implemented "select all that are visible".

Refs [UIU-3548](https://folio-org.atlassian.net/browse/UIU-3548)

https://github.com/user-attachments/assets/4a8a6b76-0ebc-4967-b16d-9b7e124ef7e1

### DIFF
--- a/src/components/EditSections/EditUserRoles/components/UserRolesModal/UserRolesModal.js
+++ b/src/components/EditSections/EditUserRoles/components/UserRolesModal/UserRolesModal.js
@@ -55,9 +55,13 @@ export default function UserRolesModal({ isOpen,
 
   const toggleRoleList = (checked, roleList) => {
     if (checked) {
-      setAssignedRoleIds({ ...assignedRoleIds, [tenantId]: roleList?.map(role => role.id) });
+      const currentIds = assignedRoleIds[tenantId] || [];
+      const currentPlusFiltered = new Set([...currentIds, ...roleList.map(role => role.id)]);
+      setAssignedRoleIds({ ...assignedRoleIds, [tenantId]: Array.from(currentPlusFiltered) });
     } else {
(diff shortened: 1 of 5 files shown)


src/components/EditSections/EditUserRoles/components/UserRolesModal/UserRolesModal.js
Dmytro-Melnyshyn, line 69: 1. Select all with a filter selects only filtered roles
2. Select all with a filter preserves pre-existing selections outside the filter
3. Uncheck all with a filter only deselects filtered roles

```
const filteredRoles = getFilteredRoles();

const toggleAllRoles = (checked) => {
    if (checked) {
      const filteredIds = filteredRoles.map(role => role.id);
      const existingIds = assignedRoleIds[tenantId] || [];
      const mergedIds = [...new Set([...existingIds, ...filteredIds])];
      setAssignedRoleIds({ ...assignedRoleIds, [tenantId]: mergedIds });
    } else {
      const filteredIdSet = new Set(filteredRoles.map(role => role.id));
      const remainingIds = (assignedRoleIds[tenantId] || []).filter(id => !filteredIdSet.has(id));
      setAssignedRoleIds({ ...assignedRoleIds, [tenantId]: remainingIds });
    }
  };
```

ryandberger (approved):
Looks good! Thanks @JohnC-80 for also adding logic to handle unselect all when filtering roles, so that unfiltered roles are not deselected too. Thanks for adding test cases to ensure correct functionality.

PR #3034, opened 2026-04-16
UIU-3385: Add Settings components for Version History in Users

slaemmer:
[UIU-3385](https://folio-org.atlassian.net/browse/UIU-3385)

## Purpose
The Users app handles more sensitive personal data than most other FOLIO apps. Before rolling out Version History in Users, institutions must be able to configure (or fully disable) what is tracked, in order to comply with local and national data protection regulations.                                                                                          
                                                                                                                              
This PR implements the initial settings UI under Settings > Users > General > Version history, giving Systems Librarians control over:                                                                                                                
- whether a version history is created at all                                                                                
- how long history is retained                                                                                               
- whether the source of changes is anonymized                                                                                
- which fields (e.g. contact info, custom fields) are excluded from tracking

## Approach
A new `VersionHistorySettings` entry is added at the end of the `settingsGeneral` list in `src/settings/sections.js`, gated behind a new `ui-users.settings.versionHistory.view` permission.                                                                      
                                                                                                                            
The UI consists of three components: 
- `VersionHistoryForm` renders the full settings form. Fields that require version history to be    
enabled are wrapped in a Tooltip explaining how to activate them. 
- `VersionHistoryWarningModal` is shown before saving whenever the new settings would delete previously stored data / not store all data in a way that differs from before. 
- `VersionHistorySettings` is the top-level container wiring these together with data fetching and custom fields loading.                                                                                                       
                                                                                                                            
On the data layer, `useAuditSettingsQuery` fetches from mod-audit and `useVersionHistorySettings` issues individual PUT requests per changed setting key. The utility functions in `utils.js` handle conversion between the API's flat settings array and form values, diffing current vs. submitted values to minimise API calls, warning detection, and form validation.                                                                                                                  
                                                                                                                            
Two new permissions are declared in `package.json`: a view permission (read access to the mod-audit group settings) and an edit permission (extends view with per-key PUT permissions for each of the four setting keys: enabled, records.retention.period, anonymize, excluded.fields).

## Pre-Merge Checklist
    - Added two now permissions, additive only.
    - Added optional dependency on audit-config interface. Setting page is not shown when interface is not provided

~~**Should be merged after the Trillium release is fixed with CHANGELOG at the correct location.**~~

**Should be merged at the same time as the version history display PR for UIU-3388 as both together complement each other (user facing vs settings).**


s3fs (approved):

src/settings/VersionHistorySettings/VersionHistoryForm.js
```diff
@@ -0,0 +1,214 @@
...
+  return (
+    <form onSubmit={handleSubmit} className={css.formWrapper}>
+      <Pane
+        defaultWidth="fill"
+        paneTitle={<FormattedMessage id="ui-users.settings.versionHistory" />}
+        footer={renderFooter()}
+      >
+        <Label>
+          <FormattedMessage id="ui-users.settings.versionHistory.createVersionHistory" />
+        </Label>
```
Dmytro-Melnyshyn: Image

<br/>

`Label` is semantically tied to form controls; using it as a standalone section heading is incorrect and will announce incorrectly to screen readers. The `Headline` may be appropriate.

slaemmer: Thanks for catching that, missed it.

Switched it to `RadioButtonGroup`'s label prop, which renders it as `legend` inside `fieldset`. Feels closer to the intent than a separate `Headline` above. 

However, stripes styles `legend.groupLabel` as small/regular weight per default, so the visual distinction disappeared. Restored it via a small CSS override on the legends content. Not happy about adding custom CSS here. Happy to use `Headline` if that's preferable to custom CSS or take any other better suggestion if there is a better stripes-native way!

Dmytro-Melnyshyn: > Switched it to RadioButtonGroup's label prop, which renders it as legend inside fieldset. Feels closer to the intent than a separate Headline above.

Is the "Create version history" related only to radio buttons? If not, then it shouldn't be part of `RadioButtonGroup`.

slaemmer: Yes - afaik it is supposed to read as "Create version history" "Never" / "Retain indefinitely" / "Retain for a set duration". "Anonymize the change source" should be separate.

src/settings/VersionHistorySettings/VersionHistoryForm.js
```diff
@@ -0,0 +1,215 @@
+import React, { useMemo } from 'react';
```
elsenhans: I think, `import React` is not needed anymore.

Dmytro-Melnyshyn (approved):

PR #3035, opened 2026-04-16
UIU-3388: Add user version history to user detail view

slaemmer:
[UIU-3388](https://folio-org.atlassian.net/browse/UIU-3388)

## Purpose
Staff users need to be able to track changes made to user records in order to investigate irregularities or answer queries. This PR adds a version history panel to the user detail view, displaying a chronological list of all recorded changes within the configured retention period, along with their source and a detailed diff per version.

## Approach

A clock icon button (`UserVersionHistoryButton`) is added to the UserDetail pane header, to the right of the tags icon. It only renders when the user has `ui-users.versionHistory.view` and audit is enabled in settings, so it is automatically hidden when Version History is set to "Never". When the panel is open, the Actions button, tags icon, and clock icon are all disabled; Actions is replaced with a tooltip-wrapped disabled surrogate ("To enable Actions, close Version history.").       
                                                                                                                            
The panel itself (`UserVersionHistory`) delegates rendering to `AuditLogPane` from `@folio/stripes/components`, following the same pattern established by ui-inventory. Audit data is fetched via `useUserAuditDataQuery`, cursor-paginated by `eventTs`). `useUserVersionHistory` batch-fetches user display names and maps the raw audit items to the shape `AuditLogPane` expects, filtering out metadata-only updates (no user-visible change) and stripping internal fields (`createdDate`, `updatedByUserId`, etc.) from diffs. `useUserVersionHistoryFormatters` builds the field label map (including custom fields), action labels, per-field value renderers (patron group, contact type, department, dates, custom field types), and the card summary item formatter.          
                                                                                                                            
A new permission `ui-users.versionHistory.view` is declared, bundling `ui-users.view` and the required mod-audit read permissions. audit-data: 1.0 is added as a required Okapi interface.

## Pre-Merge Checklist
    - Added a new permissions, additive only.
    - Added optional dependency on audit-config and audit-user interface. Version History Button is not shown when interface is not provided

~~**Should be merged after the Trillium release is fixed with CHANGELOG at the correct location.**~~

**Should be merged at the same time as the version history settings PR for UIU-3385 as both together complement each other (user facing vs settings).**


s3fs (approved):

package.json
```diff
@@ -202,6 +204,18 @@
         ],
         "visible": true
       },
+      {
+        "permissionName": "ui-users.versionHistory.view",
```
Dmytro-Melnyshyn: Let's rename it to `ui-users.settings.version-history.view` to follow the [Permissions naming convention](https://folio-org.atlassian.net/wiki/spaces/FOLIJET/pages/156368925/Permissions+naming+convention)

src/views/UserDetail/components/UserVersionHistory/UserVersionHistoryButton.js
```diff
@@ -0,0 +1,56 @@
...
+} from '@folio/stripes/components';
+import { useStripes } from '@folio/stripes/core';
+
+import { useIsAuditEnabled } from '../../../../hooks/useAuditSettingsQuery';
+
+const UserVersionHistoryButton = ({ disabled, onClick }) => {
+  const intl = useIntl();
+  const stripes = useStripes();
+  const hasPermission = stripes.hasPerm('ui-users.versionHistory.view');
+  const { isEnabled, isLoading, isError } = useIsAuditEnabled();
```
Dmytro-Melnyshyn: Let's prevent settings to be called if the permission is missing
`const { isEnabled, isLoading, isError } = useIsAuditEnabled({ enabled: !!hasPermission });`

src/views/UserDetail/UserDetail.js
```diff
@@ -770,8 +819,8 @@ class UserDetail extends React.Component {
                   {getFullName(user)}
                 </span>
               }
-              actionMenu={this.getActionMenu(get(user, 'barcode', ''), isProfilePictureFeatureEnabled)}
-              lastMenu={this.renderDetailsLastMenu(user)}
+              actionMenu={isVersionHistoryOpen ? undefined : this.getActionMenu(get(user, 'barcode', ''), isProfilePictureFeatureEnabled)}
+              lastMenu={this.renderDetailsLastMenu(user, isVersionHistoryOpen)}
```
Dmytro-Melnyshyn: This currently uses a workaround that renders a fake, disabled action menu. A cleaner approach is to add two props to `Pane` and update stripes-components so the header can render a disabled action menu with a tooltip. Example of props for `Pane`:

```
actionMenuDisabled={isVersionHistoryOpen}
actionMenuTooltip={isVersionHistoryOpen
  ? <FormattedMessage
    id="ui-users.versionHistory.actionsDisabled.tooltip"
    values={{ b: chunks => {chunks} }}
    />
  : null
}
```
Here are the corresponding [changes](https://github.com/folio-org/stripes-components/commit/3c47a18f78f0e1d1a3eae19fb4803eeff64c66d1) for stripes-components. Let's open a PR for this.

src/views/UserDetail/components/UserVersionHistory/UserVersionHistory.js
```diff
@@ -0,0 +1,70 @@
...
+import PropTypes from 'prop-types';
+
+import { AuditLogPane } from '@folio/stripes/components';
+import { useStripes } from '@folio/stripes/core';
+
+import useUserAuditDataQuery from '../../../../hooks/useUserAuditDataQuery';
+import useUserVersionHistory from '../../../../hooks/useUserVersionHistory';
+import useUserVersionHistoryFormatters from './useUserVersionHistoryFormatters';
+
+const VIEW_USER_PERMISSION = 'users.item.get';
```
Dmytro-Melnyshyn: If the user has opened the version history pane, then they already have permission to view users, as details page requires it.

src/index.js
```diff
@@ -155,6 +163,11 @@ class UsersRouting extends React.Component {
 
     return (
       <>
+        <IfInterface name="audit-user">
+          <IfInterface name="audit-config">
+            <PrefetchAuditSettings />
+          </IfInterface>
+        </IfInterface>
```
Dmytro-Melnyshyn: The loading indicator for the version history icon isn't considered a bad UX, and this setting is only needed for two pages, but will be triggered in all ones. The user may not even open the user details page in which case the request is redundant. let's remove this prefetching.

src/hooks/useUserAuditDataQuery/useUserAuditDataQuery.js
```diff
@@ -0,0 +1,50 @@
...
+import { useMemo } from 'react';
+
+import {
+  useNamespace,
+  useOkapiKy,
+} from '@folio/stripes/core';
+
+import { getNextPageParam } from './userAuditDataQueryUtils';
+
+const useUserAuditDataQuery = (userId) => {
```
Dmytro-Melnyshyn, line 11: When the `Pronouns` or `Preferred email communications` fields are updated, it isn't reflected in the version history pane. Is this expected?

src/hooks/useUserVersionHistory/useUserVersionHistory.js
```diff
@@ -0,0 +1,39 @@
...
+const useUserVersionHistory = (data) => {
+  const { formatMessage, formatDate } = useIntl();
+
+  const userIds = useMemo(
+    () => (data?.length ? uniq(data.map(entry => entry.performedBy).filter(Boolean)) : []),
+    [data],
+  );
+
+  const { users, isFetched } = useUsersQuery(
+    { query: userIds.length ? buildQueryByIds(userIds) : undefined },
```
Dmytro-Melnyshyn: I think too many updates from different staff for a user record could cause the `query` to become too long, resulting in an "HTTP 414 - URI Too Long" error. Let's use a batch request - `batchRequest` function (I think the chunk size could be increased).

src/utils/renderDate.js
```diff
@@ -0,0 +1,7 @@
+import { FormattedUTCDate, NoValue } from '@folio/stripes/components';
+
+const renderDate = dateValue => {
+  return dateValue ? <FormattedUTCDate value={dateValue} /> : <NoValue />;
```
Dmytro-Melnyshyn, line 4: Should timezone be always fixed to UTC? This may be incorrect behavior - users would likely expect to see timestamps in their local timezone.

src/views/UserDetail/UserDetail.test.js
```diff
@@ -345,58 +359,6 @@ describe('UserDetail', () => {
...
-      });
-    });
-
-    describe('when roles interface is present', () => {
-      it('shows roles accordion', async () => {
-        stripes.hasInterface = () => true;
-        renderUserDetail(stripes);
-        expect(screen.queryByText('Permissions accordion')).not.toBeTruthy();
-        expect(screen.getByText('Roles accordion')).toBeTruthy();
-      });
```
Dmytro-Melnyshyn, line 362: Why are these tests removed?

src/views/UserDetail/UserDetail.js
slaemmer: Right, that would be cleaner. Since you've already got a diff prepared - would you prefer to open the stripes-components PR, or should I? Happy either way, just want to avoid us both opening one. Once it is released I'll wire it in here and drop the workaround.

One thing I noticed on the test branch: the tooltip wraps the action menu whenever `actionMenuTooltip` is truthy, regardless of `actionMenuDisabled`. With a message like "To enable actions, close version history," that would also show on a fully enabled menu, which seems off. Could we couple the two so the API can't be misused (e.g. a single `actionMenuDisabledTooltip` prop that implies disabled and provides the tooltip text)?

src/utils/renderDate.js
slaemmer: `renderDate` is used only for date-only fields. As far as I understand forcing UTC is intentional for these in users in general: rendering them in the user's local timezone would shift them off-by-one-day for users in negative-UTC zones (e.g. a `dateOfBirth` of 2000-01-01T00:00:00Z would display as 12/31/1999 in PST). This matches how the user details page already renders the same fields (e.g. `ExtendedInfo.js`). Or did I get sth wrong here?

src/views/UserDetail/UserDetail.test.js
slaemmer: These tests were duplicates. I did not remove but deduplicate them. However, it probably is not in the scope of the PR, so should I revert the removal?

src/views/UserDetail/components/UserVersionHistory/UserVersionHistory.js
```diff
@@ -0,0 +1,70 @@
...
+      versions={versionsWithSource}
+      onClose={onClose}
+      isLoadMoreVisible={hasNextPage}
+      handleLoadMore={fetchNextPage}
+      isLoading={isLoadingMore}
+      isInitialLoading={isInitialLoading}
+      fieldLabelsMap={fieldLabelsMap}
+      fieldFormatter={fieldFormatter}
+      actionsMap={actionsMap}
+      totalVersions={versionsWithSource.length}
```
Dmytro-Melnyshyn: `totalVersions` in paneSub should probably reflect the total count of all versions from the backend - not just the number of currently loaded/visible cards.

src/utils/renderDate.js
Dmytro-Melnyshyn: I agree

src/views/UserDetail/UserDetail.test.js
Dmytro-Melnyshyn: Ahh, then yes, let's remove duplicates.

src/views/UserDetail/UserDetail.js
Dmytro-Melnyshyn: I agree with coupling the tooltip to the disabled state. I'd keep `actionMenuDisabled` as the generic flag, but rename `actionMenuTooltip` to `actionMenuDisabledTooltip` and only render it when disabled. That keeps the API hard to misuse while still allowing a disabled menu without a tooltip if another consumer ever needs that.

Dmytro-Melnyshyn: >  Once it is released I'll wire it in here and drop the workaround.

Don't wait for a release - please open a PR on stripes-components containing the Pane header props change. Once that PR is approved and merged, update ui-users to consume the new props and drop the workaround. When ui-users is ready for release, I'll point `@folio/stripes` to the latest version so it will consume those changes from stripes-components.

src/views/UserDetail/components/UserVersionHistory/UserVersionHistory.js
slaemmer: I thought about that as well but that would mean that audits that contain only non-visible changes (like metadata) are also counted but not visible. This would result in a missmatch. Calculating visible audit logs in the backend feels off and so does pre-fetching all in the frontend and filtering them to retrieve a full count. 

If you got other suggestions or think one approach is better than the current one I don't mind changing it though.

src/views/UserDetail/UserDetail.js
slaemmer: Created a ticket [STCOM-1504](https://folio-org.atlassian.net/browse/STCOM-1504) and a respective [pull/2540](https://github.com/folio-org/stripes-components/pull/2540). Would appreciate you having a look whether both look good to you.

src/hooks/useUserAuditDataQuery/useUserAuditDataQuery.js
slaemmer: No, this is definitely not expected. I looked into it and the cause was an old user schema in mod-audit that did not include the fields. I opened a [bug ticket](https://folio-org.atlassian.net/browse/MODAUD-308) and a [PR](https://github.com/folio-org/mod-audit/pull/257).

Also during this I discovered that I missed adding the `preferredEmailCommunication` label to the formatter. Added it. So for this PR there should be nothing to do anymore and the field changes are shown with the mod-audit changes.

src/hooks/useAuditSettingsQuery/useAuditSettingsQuery.js
```diff
@@ -0,0 +1,31 @@
...
+const DEFAULT_SETTINGS = [];
+
+const useAuditSettingsQuery = (options = {}) => {
+  const ky = useOkapiKy();
+  const [namespace] = useNamespace({ key: AUDIT_SETTINGS_NAMESPACE_KEY });
+
+  const { data, isLoading, isError } = useQuery(
+    [namespace, AUDIT_USER_SETTING_GROUP],
+    () => ky.get(AUDIT_SETTINGS_ENDPOINT).json(),
+    { staleTime: AUDIT_SETTINGS_STALE_TIME_MS, retry: false, ...options },
```
Dmytro-Melnyshyn, line 21: `enabled` should be a boolean value, and [hasPerm](https://github.com/folio-org/stripes-core/blob/6b191a109936ef47605e068c3ac539d56592739c/src/Stripes.js#L118) can be undefined. Let's use `{ enabled: Boolean(enabled) }`.

src/views/UserDetail/components/UserVersionHistory/UserVersionHistory.js
Dmytro-Melnyshyn: I understand the concern about the mismatch, but I'd still prefer `totalRecords` from the backend - it matches what ui-inventory does and what the spec expects (a stable count in the header). The visible increment on every 'Load more' click is more confusing UX than a count that may occasionally include metadata-only events. If the mismatch becomes a real issue, it's better addressed in mod-audit by filtering those events server-side.

src/hooks/useUserAuditDataQuery/useUserAuditDataQuery.js
slaemmer: Fix is merged.

src/views/UserDetail/UserDetail.js
slaemmer: Closed the ticket and PR due to changed requirements ([the Actions menu and Tags button should not change when the Version history pane is opened](https://folio-org.atlassian.net/browse/UIU-3566). 

Removed the tooltip and kept the actions button enabled.

src/views/UserDetail/components/UserVersionHistory/UserVersionHistory.js
slaemmer: Changed to totalRecords and opened [MODAUD-310](https://folio-org.atlassian.net/browse/MODAUD-310) to make the necessary backend changes.

src/hooks/useAuditSettingsQuery/useAuditSettingsQuery.js
slaemmer: Didn't know it could, thanks. Updated to be called with `{ enabled: Boolean(enabled) }` [here](https://github.com/folio-org/ui-users/pull/3035/changes/e11a2e62bbe5fde112ad9f475dcb30768386274c).

Dmytro-Melnyshyn (approved):

PR #3039, opened 2026-05-06
[UIU-3521] Update IfPerm in user edit form from `ui-users-roles.view` to `ui-authorization-roles.users.settings.view`

ryandberger:
- Fulfills [UIU-3521](https://folio-org.atlassian.net/browse/UIU-3521).
- Switching IfPerm checks for roles to be more specific using `ui-authorization-roles.users.settings.view` per ticket. The previously used perm `ui-users-roles.view` includes `ui-authorization-roles.users.settings.view` as a subpermission so accounts that rely on `ui-users-roles.view` will be unaffected. 
- Accounts that just have `ui-authorization-roles.users.settings.view` will benefit as they expect to see Roles in Settings and Users modules.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change history for ui-users
 
+## 13.1.0 IN PROGRESS
+
+* Update IfPerm in user edit form from `ui-users-roles.view` to `ui-authorization-roles.users.settings.view` since `ui-users-roles.view` includes extraneous subpermissions. Refs UIU-3521.
+
 ## [13.0.0] (https://github.com/folio-org/ui-users/tree/v13.0.0) (2026-04-16)
 [Full Changelog](https://github.com/folio-org/ui-users/compare/v12.1.16...v13.0.0)
 * Collect coverage from unit tests. Refs UIU-3356.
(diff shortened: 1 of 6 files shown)


src/views/UserEdit/UserForm.js
```diff
@@ -500,7 +500,7 @@ class UserForm extends React.Component {
                       initialValues={initialValues}
                       setButtonRef={this.setButtonRef}
                     /> :
-                    <IfPermission perm="ui-users.roles.view">
+                    <IfPermission perm="ui-authorization-roles.users.settings.view">
```
zburke, line 503: I wonder if are we just replacing one bug with another here. Since the rendered component is `EditUserRoles`, I would expect the permission gating it to be `.manage`.

zburke (approved):
The very premise of this ticket, using permissions from a separate application, makes me uncomfortable. Nonetheless, the request is faithfully executed.

src/views/UserEdit/UserForm.js
ryandberger: The name of the component is a little misleading. It shows all the Roles as you'd expect with `.view` but the Remove icons and Add Roles button are gated by `.manage`. This change makes our permission usage more consistent.

CHANGELOG.md
```diff
@@ -81,6 +81,7 @@
 * Handle non-404 4xx errors in UserDetail. Refs UIU-3057.
 * Update the `index.all` translation. Refs UIU-1027.
 * Select user roles' modal's "Select all" button selects all *visible* records, following convention. Refs UIU-3548.
+* Update IfPerm in user edit form from `ui-users-roles.view` to `ui-authorization-roles.users.settings.view` since `ui-users-roles.view` includes extraneous subpermissions. Refs UIU-3521.
```
Dmytro-Melnyshyn: 13.0.0 is already released. This should be added in a new section.

PR #3047, opened 2026-06-04
UIU-3592: Fix Reading Room Access Last Updated column sorting by date instead of id

Dmytro-Melnyshyn:
## Purpose

The "Last Updated" column in the Reading Room Access accordion was sorting records by the record UUID (`id`) instead of the actual `metadata.updatedDate` timestamp, producing meaningless sort order.

## Approach

- Renamed `rraColumns.ID` -> `rraColumns.UPDATED_DATE` with value `'metadata.updatedDate'` in `constant.js`.
- Updated `getReadingRoomSortedData` in `util.js` to use lodash `get()` so it resolves nested dot-notation paths when sorting.
- Updated `getFormatter.js` and `ReadingRoomAccess.js` to use the renamed constant.
- Added unit tests for `getFormatter` covering all column formatters.

## Issues

[UIU-3592](https://folio-org.atlassian.net/browse/UIU-3592)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,12 +6,12 @@
 * Add Settings > Users > Version history page to configure retention, anonymization, and excluded fields. Refs UIU-3385.
 * Add a Version history pane on the user detail view, gated on the mod-audit interface. Refs UIU-3388.
 * Send bulk override renewal requests sequentially to prevent race conditions. Refs UIU-3561.
+* In the "Last Updated" column in the "Reading Room Access" accordion, sort by date, not by id. Fixes UIU-3592.
 
 ## [13.0.1] (https://github.com/folio-org/ui-users/tree/v13.0.1) (2026-06-03)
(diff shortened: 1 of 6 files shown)


BogdanDenis (approved):

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 [Full Changelog](https://github.com/folio-org/ui-users/compare/v13.0.0...v13.0.1)
 
 * Remove extra character from `Expired time` in Lost items requiring actual cost table. Refs UIU-3580.
+* Fix Reading Room Access "Last Updated" column sorting by date instead of id. Refs UIU-3592.
```
zburke: The change is to sort by date, not id. Can we say it that way? This reads like "fix problem with solution". That makes it sound like "sort by date" was the problem when in fact it is the solution.

zburke (approved):

PR #3049, opened 2026-06-09
UIU-3586: Fix JIT AuthUser creation dialog closing on error (follow up).

Dmytro-Melnyshyn:
## Purpose

Fix the keycloak (JIT AuthUser) confirmation dialog closing and navigating away from the Edit pane when keycloak user creation fails (e.g. for a user without a username).

## Description

When clicking Confirm in the "Keycloak user record" modal and the POST to `users-keycloak/auth-users` fails, the error toast was shown but the dialog was closed and the user was navigated away. The dialog should remain open so the user can cancel or retry.

## Approach

- Replaced `Promise.all` with `.catch(sendErrorCallout)` with `Promise.allSettled` to track succeeded vs failed keycloak creations independently.
- On failure: show error callout, remove successfully created tenants from the retry list, and return early — keeping the dialog open.
- On retry: only the previously failed tenants are reattempted.
- Invalidate `jit-auth-role` cache queries only if at least one creation succeeded, to keep the cache consistent.

## Issues

[UIU-3593](https://folio-org.atlassian.net/browse/UIU-3593)

## Screencasts

https://github.com/user-attachments/assets/397368e0-ddb3-4408-ab7a-b817ebd416e9

### DIFF
--- a/src/components/Wrappers/withUserRoles.js
+++ b/src/components/Wrappers/withUserRoles.js
@@ -181,7 +181,10 @@ const withUserRoles = (WrappedComponent) => (props) => {
   const confirmCreateKeycloakUser = async (onFinish) => {
     // Create keycloak records for all tenants that need them.
     const results = await Promise.allSettled(
-      tenantsWithoutKeycloakRef.current.map((tid) => createKeycloakUserForTenant(tid).then(() => tid))
+      tenantsWithoutKeycloakRef.current.map(async (tid) => {
+        await createKeycloakUserForTenant(tid);
+        return tid;
+      })
     );
 
     const succeeded = results.filter(r => r.status === 'fulfilled').map(r => r.value);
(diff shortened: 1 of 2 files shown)


SKarolFolio (approved):

src/components/Wrappers/withUserRoles.js
```diff
@@ -180,12 +180,25 @@ const withUserRoles = (WrappedComponent) => (props) => {
 
   const confirmCreateKeycloakUser = async (onFinish) => {
     // Create keycloak records for all tenants that need them.
-    await Promise.all(
-      tenantsWithoutKeycloakRef.current.map((tid) => createKeycloakUserForTenant(tid).catch(sendErrorCallout))
+    const results = await Promise.allSettled(
+      tenantsWithoutKeycloakRef.current.map((tid) => createKeycloakUserForTenant(tid).then(() => tid))
```
Dmitriy-Litvinenko: Could you please double check can we fix current problem?

Dmitriy-Litvinenko (approved):

PR #3062, opened 2026-08-04
UIU-3611: Refactored src/settings/index.js to use Settings instead of custom SettingsPage

Dmitriy-Litvinenko:
## Purpose
User settings pane does not highlight currently selected setting

## Approach
Migrate Settings to use stripes Settings component

Replace custom SettingsPage component with stripes-smart-components Settings component to align with platform standards and match the pattern used in other FOLIO modules like ui-circulation. The business logic written was also transferred.

Changes:
- Refactored src/settings/index.js to use Settings from @folio/stripes/smart-components instead of custom SettingsPage
- Removed manual routing logic from src/index.js as stripes Settings handles routing internally
- Added interface-based filtering logic in Settings component to conditionally include/exclude sections and pages based on interface availability (handles 'interface' and 'unlessInterface' properties)
- Created comprehensive test suite (src/settings/index.test.js) with 18 test cases covering rendering, interface filtering, props passing, edge cases, and integration

# Refs
https://issues.folio.org/browse/UIU-3611


src/settings/index.js
```diff
@@ -1,27 +1,68 @@
...
 
-const Settings = ({ children, match }) => {
+const Settings = (props) => {
   const { formatMessage } = useIntl();
-  const { path } = match;
+  const { stripes } = props;
+
+  // Filter sections and pages based on interface requirements
+  // This is necessary because stripes Settings component doesn't support IfInterface
+  const filteredSections = sections.map(section => {
```
vashjs, line 19: As you have couple elements here, my comment can be ignored. But in general I think filter() should be used before map(). You will map less elements in the end of the day.

